PR #5, opened 2025-09-17
feat: Hookup google login

NickBlow:
feat: Hookup google login

feat: Hook up integrations page


apps/platform/.dev.vars.example
BlankParticle, line 1: We can use normal `.env` files now

apps/platform/backend/trpc/context.ts
```diff
@@ -0,0 +1,21 @@
...
+  const auth = getAuth();
+  const session = await auth.api.getSession({
+    headers: opts.req.headers,
+  });
+
+  return {
+    db,
+    session,
+    user: session?.user || null,
+  };
```
BlankParticle: put `db` and `auth` on Hono context, then everything can access it

BlankParticle: ok done

PR #21, opened 2025-09-18
feat: Estate prefix

NickBlow:


apps/os/app/routes/root-redirect.tsx
```diff
@@ -0,0 +1,59 @@
+import { useEffect } from "react";
+import { useNavigate } from "react-router";
+import { authClient } from "../lib/auth-client.ts";
+import { getSelectedEstate } from "../lib/estate-cookie.ts";
+import { trpc } from "../lib/trpc.ts";
+import { GlobalLoading } from "../components/global-loading.tsx";
+
+export default function RootRedirect() {
```
BlankParticle: We can take advantage of `loader` here
we just need to make a serverSide trpc client to use that

NickBlow: can just call the function directly i guess?

BlankParticle: yeah thats also possible, but we need to setup the trpc client for jonas anyways

PR #38, opened 2025-09-19
Slack webhook store + message history for agent

zakhardavydov:
Fixes https://linear.app/iterate-com/issue/ITE-2944/bring-back-slack-webhook-storage-and-associated-features

- Created a new slack_webhook_events database table to store all incoming Slack webhook events with indexed fields for efficient querying by channel, thread, timestamp, and type, enabling reliable message history tracking.
- Enhanced the Slack agent to fetch and provide previous messages in a thread as context to the LLM, replacing the previously commented-out functionality with working database queries that retrieve thread conversations chronologically.
- Refactored Slack event processing to better handle message metadata extraction (timestamps, channels, thread IDs) and routing, with enhanced support for reaction events and more robust event classification using the stored webhook data.

### DIFF
--- a/apps/os/backend/agent/url-content-handler.ts
+++ b/apps/os/backend/agent/url-content-handler.ts
@@ -14,7 +14,7 @@ interface UrlTypeInfo {
 
 function extractDetailsFromSlackLink(url: string) {
   const url = new URL(input.link);
-  const [, , channelId, pts] = url.pathname.split("/");
+  const [_whateverThisIs, _andThisToo, channelId, pts] = url.pathname.split("/");
   const ts = pts.slice(1);
   const threadTs = url.searchParams.get("thread_ts");
 
--- a/apps/os/backend/db/schema.ts
+++ b/apps/os/backend/db/schema.ts
@@ -263,7 +263,7 @@ export const agentInstanceRouteRelations = relations(agentInstanceRoute, ({ one 
 }));
 
 export const slackWebhookEvent = pgTable(
-  "slack_webhook_events",
+  "slack_webhook_event",
   (t) => ({
     id: iterateId("slackevent"),
     data: t.jsonb().$type<SlackEvent>().notNull(),
--- a/apps/os/backend/integrations/slack/slack.ts
+++ b/apps/os/backend/integrations/slack/slack.ts
@@ -338,6 +338,6 @@ async function getRoutingKey({
   if (!payload.event || !payload.team_id) {
     throw new Error("No event or team_id found in slack webhook payload");
   }
-  const prefix = `slack-${estateId}-team-${payload.team_id}`;
-  return `${prefix}-ts-${threadTs}`;
+  const suffix = `slack-${estateId}-team-${payload.team_id}`;
+  return `ts-${threadTs}-${suffix}`;
 }
--- a/apps/os/backend/db/migrations/0004_loose_tyrannus.sql
+++ b/apps/os/backend/db/migrations/0004_loose_tyrannus.sql
@@ -1,4 +1,4 @@
-CREATE TABLE "slack_webhook_events" (
+CREATE TABLE "slack_webhook_event" (
 	"id" text PRIMARY KEY NOT NULL,
 	"data" jsonb NOT NULL,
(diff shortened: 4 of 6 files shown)


apps/os/backend/agent/url-content-handler.ts
```diff
@@ -13,6 +12,15 @@ interface UrlTypeInfo {
   slackMatch?: RegExpMatchArray;
 }
 
+function extractDetailsFromSlackLink(url: string) {
+  const url = new URL(input.link);
+  const [, , channelId, pts] = url.pathname.split("/");
```
mmkal:
```suggestion
  const [_whateverThisIs, _andThisToo, channelId, pts] = url.pathname.split("/");
```

apps/os/backend/db/schema.ts
```diff
@@ -259,3 +261,32 @@ export const agentInstanceRouteRelations = relations(agentInstanceRoute, ({ one
     references: [agentInstance.id],
   }),
 }));
+
+export const slackWebhookEvent = pgTable(
+  "slack_webhook_events",
```
jonastemplestein:
```suggestion
  "slack_webhook_event",
```

apps/os/backend/db/schema.ts
```diff
@@ -259,3 +261,32 @@ export const agentInstanceRouteRelations = relations(agentInstanceRoute, ({ one
...
+    ts: t.text("ts"),
+    thread_ts: t.text("thread_ts"),
+    channel: t.text("channel"),
+    type: t.text("type"),
+    subtype: t.text("subtype"),
+    user: t.text("user"),
+    estateId: t.text().notNull(),
+    ...withTimestamps,
+  }),
+  (table) => [
```
jonastemplestein, line 279: I think there is some way to index the json blob also but we can add that later

apps/os/backend/agent/url-content-handler.ts
```diff
@@ -47,19 +55,19 @@ async function determineUrlType(url: string, headResponse?: Response): Promise<U
   };
 }
 
-async function handleSlackUrl(url: string, _slackMatch: RegExpMatchArray) {
+async function handleSlackUrl(params: { url: string; db: DB }) {
```
mmkal, line 58: not sure it's worth passing the db around everywhere - is it to avoid calling `getDb()`

apps/os/backend/integrations/slack/slack.ts
```diff
@@ -307,31 +326,18 @@ slackApp.post("/webhook", async (c) => {
...
-  }
-
-  console.warn(
-    "Didn't know how to turn this slack webhook payload into a routing key and durable object name",
-    payload,
-  );
-  throw new Error(
-    "Didn't know how to turn this slack webhook payload into a routing key and durable object name",
-  );
+  return `${prefix}-ts-${threadTs}`;
```
mmkal: since we're doing this now - I think a suffix is better, maybe because I'm used to it. But the estate id is not useful info to look at it, so IMO it should go at the end?

apps/os/backend/db/schema.ts
```diff
@@ -259,3 +261,32 @@ export const agentInstanceRouteRelations = relations(agentInstanceRoute, ({ one
...
+  (t) => ({
+    id: iterateId("slackevent"),
+    data: t.jsonb().$type<SlackEvent>().notNull(),
+    ts: t.text("ts"),
+    thread_ts: t.text("thread_ts"),
+    channel: t.text("channel"),
+    type: t.text("type"),
+    subtype: t.text("subtype"),
+    user: t.text("user"),
+    estateId: t.text().notNull(),
```
jonastemplestein, line 276: I think an argument could be made for making this optional, to debug cases where slack thinks an oauth connection succeeded but we didn't record it right on our end. 

And possibly also to join via slack team id

But I wouldn't change it now if it's working

apps/os/backend/agent/url-content-handler.ts
zakhardavydov: Yep, and make it explicit that this function requires db access

PR #39, opened 2025-09-19
nullable routingKey

mmkal:
@jonastemplestein small one

### DIFF
--- a/apps/os/backend/integrations/slack/slack.ts
+++ b/apps/os/backend/integrations/slack/slack.ts
@@ -199,7 +199,7 @@ slackApp.post("/webhook", async (c) => {
 //       return;
 //   }
 
-//   let slackAgentInstanceName: string || case null = n":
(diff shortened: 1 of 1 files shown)


apps/os/backend/integrations/slack/slack.ts
```diff
@@ -327,11 +331,95 @@ function getRoutingKey({ payload, estateId }: { payload: SlackWebhookPayload; es
...
-  console.warn(
-    "Didn't know how to turn this slack webhook payload into a routing key and durable object name",
-    payload,
-  );
-  throw new Error(
-    "Didn't know how to turn this slack webhook payload into a routing key and durable object name",
-  );
+  const eventType = payload.event.type;
+  switch (eventType) {
+    case "user_change":
```
mmkal, line 336: i got cursor to add all of them exhaustively so we get a compiler error if there's a new event type in future until we explicitly map it to a routing key or decide not to

PR #41, opened 2025-09-19
Add sdk package back

jonastemplestein:
## Changes to os package
Exports new sdk folder from os package

```typescript
  "exports": {
    "./sdk": "./sdk/index.ts"
  },
```

This just exports a bunch of stuff we already have

```typescript
import dedent from "dedent";

export {
  contextRulesFromFiles,
  matchers,
  defineRule,
  defineRules,
} from "../backend/agent/context.ts";
export { defineConfig } from "./iterate-config.ts";
export { f } from "../backend/agent/prompt-fragments.ts";
export { dedent };

export type { ToolSpec } from "../backend/agent/tool-schemas.ts";
export type { PromptFragment } from "../backend/agent/prompt-fragments.ts";
export type { ContextRule } from "../backend/agent/context.ts";
```

## New @iterate-com/sdk package
This does two things
- re-exports everything that @iterate-com/os/sdk exports
- adds a tiny CLI bin called `iterate` so you can do `pnpm iterate iterate.config.ts` to print out an iterate config

## New estates/ folder
- Contains various iterate estate configurations, including our own "main" one in estates/iterate/iterate.config.ts

# Next steps 
### Today
- make it possible to point `pnpm dev` at a local iterate config 
- actually publish os and sdk packages
- will then run `pnpm iterate iterate.config.ts` in a CF container

### Next week
- create mechanism for editing customer repos so we can migrate them

### Whenever we get to it
- use rollup or whatever to make sdk package super mini small
- move context.ts and prompt-fragments.ts etc into the sdk/ folder (and out of the agent/ folder)

### DIFF
--- a/packages/sdk/cli.js
+++ b/packages/sdk/cli.js
@@ -15,10 +15,14 @@ async function main() {
   const code = `console.log(JSON.stringify((await import(${JSON.stringify(fileUrl)})).default, null, 2));`;
 
   // Run node synchronously with module eval and TS stripping
-  const result = spawnSync('pnpx', ['tsx', '--input-type=module', '--experimental-strip-types', '-e', code], {
-    stdio: 'inherit',
-    cwd: process.cwd()
-  });
+  const result = spawnSync(
+    "pnpx",
+    ["tsx", "--input-type=module", "--experimental-strip-types", "-e", code],
+    {
+      stdio: "inherit",
+      cwd: process.cwd(),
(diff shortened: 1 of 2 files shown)


packages/sdk/cli.js
mmkal, line 1: let's do `mv dist/cli.js cli.js` - the `dist/` folder is usually gitignored/generated.

PR #45, opened 2025-09-19
avoid ts-expect-error on SlackAgent.getStubByName

mmkal:
@jonastemplestein your generics tricks and this-binding were too much for the compiler to keep track of - I simplified and accepted the tradeoff that calling certain static methods on `SlackAgent` gives you a stub typed for an `IterateAgent` rather than `SlackAgent`. That was a problem in zero of the places we use it so far, and if a future case arises, it's solveable by a pretty innocuous explicit cast like `SlackAgent.getStub(...) as DurableObjectStub<SlackAgent>`

### DIFF
--- a/apps/os/backend/agent/iterate-agent.ts
+++ b/apps/os/backend/agent/iterate-agent.ts
@@ -133,22 +133,19 @@ export class IterateAgent<Slices extends readonly AgentCoreSlice[] = CoreAgentSl
   override observability = undefined;
 
   // Resolve the DO namespace for the concrete subclass at runtime
-  static getNamespace<T extends typeof IterateAgent>(
-    this: T,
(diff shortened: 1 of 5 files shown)


apps/os/backend/agent/iterate-agent.ts
```diff
@@ -231,19 +215,16 @@ export class IterateAgent<Slices extends readonly AgentCoreSlice[] = CoreAgentSl
       matchingAgents.map((record) => this.getStubFromDatabaseRecord(record)),
     );
 
-    return stubs;
+    return stubs as unknown[] as DurableObjectStub<IterateAgent>[];
```
mmkal, line 218: for some reason without this one I got `The inferred type of this node exceeds the maximum length the compiler will serialize. An explicit type annotation is needed.`!!

PR #46, opened 2025-09-19
Small tidy ups

jonastemplestein:
- bring back slack reaction and update message tools for agent
- create `schema` export from db client module so we can use `schema.someTable` everywhere and get auto-imports
- deleted a little bit of code

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -389,12 +389,6 @@ importers:
         specifier: workspace:*
         version: link:../../packages/sdk
 
-  estates/iterate:
-    devDependencies:
-      '@iterate-com/sdk':
-        specifier: workspace:*
-        version: link:../../packages/sdk
-
   estates/sample-pirate:
     devDependencies:
       '@iterate-com/sdk':


apps/os/backend/agent/iterate-agent.ts
```diff
@@ -910,7 +921,9 @@ export class IterateAgent<Slices extends readonly AgentCoreSlice[] = CoreAgentSl
     return {};
   }
   getAgentDebugURL() {
-    return { debugURL: `${this.env.VITE_PUBLIC_URL}/agents/${this.constructor.name}/${this.name}` };
+    return {
+      debugURL: `${this.env.VITE_PUBLIC_URL}/org_01k5ffm2z9fexvxbfzt2hvstzb/est_01k5ffm2zcfexvxbge5f6excbe/agents/${this.constructor.name}/${this.name}`,
```
zakhardavydov: Is this hard code intended?

apps/os/backend/integrations/slack/slack.ts
```diff
@@ -101,31 +111,53 @@ slackApp.post("/webhook", async (c) => {
...
     reason: "Slack webhook received",
   });
 
+  // if we just created the agent, we quickly respond with the :eyes: emoji
+  if (created) {
+    c.executionCtx.waitUntil((async () => {
+      if (!botUserId || !body.event || !messageMetadata) {
+        return;
+      }
+      const shouldInclude = shouldIncludeEventInConversation(body.event, botUserId);
```
zakhardavydov: Do we need to add this check?

PR #47, opened 2025-09-19
feat: Container iterate.config.ts execution

NickBlow:


apps/os/backend/worker.ts
```diff
@@ -127,6 +128,38 @@ app.get("/api/ws/:organizationId", async (c) => {
...
+// Test build endpoint for sandbox
+app.post("/api/test-build", async (c) => {
+  try {
+    // Parse the request body
+    const body = await c.req.json<{
+      githubRepoUrl: string;
+      githubToken: string;
+      commitHash?: string;
+      workingDirectory?: string;
+    }>();
```
BlankParticle: use `@hono/zod-validator`

apps/os/backend/sandbox/run-config.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  // Validate working directory path if provided
+  if (workingDirectory && workingDirectory.startsWith("/")) {
+    return {
+      error: "Working directory should be a relative path within the repository",
+    };
+  }
+
+  // Get sandbox instance
+  const sandbox = getSandbox(env.SANDBOX, `test-build-${Date.now()}`);
```
mmkal: not sure how long this will last but since this is prod can we use a less "test"-ish and collisionish name? like `sandbox-${createUUID()}`

BlankParticle: actually just use `typeid("sndbx")`, we alr use it everywhere

PR #52, opened 2025-09-19
Nickblow/webhook hookup

NickBlow:


apps/os/backend/sandbox/run-config.ts
```diff
@@ -49,8 +51,35 @@ export async function runConfigInSandbox(
...
+    if (checkoutResult.exitCode !== 0) {
+      return {
+        error: "Failed to checkout branch",
+        details: checkoutResult.stderr,
+        commitHash: branch,
+      };
+    }
+
+    // If a specific commit hash is also provided, checkout that commit on the branch
+    if (commitHash) {
```
jonastemplestein: Why is this nested under if(branch)? 

I think we can reduce this all to a single `await sandbox.exec(...)`?

apps/os/backend/worker.ts
```diff
@@ -123,6 +132,215 @@ app.get("/api/ws/:organizationId", async (c) => {
   );
 });
 
+// GitHub webhook handler
+app.post("/api/webhooks/github", async (c) => {
```
jonastemplestein: Why not `/api/integrations/github/webhook`? There's already a github app mounted at `/api/integrations/github` above

In general I would also prefer if we didn't make worker.ts huge and treated it more like a router and  configuration file

apps/os/app/routes/estate.tsx
```diff
@@ -131,9 +157,26 @@ function EstateContent() {
...
 
+  const { data: builds, isLoading: buildsLoading } = useSuspenseQuery(
+    trpc.estate.getBuilds.queryOptions({
+      estateId: estateId,
+      limit: 10,
+    }),
+  );
+
+  // Initialize form values when connected repo data loads
+  React.useEffect(() => {
```
NickBlow: this is a stupid useeffect, will remove

apps/os/env.ts
```diff
@@ -14,6 +14,13 @@ export type CloudflareEnv = Env & {
   GITHUB_APP_CLIENT_SECRET: string;
   GITHUB_APP_PRIVATE_KEY: string;
   GITHUB_APP_SLUG: string;
+  GITHUB_WEBHOOK_SECRET: string;
+
+  // temporarily using this old env var i found in doppler until i can hook up to better auth accounts table
+  SLACK_PROXY_BOT_TOKEN: string;
```
jonastemplestein: I think this might have come back in a conflict or something. It's not used on main

apps/os/env.ts
```diff
@@ -14,6 +14,13 @@ export type CloudflareEnv = Env & {
...
+
+  // Durable Object bindings
+  ORGANIZATION_WEBSOCKET: DurableObjectNamespace;
```
jonastemplestein: Does `pnpm wrangler types` not generate this into `worker-configuration.d.ts`

apps/os/backend/sandbox/run-config.ts
```diff
@@ -49,8 +51,35 @@ export async function runConfigInSandbox(
     };
   }
 
-  // Checkout specific commit if provided
-  if (commitHash) {
+  // Checkout specific branch or commit
+  if (branch) {
```
NickBlow: actually this can be consolidated, don't need two code paths

apps/os/backend/integrations/github/github-utils.ts
```diff
@@ -17,3 +20,92 @@ export const generateGithubJWT = async () => {
...
+    )
+    .limit(1);
+
+  return githubInstallation;
+};
+
+export const getGithubRepoForEstate = async (db: DB, estateId: string) => {
+  const [estate] = await db
+    .select({
+      connectedRepoId: schemas.estate.connectedRepoId,
```
jonastemplestein, line 48: I wouldn't do anything about it now. But if we did want to in the future say "every estate has a git repo" and just host some of the repos ourselves using gitlip in a durable object, the concept of `connectedRepoId` will probably not work anymore

That's a (very good) github specific concept. But it isn't as easily expanded to non-github repos (as e.g. a URL would be)

PR #58, opened 2025-09-22
work around db deadlock

mmkal:
instead of trying to fix deadlock this morning, work around it:

1. we're already querying the db in the static get-an-agent helper methods. just add a join to also pull the context rules at that time and add them to the `databaseRecord` field
2. use `this.databaseRecord.contextRules` in the getContextRules method

Next up (once this PR has unblocked today's goal of make-context-rules-work):

1. add a todo to uncomment the database query
2. add a timeout to the database query, with a console.error when it exceeds 100ms - this suggests deadlock
3. fix the deadlock or figure out why it's happening

Test of it working:

```
pnpm dev -c estates/sample-pirate/iterate.config.ts
```

Indeed speaks like a pirate:

image

### DIFF
--- a/apps/os/backend/agent/iterate-agent.ts
+++ b/apps/os/backend/agent/iterate-agent.ts
@@ -219,7 +219,7 @@ export class IterateAgent<Slices extends readonly AgentCoreSlice[] = CoreAgentSl
       matchingAgents.map((record) =>
         this.getStubFromDatabaseRecord({
           ...record,
-          contextRules: record.estate.iterateConfigs[0].config.contextRules ?? [],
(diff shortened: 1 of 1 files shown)


apps/os/backend/agent/agent-core.test.ts
```diff
@@ -2647,7 +2647,21 @@ describe("AgentCore ephemeralPromptFragments", () => {
...
+            This is important ephemeral context data
+          </context>
+        </test-context>
+
+        <another-context>
+          <rule>
+            Another piece of context information
+          </rule>
+        </another-context>"
+      `);
```
mmkal, line 2664: this is broken on main (jonas's fix to my bug also fixed this snapshot)

PR #59, opened 2025-09-22
re-add deadlock with a timeout + error log

mmkal:
image

### DIFF
--- a/apps/os/backend/agent/iterate-agent.ts
+++ b/apps/os/backend/agent/iterate-agent.ts
@@ -1,3 +1,4 @@
+import pTimeout from "p-suite/p-timeout";
 import { Agent as CloudflareAgent } from "agents";
 import { formatDistanceToNow } from "date-fns";
 import { z } from "zod/v4";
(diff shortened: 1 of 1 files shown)


apps/os/backend/agent/iterate-agent.ts
```diff
@@ -702,14 +703,14 @@ export class IterateAgent<Slices extends readonly AgentCoreSlice[] = CoreAgentSl
...
-    // todo: uncomment this - with a timeout + warning when it's too slow, it suggests we have some kind of deadlock
-    // when querying the db during do initialisation
-    // const dbConfig = await this.db.query.iterateConfig.findFirst({
-    //   where: eq(schema.iterateConfig.estateId, this.databaseRecord.estateId),
-    // });
-    const rules = [...defaultRules, ...this.databaseRecord.contextRules];
+    const { db, databaseRecord } = this;
+    // sadly drizzle doesn't support abort signals yet https://github.com/drizzle-team/drizzle-orm/issues/1602
+    const rulesFromDb = await pTimeout(IterateAgent.getRulesFromDB(db, databaseRecord.estateId), {
+      milliseconds: 250,
```
BlankParticle, line 710: give this like atleast a second lol, it might be slow even when we fix it

mmkal: hmm this will contribute to our bot's response speed though, and it's really not that necessary. we could even skip hitting the db here _forever_ on DO initialisation, because we've just got up-to-date context rules a few milliseconds ago

BlankParticle: oh yeah, this is going to get always hit, makes sense to not wait for it

PR #62, opened 2025-09-22
mcp: oauth backed by better auth

zakhardavydov:
Implements OAuth authentication for Model Context Protocol (MCP) server connections, replacing the stub implementation with a full BetterAuth integration.

### Key Changes

- New BetterAuthMCPOAuthProvider: Complete OAuth provider with token management, state handling, and database integration
- MCP OAuth Callback Endpoint: /integrations/callback/mcp handles authorization codes and triggers agent reconnection
- Enhanced Connection Flow: Added reconnection support and improved error handling for OAuth failures
- Frontend Integration: OAuth callback parameter handling in agents page with automatic URL cleanup
- Security: Implements OAuth 2.0 best practices with state validation and estate permission checks

### DIFF
--- a/apps/os/backend/agent/mcp/mcp-oauth-provider.ts
+++ b/apps/os/backend/agent/mcp/mcp-oauth-provider.ts
@@ -10,8 +10,27 @@ import type { AgentDurableObjectInfo } from "../../auth/oauth-state-schemas.ts";
 import { DynamicClientInfo } from "../../auth/oauth-state-schemas.ts";
 
 /**
- * Inspired by agents SDK implementation https://github.com/cloudflare/agents/blob/4e087816e8c011f87eedb3302db80724fe6080ac/packages/agents/src/index.ts
- * Durable Object version reference implementation https://github.com/cloudflare/agents/blob/6db2cd6f1497705f8636b1761a2db364d49d4861/packages/agents/src/mcp/do-oauth-client-provider.ts
+ * Connector between MCP OAuthClientProvider and our better auth plugin
+ *
+ * Inspired by agents SDK implementation: https://github.com/cloudflare/agents/blob/4e087816e8c011f87eedb3302db80724fe6080ac/packages/agents/src/index.ts
+ * Durable Object version reference implementation: https://github.com/cloudflare/agents/blob/6db2cd6f1497705f8636b1761a2db364d49d4861/packages/agents/src/mcp/do-oauth-client-provider.ts
+ *
+ * MCPClientManager requires reconnect argument with id (serverId from previos connect), oauthClientId, and oauthCode from oauth callback (cloudflare craft).
+ * It turns out we don't need to pass the serverId and the reconnect will still work as expected.
+ * This is good because we are creating a new MCPClientManager for each connect request (and a newserverId is generated every time).
+ *
+ * So we call MCPClientManager.connect twice. First without reconnect, then with reconnect.
+ * In the first call, we create the dynamic client and save the account to the dynamic_client_info table.
+ * We save code verifier to the verification table for 10 minutes.
+ * We save state object to the verification table for 10 minutes. We add agentDurableObject to the state object so we can get the stub in the callbackMCP.
+ * Then we construct the redirect url and let the connector handle the translation of the url to MCP:OAUTH_REQUIRED event.
+ * Then LLM generates a message to the user to authorize the connection.
+ * The user clicks the url and is redirected to the authorization url.
+ * On redirect, we call callbackMCP from integrations auth plugin.
+ * We add another connect request with reconnect argument.
+ * This time OAuthClientProvider will exchange the code for tokens and save them to the account table.
+ * Then when it requires to get tokens(), it will get them from the account table.
+ * Congrats, we successfully connected to the MCP server.
  */
 export class MCPOAuthProvider implements AgentsOAuthProvider {
   clientId: string | undefined;


apps/os/backend/agent/mcp/mcp-event-hooks.ts
```diff
@@ -260,10 +271,46 @@ export async function handleMCPConnectRequest(
...
+
+  const agentClassName = agentDurableObjectName.startsWith("SlackAgent-")
+    ? "SlackAgent"
+    : "IterateAgent";
+
+  // Get the organization ID for the estate to construct proper callback URL
+  let organizationId: string | undefined;
+  if (!finalRedirectUrl) {
+    try {
+      const estateWithOrg = await db.query.estate.findFirst({
```
jonastemplestein: FYI according to cloudflare it takes up to 30ms to create the DB connection before the first query - I don't have enough context to know whether this matters here

apps/os/backend/agent/agent-core.ts
```diff
@@ -341,7 +340,7 @@ export class AgentCore<
   }
 
   // Dependencies & slices ---------------------------------------------------
-  private readonly deps: MergedDepsForSlices<Slices>;
+  readonly deps: MergedDepsForSlices<Slices>;
```
jonastemplestein: What's the use-case for accessing the deps from outside the agent core class? Is a bit surprising that we on the one hand _inject_ some dependencies into each `AgentCore` but then also somehow access them from outside

apps/os/backend/auth/integrations.ts
```diff
@@ -73,6 +75,158 @@ export const integrationsPlugin = () =>
   ({
     id: "integrations",
     endpoints: {
+      // MCP OAuth callback endpoint
```
jonastemplestein, line 79: This is a little above my paygrade, so this is just to say I haven't carefully reviewed this part

apps/os/app/routes/agents.tsx
```diff
@@ -1612,6 +1617,49 @@ export default function AgentsPage() {
     },
   });
 
+  // Handle MCP OAuth callback
+  useEffect(() => {
```
jonastemplestein: Is this here functionality just specific to OAUTH initiated from the agents debug UI?

If so, I wonder if we should just drop it and instead rely on the oauth links that the agent produces?

And if this is indeed somehow used for MCP oauth redirects originating from a slack conversation with the bot, I'd move it out of this file, which otherwise just contains our agents UI

jonastemplestein (commented):
Nice job getting this working!

I have very little context on how our MCP implementation works or what the interesting bits here are, so have skimmed it and left a few high level comments

I'd definitely want @mmkal and @BlankParticle to know how this works - but as far as I'm concerned it's not a blocker to merging, given we just want to get everything working

It would have helped me a lot if we'd spent 10 mins ahead of time talking through any interesting design points / for you to tell me how to think about this problem and solution

apps/os/app/routes/agents.tsx
zakhardavydov: Agreed, we should just drop it

apps/os/backend/agent/agent-core.ts
zakhardavydov: Reverted it to private

apps/os/backend/agent/mcp/mcp-oauth-provider.ts
```diff
@@ -1,51 +1,294 @@
...
+          providerMetadata: {
+            serverUrl: this.params.serverUrl,
+            integrationSlug: this.params.integrationSlug,
+            clientId: this.clientId,
+          },
+          createdAt: new Date(),
+          updatedAt: new Date(),
+        })
+        .onConflictDoNothing();
+    }
```
BlankParticle: Wrap in a transaction
when you create a account use `.returning()` to get back the id, incase we change it later

apps/os/backend/agent/mcp/mcp-oauth-provider.ts
```diff
@@ -1,51 +1,294 @@
...
+  }
+
+  async saveClientInformation(info: any): Promise<void> {
+    this.clientId = info.client_id;
+
+    const verificationKey = `mcp-client-${this.providerId}-${info.client_id}`;
+    const expiresAt = new Date(Date.now() + 365 * 24 * 60 * 60 * 1000); // 1 year
```
BlankParticle: why do we need to store the state for an year ?
10min is like enough time to complete an oauth flow

apps/os/backend/agent/mcp/mcp-oauth-provider.ts
```diff
@@ -1,51 +1,294 @@
...
+        createdAt: new Date(),
+        updatedAt: new Date(),
+      });
+
+      this.accountId = newAccountId;
+
+      await this.params.db
+        .insert(schema.estateAccountsPermissions)
+        .values({
+          id: typeid("eap").toString(),
```
BlankParticle: `id` is handled by drizzle, drop it

apps/os/backend/agent/mcp/mcp-oauth-provider.ts
```diff
@@ -1,51 +1,294 @@
...
+          estateId: this.params.estateId,
+          createdAt: new Date(),
+          updatedAt: new Date(),
+        })
+        .onConflictDoNothing();
+
+      await this.params.db
+        .insert(schema.providerEstateMapping)
+        .values({
+          id: typeid("pem").toString(),
```
BlankParticle: same here

apps/os/backend/agent/mcp/mcp-oauth-provider.ts
```diff
@@ -1,51 +1,294 @@
...
+
+    const verificationKey = `mcp-client-${this.providerId}-${info.client_id}`;
+    const expiresAt = new Date(Date.now() + 365 * 24 * 60 * 60 * 1000); // 1 year
+
+    await this.params.db
+      .delete(schema.verification)
+      .where(eq(schema.verification.identifier, verificationKey));
+
+    await this.params.db.insert(schema.verification).values({
+      id: typeid("ver").toString(),
```
BlankParticle: and here

apps/os/backend/agent/mcp/mcp-event-hooks.ts
```diff
@@ -260,10 +271,46 @@ export async function handleMCPConnectRequest(
     return events;
   }
 
+  const db = getDb();
+  const auth = getAuth(db);
+
+  const agentClassName = agentDurableObjectName.startsWith("SlackAgent-")
+    ? "SlackAgent"
+    : "IterateAgent";
```
mmkal: could we do someting like `.split('-')[0]` here to avoid hard-coding in too many places. when we change the class hierarchy this will bite us

apps/os/backend/agent/mcp/mcp-oauth-provider.ts
```diff
@@ -1,51 +1,294 @@
...
+        })
+        .where(eq(schema.account.id, existingAccount.id));
+
+      this.accountId = existingAccount.id;
+    } else {
+      const newAccountId = typeid("acc").toString();
+
+      await this.params.db.insert(schema.account).values({
+        id: newAccountId,
+        accountId: this.clientId,
```
BlankParticle: we should call this `externalUserId` or something, clientId sounds like the OAuth Client ID
unless thats what this actually is, if which it is, we need to re-think this

apps/os/backend/agent/mcp/mcp-oauth-provider.ts
```diff
@@ -1,51 +1,294 @@
...
+import { eq, and } from "drizzle-orm";
+import { typeid } from "typeid-js";
+import type { Auth } from "../../auth/auth.ts";
+import type { DB } from "../../db/client.ts";
+import * as schema from "../../db/schema.ts";
 
-export class MCPOAuthProvider implements AgentsOAuthProvider {
-  clientId: string;
-  serverId: string;
+export class BetterAuthMCPOAuthProvider implements AgentsOAuthProvider {
```
mmkal: is this implementation based on some better-auth/cloudflare docs somewhere? would be good to link to those docs here so we get a sense of where the inspiration came from!

mmkal: I think this should keep the old name `MCPOAuthProvider`, better-auth is an implementation detail. Also you're aliasing it to the old name below but not using the alias

BlankParticle (commented):
somewhat confusing flow, whats being stored where
would appreciate a high level overview on a call
before we merge this to catch any potential issues before its too late

apps/os/backend/agent/mcp/mcp-event-hooks.ts
```diff
@@ -260,10 +271,46 @@ export async function handleMCPConnectRequest(
...
+        db,
+        userId: userId!,
+        estateId: estateId,
+        integrationSlug: guaranteedIntegrationSlug,
+        serverUrl: formattedServerUrl,
+        callbackURL:
+          finalRedirectUrl ||
+          (organizationId
+            ? `${import.meta.env.VITE_PUBLIC_URL}/${organizationId}/${estateId}/agents/${agentClassName}/${agentDurableObjectName}`
+            : `${import.meta.env.VITE_PUBLIC_URL}/agents/${agentClassName}/${agentDurableObjectName}`),
```
mmkal: which url is which/why do they both work?

BlankParticle (changes requested):
blocking this merge for now

apps/os/backend/auth/integrations.ts
```diff
@@ -73,6 +75,158 @@ export const integrationsPlugin = () =>
...
+            agentDurableObjectName?: string;
+            serverId?: string;
+          };
+
+          if (state) {
+            const stateValue = await ctx.context.internalAdapter.findVerificationValue(state);
+            if (!stateValue) {
+              return ctx.json({ error: "Invalid or expired state" }, { status: 400 });
+            }
+            stateData = JSON.parse(stateValue.value);
```
mmkal: can we define a zod schema for this - feels likely that we'll end up with the wrong fields and this would be the right place to catch it via a `StateData.parse(...)`

mmkal: let's leave instructions about what the person who sees this should do? (either us, the developers, or the client user?)

apps/os/backend/auth/integrations.ts
```diff
@@ -73,6 +75,158 @@ export const integrationsPlugin = () =>
...
+          const {
+            var: { db },
+          } = getContext<{ Variables: Variables; Bindings: CloudflareEnv }>();
+
+          const estate = await db.query.estate.findFirst({
+            where: eq(schema.estate.id, stateData.estateId),
+            with: {
+              organization: {
+                with: {
+                  members: true,
```
mmkal: let's put the memebership check here, generally better practice to avoid telling non members that the estate even exists. like how github 404 looks the same whether you hit a private repo url or one that doesn't exist

apps/os/backend/auth/integrations.ts
zakhardavydov: Already done that

apps/os/backend/agent/mcp/mcp-oauth-provider.ts
zakhardavydov: this is the dynamic client info

apps/os/backend/agent/mcp/mcp-oauth-provider.ts
zakhardavydov: This is a dynamic client id, yes

apps/os/backend/agent/mcp/mcp-oauth-provider.ts
BlankParticle: I think it would be probably better to store it along with the account in some kind of metadata table. Verification table is supposed to be expiring.
If it doesn't work with the account table we can just make a new table for it

apps/os/backend/agent/mcp/mcp-oauth-provider.ts
```diff
@@ -1,51 +1,301 @@
...
-    return {
-      access_token: "test",
-      token_type: "Bearer",
-      refresh_token: "test",
-    };
+    if (this.params.reconnect) {
+      this.clientId = this.params.reconnect.oauthClientId;
+      this.serverId = this.params.reconnect.id;
+    }
+
```
mmkal:
```suggestion
```
remove

apps/os/backend/agent/mcp/mcp-oauth-provider.ts
```diff
@@ -1,51 +1,301 @@
...
+  constructor(
+    private params: {
+      auth: Auth;
+      db: DB;
+      userId: string;
+      estateId: string;
+      integrationSlug: string;
+      serverUrl: string;
+      callbackURL: string | undefined;
+      env?: { VITE_PUBLIC_URL?: string };
```
mmkal: why is env optional

apps/os/backend/agent/mcp/mcp-oauth-provider.ts
```diff
@@ -1,51 +1,301 @@
...
+      reconnect?: {
+        id: string;
+        oauthClientId?: string;
+        oauthCode?: string;
+      };
```
mmkal: We should compensate for cloudflare's weird implementation and lack of docs here. 

1. add jsdoc explaining what this does/how it's used
2. remove the `id` string, and make the other two required

as it is we're leaking cloudflare's (wrong) abstractions. we can pass `id: "cloudflare-requires-id"` in this class, to protect users of the class from worry about this hack

apps/os/backend/agent/mcp/mcp-event-hooks.ts
```diff
@@ -260,10 +269,21 @@ export async function handleMCPConnectRequest(
...
     ? new MCPOAuthProvider({
-        clientId: "iterate-mcp-client",
-        serverId: formattedServerUrl,
+        auth,
+        db,
+        userId: userId!,
+        estateId: estateId,
+        integrationSlug: guaranteedIntegrationSlug,
+        serverUrl: formattedServerUrl,
+        callbackURL: finalRedirectUrl,
```
mmkal: is `finalRedirectUrl` always defined (i.e. is it `string` rather than `string | undefined`)? if so, let's make it required in the constructor and then remove all the various places where we have to account for it being missing.

apps/os/backend/agent/mcp/mcp-event-hooks.ts
```diff
@@ -260,10 +269,21 @@ export async function handleMCPConnectRequest(
...
   const oauthProvider = requiresAuth
     ? new MCPOAuthProvider({
-        clientId: "iterate-mcp-client",
-        serverId: formattedServerUrl,
+        auth,
+        db,
+        userId: userId!,
+        estateId: estateId,
+        integrationSlug: guaranteedIntegrationSlug,
+        serverUrl: formattedServerUrl,
```
mmkal, line 282: Url/URL inconsistent casing!

apps/os/backend/agent/mcp/mcp-oauth-provider.ts
```diff
@@ -1,51 +1,301 @@
...
+        eq(schema.account.providerId, this.providerId),
+      ),
+    });
+
+    if (account?.accessToken) {
+      this.dbAccountId = account.id;
+      return {
+        access_token: account.accessToken,
+        token_type: "Bearer",
+        refresh_token: account.refreshToken || undefined,
```
mmkal, line 93: let's link to the bit of cloudflare code that refreshes

apps/os/backend/agent/mcp/mcp-oauth-provider.ts
```diff
@@ -1,51 +1,301 @@
...
   }
 
   get clientMetadata() {
     return {
-      client_name: "test",
-      client_uri: "test",
-      redirect_uris: ["test"] as string[],
+      client_name: `Iterate MCP Client - ${this.params.integrationSlug}`,
+      client_uri: this.baseUrl,
+      redirect_uris: [this.redirectUrl] as string[],
```
mmkal: why the cast here?

apps/os/backend/agent/mcp/mcp-oauth-provider.ts
```diff
@@ -1,51 +1,301 @@
...
+      return undefined;
+    }
+
+    try {
+      const clientInfo = DynamicClientInfo.parse(JSON.parse(verification.value));
+      return clientInfo;
+    } catch (error) {
+      console.error("Failed to parse client information:", error);
+      return undefined;
+    }
```
mmkal:
```suggestion
    const clientInfo = DynamicClientInfo.safeParse(JSON.parse(verification.value));
    if (!clientInfo.success) {
      console.error(`Failed to parse client information: ${z.prettifyError(clientInfo.error)}`)
      return undefined;
    }

    return clientInfo.data
```

apps/os/backend/agent/mcp/mcp-oauth-provider.ts
```diff
@@ -1,51 +1,301 @@
...
+      .delete(schema.verification)
+      .where(eq(schema.verification.identifier, verificationKey));
+
+    await this.params.db.insert(schema.verification).values({
+      identifier: verificationKey,
+      value: JSON.stringify(info),
+      expiresAt,
+      createdAt: new Date(),
+      updatedAt: new Date(),
+    });
```
mmkal: make a new table! don't abuse the poor `verifications` table

apps/os/backend/agent/mcp/mcp-oauth-provider.ts
```diff
@@ -1,51 +1,301 @@
...
+    await this.params.db
+      .delete(schema.verification)
+      .where(eq(schema.verification.identifier, verificationKey));
+
+    await this.params.db.insert(schema.verification).values({
+      identifier: verificationKey,
+      value: JSON.stringify(info),
+      expiresAt,
+      createdAt: new Date(),
+      updatedAt: new Date(),
```
BlankParticle: MORE TABLES 👏

apps/os/backend/agent/mcp/mcp-oauth-provider.ts
```diff
@@ -1,51 +1,301 @@
...
+    });
+  }
+
+  async redirectToAuthorization(authUrl: URL): Promise<void> {
+    const client_id = authUrl.searchParams.get("client_id");
+    if (client_id) {
+      this.clientId = client_id;
+    }
+
```
mmkal:
```suggestion
```

apparently we don't need?

apps/os/backend/agent/mcp/mcp-oauth-provider.ts
```diff
@@ -1,51 +1,301 @@
...
+        },
+        createdAt: new Date(),
+        updatedAt: new Date(),
+      });
+    }
+  }
+
+  async saveClientInformation(_info: unknown): Promise<void> {
+    const info = DynamicClientInfo.parse(_info);
+    this.clientId = info.client_id;
```
mmkal, line 195: seems like this should be the only place where we set clientId

mmkal: and we should document who sets clientId when on the field declaration

apps/os/backend/auth/integrations.ts
```diff
@@ -73,6 +76,148 @@ export const integrationsPlugin = () =>
...
+          method: "GET",
+          query: z.object({
+            code: z.string(),
+            state: z.string().optional(),
+            error: z.string().optional(),
+            error_description: z.string().optional(),
+          }),
+        },
+        async (ctx) => {
+          const { code, state, error, error_description } = ctx.query;
```
mmkal: alias to stateId

apps/os/backend/agent/mcp/mcp-event-hooks.ts
zakhardavydov: Not always required

apps/os/backend/auth/integrations.ts
```diff
@@ -73,6 +76,148 @@ export const integrationsPlugin = () =>
...
+          let stateData: z.infer<typeof MCPOAuthState>;
+
+          if (state) {
+            const stateValue = await ctx.context.internalAdapter.findVerificationValue(state);
+            if (!stateValue) {
+              return ctx.json({ error: "Invalid or expired state" }, { status: 400 });
+            }
+
+            try {
+              stateData = MCPOAuthState.parse(JSON.parse(stateValue.value));
```
mmkal: again let's use safeParse

apps/os/backend/auth/integrations.ts
```diff
@@ -73,6 +76,142 @@ export const integrationsPlugin = () =>
...
+              {
+                error: "Missing state parameter",
+                details:
+                  "The MCP OAuth provider did not return the state parameter. This violates OAuth 2.0 security standards.",
+                code,
+                helpUrl: "https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2",
+              },
+              { status: 400 },
+            );
+          }
```
mmkal: can we move this to an early return so the happy path stands out more clearly

apps/os/backend/auth/integrations.ts
```diff
@@ -73,6 +76,142 @@ export const integrationsPlugin = () =>
...
+              organization: {
+                with: {
+                  members: true,
+                },
+              },
+            },
+          });
+
+          if (!estate) {
+            return ctx.json({ error: "Estate not found" }, { status: 404 });
```
mmkal, line 155: we should log an error here, seems weird we'd get an estateId that doesn't exist.

apps/os/backend/auth/integrations.ts
```diff
@@ -73,6 +76,142 @@ export const integrationsPlugin = () =>
...
+          }
+
+          const isMember = estate.organization.members.some(
+            (member) => member.userId === state.userId,
+          );
+
```
mmkal: can this be in the sql query?

apps/os/backend/agent/iterate-agent.ts
```diff
@@ -530,17 +538,20 @@ export class IterateAgent<Slices extends readonly AgentCoreSlice[] = CoreAgentSl
           const mcpEvent = event as Extract<typeof event, { type: MCPRelevantEvent }>; // ideally typescript would narrow this for us but `.includes(...)` is annoying/badly implemented. ts-reset might help
           this.ctx.waitUntil(
             (async () => {
+              if (!this.databaseRecord) {
+                throw new Error("Database record not found");
+              }
```
mmkal:
```suggestion
```

apps/os/backend/agent/iterate-agent.ts
```diff
@@ -588,6 +599,14 @@ export class IterateAgent<Slices extends readonly AgentCoreSlice[] = CoreAgentSl
         //     // }),
         //   // );
       },
+      lazyConnectionDeps: {
+        getDurableObjectInfo: () => this.hydrationInfo,
+        getEstateId: () => this.databaseRecord?.estateId || "",
```
mmkal:
```suggestion
        getEstateId: () => this.databaseRecord.estateId,
```

apps/os/backend/agent/iterate-agent.ts
```diff
@@ -142,6 +142,10 @@ export class IterateAgent<Slices extends readonly AgentCoreSlice[] = CoreAgentSl
     return env.ITERATE_AGENT;
   }
 
+  static getClassName(): string {
```
mmkal: why do we need this?

apps/os/backend/agent/mcp/mcp-oauth-provider.ts
```diff
@@ -1,51 +1,263 @@
...
+            providerId: this.providerId,
+            userId: this.params.userId,
+            accessToken: tokens.access_token,
+            refreshToken: tokens.refresh_token,
+            accessTokenExpiresAt: expiresAt,
+            scope: "",
+          })
+          .returning();
+
+        this.dbAccountId = newAccount.id;
```
mmkal: this should go at the end of the transaction. if we rollback because of an error in the next query or later in the transaction, the field will be wrongly set to a value that doesn't really exist int he db. in facetwe should probably do something along the lines of

```ts
this.dbAccountId = await this.params.db.transaction(async tx => {
  const [newAccount] = await tx.insert(...)
  await tx.insert(...)
  return newAccount.id
})
```

apps/os/backend/agent/mcp/mcp-slice.ts
```diff
@@ -95,6 +95,13 @@ const mcpConnectRequestFields = {
     requiresAuth: z.boolean().default(true),
     triggerLLMRequestOnEstablishedConnection: z.boolean().default(false),
     headers: z.record(z.string(), z.string()).optional(),
+    reconnect: z
+      .object({
+        id: z.string(),
+        oauthClientId: z.string().optional(),
+        oauthCode: z.string().optional(),
```
mmkal: similar to my comment on the constructor
```suggestion
        oauthClientId: z.string(),
        oauthCode: z.string(),
```

PR #66, opened 2025-09-23
remove concept of "user's request" and "completely resolving the request" anchor to helping users achieve their goal

fnloza:
### TL;DR

Updated the Slack agent prompt with clearer guidelines for tool usage, improved communication rules, and fixed typos.

### What changed?

- Refined the agent turn rules to focus on helping users achieve their goals
- Added explicit instructions for handling cases when required tools aren't available
- Standardized capitalization of "Slack" throughout the document
- Fixed typo in tool name from `getURLContent` to `getUrlContent`
- Improved guidance on when to end a turn and yield back to users
- Enhanced instructions for connecting to remote MCP servers
- Clarified image generation guidelines, especially for emoji creation
- Fixed minor grammatical errors and improved phrasing throughout


apps/os/backend/agent/default-context-rules.ts
```diff
@@ -16,46 +16,52 @@ const defaultSlackAgentPrompt = dedent`
   # The agent loop
   You are an AI agent that operates in a loop. 
 
-  A loop takes place in the context of a slack thread and has alternating "agent turns" and "user turns".
+  A loop takes place in the context of a Slack thread and has alternating "agent turns" and "user turns".
```
jonastemplestein:
```suggestion
  The agent loop takes place in the context of a Slack thread and has alternating "agent turns" and "user turns".
```

apps/os/backend/agent/default-context-rules.ts
```diff
@@ -16,46 +16,52 @@ const defaultSlackAgentPrompt = dedent`
...
   ### Communication Rules:
   - greet casually on first message only
   - Tone: lowercase, casual, conversational
-  - Use emojis like in your messages to aid visual interpretation 🎉✅⏳🔴⚡📋🎯, but do sparingly- don't overuse them.
-  - You are addressing a colleague/ group of colleagues in slack -- use direct address like "you", never refer to them in the third person like "user" or "users"
+  - Use emojis like in your messages to aid visual interpretation 🎉✅⏳🔴⚡📋🎯, but do sparingly - don't overuse them.
+  - You are addressing a colleague/ group of colleagues in Slack -- use direct address like "you", never refer to them in the third person like "user" or "users"
   - If you are addressing the entire team / making an announcement, use "we" to refer to the team.
-  - Never pretend hallucinate things you don't know or parameters you don't have. Always be honest.
+  - Never pretend to hallucinate things you don't know or parameters you don't have. Always be honest.
```
jonastemplestein:
```suggestion
  - Never make up or guess facts or function tool parameters. Always be honest.
```

apps/os/backend/agent/default-context-rules.ts
```diff
@@ -66,7 +72,7 @@ const defaultSlackAgentPrompt = dedent`
   - Prefer inline links like "<URL|this image> is cool". Don't do <URL|click here to open>"
   - Mentions: <@user_id>
   - Never use: Markdown tables (use lists/bullets)
-  - Use the getURLContent tool to retrieve the contents of slack messages that users link to (including the entire history of the linked thread)
+  - Use the getURLContent tool to retrieve the contents of Slack messages that users link to (including the entire history of the linked thread)
```
jonastemplestein, line 103: If you change it to `getUrlContent` in one place, it should be consistent here

That said, my PR changes it to be `getURLContent` so you should use that consistently

estates/iterate/rules/general.md
```diff
@@ -5,23 +5,23 @@ match: always()
...
+We use github for version control, whenever you're asked to do something with Github, connect to to the github MCP with the following parameters
 
-   serverUrl: "https://api.githubcopilot.com/mcp/",
-      mode: "company",
-      integrationSlug: "github",
-      allowedTools: [
-        // Core user and authentication
-        "get_me",
-        "search_users",
+serverUrl: "https://api.githubcopilot.com/mcp/",
```
jonastemplestein, line 18: This seems broken?

FWIW I also think the github MCP should be avoided at all costs - it's not a good example unless you heavily heavily filter the set of tools

estates/iterate/rules/how-we-use-github.md
```diff
@@ -6,5 +6,5 @@ match: hasMCPConnection("https://api.githubcopilot.com/mcp/")
 # How We Use Github
 
 - Results limit: All PR queries return only the top 5 most recent PRs (sorted by creation date, newest first). Never show more than 10 PRs at a time.
-- Our monorepo is "os", and our own estate repo is: "iterate" 
-- Links & style: Link PRs as <{pr.html_url}|#{pr.number}>. Use bullets, include key counts and simple latency metrics; keep it brief.
\ No newline at end of file
+- Our monorepo is "os", and our own estate repo is: "iterate"
```
jonastemplestein:
```suggestion
- Our entire company is represented by a single monorepo iterate-com/iterate
```

estates/iterate/rules/general.md
fnloza: but I think customers want to use it / are likely to try this which is why I added it here and not in default context rules- are you saying we should use the integration instead of MCP for github stuff? 

fwiw I think the MCP works if you allowlist a subset of tools, I haven't pressure tested yet - hence why I added it here.

jonastemplestein: Sorry it was a bad comment - I meant you broke the indentation

estates/iterate/rules/reminders.md
```diff
@@ -1,6 +1,6 @@
-<!-- ---
-slug: general
-match: always()
+---
+slug: reminders_draft
+match: never()
```
jonastemplestein: where is this coming from? i don't thinkthis is used for any purpose?

apps/os/backend/agent/default-context-rules.ts
```diff
@@ -16,46 +16,72 @@ const defaultSlackAgentPrompt = dedent`
...
 
-  Once you end your turn, it's the user's turn. When they respond, you will be woken up and given a slack webhook representing the action the user took - generally either sending you a message or reacting with an emoji.
+  Once you end your turn, it's the user's turn. When they respond, you will be woken up and given a Slack webhook representing the action the user took - generally either sending you a message or reacting with an emoji.
 
   ## Your agent turn
   When it is your turn, review available context and conversation history and decide what tools to call and whether or not to end your turn. sendSlackMessage() is by far the most important tool. 
   - sendSlackMessage({ text}) // you will use this MOST of the time to send a message to the user. 
   - sendSlackMessage({ text, endTurn: true}) // set endTurn: true when you're ready to end your turn and yield to the user
 
   ### Your turn rules
```
mmkal: "Your turn rules" is weird

apps/os/backend/agent/default-context-rules.ts
```diff
@@ -16,46 +16,72 @@ const defaultSlackAgentPrompt = dedent`
...
-  - Please keep going until the query is completely resolved, before ending your turn and yielding back to the user.
-  - Only end your turn when:
-    - you are sure that the query is resolved and you have shared evidence of completion, e.g the relevant output (links, images, answers, etc.)
-    - you need input from the user in order to progress with your task
-    - you have tried multiple approaches and are still stuck on your 3rd attempt at handling this particular query
-  - Autonomously resolve the query to the best of your ability, using the tools available to you, before coming back to the user. Do NOT guess or make up an answer.
+  - Review available context to understand what the user(s) is(are trying to achieve.
```
mmkal: What motivated this one? I would have thought this is the default behaviour of an LLM, when did we see it not do this?

apps/os/backend/agent/default-context-rules.ts
```diff
@@ -16,46 +16,72 @@ const defaultSlackAgentPrompt = dedent`
...
 
   ### Your turn rules
-  - Please keep going until the query is completely resolved, before ending your turn and yielding back to the user.
-  - Only end your turn when:
-    - you are sure that the query is resolved and you have shared evidence of completion, e.g the relevant output (links, images, answers, etc.)
-    - you need input from the user in order to progress with your task
-    - you have tried multiple approaches and are still stuck on your 3rd attempt at handling this particular query
-  - Autonomously resolve the query to the best of your ability, using the tools available to you, before coming back to the user. Do NOT guess or make up an answer.
+  - Review available context to understand what the user(s) is(are trying to achieve. 
+  - Keep trying to help the user(s) achieve their goal using the tools available to you, before ending your turn and yielding back to the user.
```
mmkal: "Keep trying" doesn't really make clear when it's a good idea to keep going, the below is a good list of when it's not though

apps/os/backend/agent/default-context-rules.ts
```diff
@@ -16,46 +16,72 @@ const defaultSlackAgentPrompt = dedent`
...
+    - you have tried multiple approaches and are still stuck on your 3rd attempt at helping the user(s) achieve their goal
+  - Always be honest about what you can and can't do, and assess whether you can help them achieve their goal given the tools available to you BEFORE making any promises to the user(s). 
 
   ### Calling Tools 
-  - You should make use of parallel tool calls as much as possible
-  - Whenever you call one or more tools, you should in parallel call sendSlackMessage with a brief description of what you're doing. This must be in italics. For example
-  - Use only the allowed tools made available in your environment.
-  - After each tool call or code edit, provide a 1-2 line validation of the result before proceeding or self-correcting if needed.
+  - You should make use of parallel tool calls as much as possible, unless a series of actions can only be performed in sequence (ie. you need to get the result of one tool call to be able to call the next tool).
+  - Whenever you call one or more tools (other than sendSlackMessage), in parallel call sendSlackMessage with a brief description of what you're doing. This must be in italics.
```
mmkal:
```suggestion
  - Whenever you call one or more tools, in parallel call sendSlackMessage with a brief description of what you're doing. This must be in italics. 
```
pure guess but gut feeling is the parenthetical is just confusing here

apps/os/backend/agent/default-context-rules.ts
```diff
@@ -16,46 +16,72 @@ const defaultSlackAgentPrompt = dedent`
...
+  - Always be honest about what you can and can't do, and assess whether you can help them achieve their goal given the tools available to you BEFORE making any promises to the user(s). 
 
   ### Calling Tools 
-  - You should make use of parallel tool calls as much as possible
-  - Whenever you call one or more tools, you should in parallel call sendSlackMessage with a brief description of what you're doing. This must be in italics. For example
-  - Use only the allowed tools made available in your environment.
-  - After each tool call or code edit, provide a 1-2 line validation of the result before proceeding or self-correcting if needed.
+  - You should make use of parallel tool calls as much as possible, unless a series of actions can only be performed in sequence (ie. you need to get the result of one tool call to be able to call the next tool).
+  - Whenever you call one or more tools (other than sendSlackMessage), in parallel call sendSlackMessage with a brief description of what you're doing. This must be in italics. 
+  - After each tool call, provide a 1-2 line validation of the result before proceeding or self-correcting if needed.
```
mmkal: what does a 1-2 line validation look like?

apps/os/backend/agent/default-context-rules.ts
```diff
@@ -16,46 +16,72 @@ const defaultSlackAgentPrompt = dedent`
...
+  - Otherwise, if you are not making any tool calls (other than sendSlackMessage) --> respond immediately with one concise message.
```
mmkal:
```suggestion
  - Otherwise, if you are not making any tool calls --> respond immediately with one concise message. 
   For example, if: 
```

apps/os/backend/agent/default-context-rules.ts
```diff
@@ -16,46 +16,72 @@ const defaultSlackAgentPrompt = dedent`
...
   - sendSlackMessage({text: "_fetching..."})
-  Second LLMresponse in agent turn (after getURLContent complete):
+  Second LLM response in agent turn (after getURLContent complete):
   - sendSlackMessage({ text: "fetched the contents of the URL. here they are: ...", endTurn: true }) 
 
+
+  Examples: responding immediately - if you are not making any tool calls 
+  \`\`\`js
+  // Quick answer 
+  sendSlackMessage({ text: "4", endTurn: true }) // user: "what's 2+2?"
```
mmkal: this seems very confusing that the answer is before the question
```suggestion
  // user: "what's 2+2?"
  // result:
  sendSlackMessage({ text: "4", endTurn: true })
```

apps/os/backend/agent/default-context-rules.ts
```diff
@@ -16,46 +16,72 @@ const defaultSlackAgentPrompt = dedent`
...
+  sendSlackMessage({ text: "Paris", endTurn: true }) // user: "what's the capital of France?"
+
+  // Casual content
+  sendSlackMessage({ text: "why did the developer go broke?\n\nbecause they used up all their cache.", endTurn: true }) // user: "tell me a joke"
+
+  // Clarifying question
+  sendSlackMessage({ text: "got it — which repo are you talking about?", endTurn: true }) // user: "fix the CI please"
+
+  // Something you can't help with 
+  sendSlackMessage({ text: "I can't pick up groceries directly, would you like me to setup a reminder for you?", endTurn: true }) // user: "pick up some groceries on the way home"
```
mmkal: same with these

estates/iterate/rules/reminders.md
mmkal: nope would be great if we did but we don't parse frontmatter right now

fnloza: I was playing around w these last week, but was blocked on testing so it got stuck in draft. 

I was porting over the rules yesterday, had to fix the helper function in the process-- which Nick blow needed, so we agreed to merge and then tinker. 

@mmkal what do you mean by not parsing front-matter right now? as in not use matchers?

apps/os/backend/agent/default-context-rules.ts
fnloza: updated - but it's not strictly true (it will always need to make a tool call) (at least sendSlackMessage)

apps/os/backend/agent/default-context-rules.ts
fnloza: it works! but I updated 👍

apps/os/backend/agent/default-context-rules.ts
fnloza: deleted - I was just thinking it helps ground the rest of this.

apps/os/backend/agent/default-context-rules.ts
fnloza: Updated- an interesting one to think about is whether it would be better to display this as a developer message / webhook (as it actually happens in practice). Trade-off is that it's more inefficient 
{
    "user": "U08BU594KGC",
    "text": "what's 2+2?",
    "ts": "1758645495.778819",
    "createdAt": "2025-09-23T16:38:15.778Z"
  }

apps/os/backend/agent/default-context-rules.ts
fnloza: I had to add it back - it's stopping it from sending duplicate messages, sort of guides it to make a decision based on existing context. 

https://iterate-com.slack.com/archives/C08R1SMTZGD/p1758711014818059?thread_ts=1758711010.382759&cid=C08R1SMTZGD

PR #68, opened 2025-09-23
re-instate web search, get url contents and image sharing

jonastemplestein:
- Improved file sharing developer messages in agent-core
- Re-instated `webSearch` and `getURLContent` functionality tools (moved from slack agent to iterate agent)
- Re-instated file sharing 1) from user to LLM via slack upload and 2) from function tool output (e.g. screenshot of a webpage) to LLM and slack
- Fixed bug in getURLContent when a slack link at the root of a thread was shared
- Reduced the degrees of freedom in webSearch so that it only takes a query. Configured it to automatically produce a context string optimised for LLM consumption from exa searches

### DIFF
--- a/apps/os/backend/default-tools.ts
+++ b/apps/os/backend/default-tools.ts
@@ -196,9 +196,7 @@ const ContentsRequestInput = z
     urls: z.array(z.string().url()).describe("Array of URLs to extract content from"),
     ids: z
       .array(z.string())
-      .describe(
-        "Deprecated - use 'urls' instead. Array of document IDs from previous searches",
-      )
+      .describe("Deprecated - use 'urls' instead. Array of document IDs from previous searches")
       .optional(),
   })
   .merge(ContentsOptions);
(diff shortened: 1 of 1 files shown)


apps/os/backend/default-tools.ts
```diff
@@ -0,0 +1,793 @@
...
+// - this should format webhooks consistently with slack-slice
+// - this should upload files to the iterate file system that are shared on that thread
+//
+async function getURLContentFromSlack(params: { url: string; db: DB }) {
+  try {
+    const { threadTs, channelId } = extractDetailsFromSlackLink(params.url);
+    const { ts } = extractDetailsFromSlackLink(params.url);
+    console.log("threadTs", threadTs);
+    console.log("ts", ts);
+    console.log("channelId", channelId);
```
zakhardavydov:
```suggestion
```

Delete it

apps/os/worker-configuration.d.ts
zakhardavydov, line 1: Shouldn't this be gitignored?

jonastemplestein: Why?

PR #70, opened 2025-09-23
port braintrust and posthog tracing

myst-6:
Braintrust:
- ~~Removed openai to braintrust conversion, because they fixed the issue.~~ - How silly I was to think that it was fixed.
- Create agent span in the background if not injected.
- The `setBraintrustParentSpanExportedId` etc. evals plumbing is there for later.
- There is a brain icon on the agent details page where you can instantly go to the braintrust trace.

Posthog:
- Removed a lot of vibe coded fluff, but if I removed something critical please tell me and I can restore it.
- Opted not to use the default posthog wrapper for now, because it is a bit clunky to get working (its not a 'wrapper' it's actually a completely different class with a few different method signatures..) and the current one is not that long and works well.
- Analytics and traces both back.

Resolves [ITE-2957: bring back posthog tracing and analytics events](https://linear.app/iterate-com/issue/ITE-2957/bring-back-posthog-tracing-and-analytics-events) and [ITE-2905: Re-Introduce Agent Braintrust Spans (Faster)](https://linear.app/iterate-com/issue/ITE-2905/re-introduce-agent-braintrust-spans-faster)

### DIFF
--- a/apps/os/app/components/agent-reduced-state.tsx
+++ b/apps/os/app/components/agent-reduced-state.tsx
@@ -1,7 +1,6 @@
 import { useMemo, useState } from "react";
 import jsonata from "jsonata/sync";
 import type { AugmentedCoreReducedState } from "../../backend/agent/agent-core-schemas.ts";
-import { cn } from "../lib/utils.ts";
 import { SerializedObjectCodeBlock } from "./serialized-object-code-block.tsx";
 import { AutoComplete } from "./autocomplete.tsx";
 
--- a/apps/os/backend/agent/iterate-agent.ts
+++ b/apps/os/backend/agent/iterate-agent.ts
@@ -872,7 +872,7 @@ export class IterateAgent<Slices extends readonly AgentCoreSlice[] = CoreAgentSl
 
   // get the braintrust parent span exported id from the state
   // if it's not set, create it and set it in the state
-  async getOrCreateBraintrustParentSpanExportedId(estateName?: string) {
+  async getOrCreateBraintrustParentSpanExportedId(estateName: string) {
     if (this.state.braintrustParentSpanExportedId) {
       return this.state.braintrustParentSpanExportedId;
     } else {
--- a/apps/os/backend/utils/utils.ts
+++ b/apps/os/backend/utils/utils.ts
@@ -26,19 +26,14 @@ export function ensureString(value: unknown): string {
 export type DistributiveOmit<T, K extends PropertyKey> = T extends unknown ? Omit<T, K> : never;
 
 export function getEnvironmentName(env: {
-  ITERATE_USER?: string;
+  ITERATE_USER: string;
   STAGE__PR_ID?: string;
-  ESTATE_NAME?: string;
+  ESTATE_NAME: string;
 }) {
   const { STAGE__PR_ID, ITERATE_USER, ESTATE_NAME } = env;
   if (STAGE__PR_ID) {
     return `pr-${STAGE__PR_ID}`;
   }
-  if (ITERATE_USER) {
-    return `local-${ITERATE_USER}`;
-  }
-  if (ESTATE_NAME) {
-    return `estate-${ESTATE_NAME}`;
-  }
-  throw new Error("Failed to resolve stage");
+  const isDev = !!ITERATE_USER;
+  return isDev ? `local-${ITERATE_USER}` : `estate-${ESTATE_NAME}`;
 }
--- a/apps/os/env.ts
+++ b/apps/os/env.ts
@@ -16,7 +16,6 @@ export type CloudflareEnv = Env & {
   GITHUB_APP_SLUG: string;
   EXPIRING_URLS_SIGNING_KEY: string;
   GITHUB_WEBHOOK_SECRET: string;
-  ITERATE_USER?: string;
   STAGE__PR_ID?: string;
   EXA_API_KEY: string;
   CLOUDFLARE_API_TOKEN: string;


apps/os/backend/agent/braintrust-wrapper.ts
```diff
@@ -306,18 +59,26 @@ export async function logLLMRequestToBraintrust(
...
+  openai: OpenAI;
+  getBraintrustParentSpanExportedId?: () => Promise<string>;
+  waitUntil: (promise: Promise<void>) => void;
+}) {
   // proxy the openai.responses.stream method
   return new Proxy(openai, {
     get(target, prop) {
       if (prop !== "responses") {
-        // @ts-expect-error - prop is a string
+        // @ts-expect-error
```
mmkal: let's cast `target` to something like `Record<string | number | symbol, unknown>` so we don't have to put ts-expect-error everywhere

apps/os/backend/agent/openai-client.ts
```diff
@@ -1,41 +1,28 @@
...
-  braintrustParentSpanExportedId?: string;
+  posthog: {
+    estateName: string;
+    environmentName: string;
+    traceId: string;
+  };
+  braintrust?: {
+    projectName?: string;
+    getBraintrustParentSpanExportedId?: () => Promise<string>;
+  };
```
mmkal, line 26: can we drop all the optionals here? Obviously openai does work without braintrust, but in our codebase do we have anywhere where we create a client without braintrust? 

even more so the optionals within the braintrust config - let's make them required just to reduce the amount of `?.()` and `??` we need to put everywhere

apps/os/backend/agent/openai-client.ts
```diff
@@ -57,20 +44,28 @@ export async function openAIProvider(opts: {
...
 
   // Apply Braintrust wrapper on top
-  if (braintrustKey) {
-    // Note: braintrustParentSpanExportedId is created and cached in agent state in apps/platform/backend/worker.ts
-    // The Span is created with the Durable Object Name, just like trace.traceId, and so traceId is not needed here.
-    getBraintrustLogger({ braintrustKey, projectName: opts.projectName || "iterate" }); // ensure logger is initialized
-    openai = braintrustOpenAIWrapper(openai, opts.braintrustParentSpanExportedId);
+  if (braintrustKey && opts.braintrust) {
+    // note: this second import is cached if the first one happened already
+    const waitUntil = await import("cloudflare:workers")
```
mmkal: what's the reason for importing like this rather than at the top level?

mmkal: oh and it looks like we do this in two places - could move to a helper

apps/os/backend/agent/posthog-event-processor.ts
```diff
@@ -0,0 +1,110 @@
...
+}: {
+  posthog: OSPosthog;
+  data: SliceEventData;
+}) {
+  const event = data.event;
+  const coreEventTypeLower = event.type.toLowerCase() as Lowercase<
+    MergedEventForSlices<SlackAgentSlices>["type"]
+  >;
+
+  // TODO: decide what we need to actually track
```
mmkal: this is a looot of data in the meantime. probably fine, but seems extreme

apps/os/backend/agent/posthog-event-processor.ts
```diff
@@ -0,0 +1,110 @@
...
+
+        if (!participant) {
+          throw new Error(`Participant not found for user ${event.user}`);
+        }
+
+        const [internalUserId, participantData] = participant;
+        posthog.identify(internalUserId, {
+          type: "user",
+          email: participantData.email ?? null,
+        });
```
mmkal: in the browser `posthog.identify(...)` sets the user id to whatever you pass it, not sure how it works in node but not what I'd expect a function called `getDistinctIdForSliceEventData` to do

let's move all the posthog.identifys out of this, and call it from outside once we return the `id` (or not call it, and instead pass `user_id` to posthog.track which I assume there's an ok way of doing?

apps/os/backend/utils/braintrust-client.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export async function makeBraintrustSpan(opts: {
+  braintrustKey: string;
+  projectName: string;
+  spanName: string;
+}) {
+  const { braintrustKey, projectName, spanName } = opts;
+  const logger = getBraintrustLogger({ braintrustKey, projectName });
+  const span = logger.startSpan({ name: spanName, type: "task" });
+  await span.flush();
```
mmkal, line 19: how long does this normally take?

apps/os/backend/agent/braintrust-wrapper.ts
myst-6: they are all different 'target's but ill just cast prop to keyof typeof prop since it is

apps/os/backend/agent/openai-client.ts
myst-6: in the agent the function is used but not projectName
in evals the projectName is provided but not the function
in the file upload handler we are just uploading to openai without wanting braintrust
so all of those question marks are needed imo

apps/os/backend/agent/posthog-event-processor.ts
myst-6: true - once we know what we want we can definitely cut it down

apps/os/backend/utils/braintrust-client.ts
myst-6: I don't remember, but I think it was a big contributing factor to the agent's latency before it was put in the background.
Why do you ask?

apps/os/backend/agent/posthog-event-processor.ts
myst-6: 👍

apps/os/backend/agent/openai-client.ts
myst-6: not sure, I thought it was put intentionally by whoever wrote the posthog wrapper before, but maybe it was vibe coded in when somebody didn't know you could do .cf.test.ts

either way it works as a top-level import so it's fine

apps/os/backend/agent/braintrust-wrapper.ts
```diff
@@ -306,19 +65,25 @@ export async function logLLMRequestToBraintrust(
...
  * @returns A new openai client that is wrapped in braintrust
  */
-export function braintrustOpenAIWrapper(openai: OpenAI, braintrustParentSpanExportedId?: string) {
+export function braintrustOpenAIWrapper({
+  openai,
+  getBraintrustParentSpanExportedId,
+  waitUntil,
+}: {
+  openai: OpenAI;
+  getBraintrustParentSpanExportedId?: () => Promise<string>;
```
mmkal: I think this and the many similar /related ones can be non optional

apps/os/backend/agent/braintrust-wrapper.ts
```diff
@@ -306,19 +65,25 @@ export async function logLLMRequestToBraintrust(
...
+  waitUntil: (promise: Promise<void>) => void;
+}) {
   // proxy the openai.responses.stream method
   return new Proxy(openai, {
     get(target, prop) {
       if (prop !== "responses") {
-        // @ts-expect-error - prop is a string
-        return target[prop];
+        return target[prop as keyof typeof target];
```
mmkal, line 81: Can we do one cast in one place

apps/os/backend/utils/staging.ts
```diff
@@ -0,0 +1,17 @@
+export function getStage(env: {
```
mmkal: I don't think we should import from staging.ts in production. Also this function is confusing - what does "stage" mean here?

apps/os/backend/agent/braintrust-wrapper.ts
mmkal: oh maybe we can't, based on your other comment. i might try checking this out and pushing something

apps/os/backend/agent/braintrust-wrapper.ts
mmkal: as in you can just find and replace `getBraintrustParentSpanExportedId?` to `getBraintrustParentSpanExportedId` and I am pretty sure everything will compile (and run?)

mmkal: pushed c556e3c572c2d3c9b0e1453967f0c389cc90912f - can be reverted if necessary but types were happy

myst-6: oh yeh thanks

apps/os/backend/utils/staging.ts
myst-6: Oh sorry stage is the wrong word - I meant getEnvironment or similar probably 

I want to identify staging/production/development, and give each one a unique string depending on the (A) iterate user in dev (B) pr number in staging (C) estate name in prod. What do you think we should call this function (and the file / should we move it somewhere else?)

apps/os/backend/utils/braintrust-client.ts
zakhardavydov: Up to couple of seconds. I have seen it just throwing exception for example too

apps/os/backend/utils/staging.ts
myst-6: i called it getEnvironmentName, if it's really bad we can make another PR to fix it

mmkal (approved):

PR #74, opened 2025-09-23
Feature: participant management and slack provider user sync

zakhardavydov:
- Implemented Slack participant management - Restored functionality to add users as participants when they join conversations
- Implemented Slack participant mentions - Restored functionality to add users as mentioned participants when they're @ mentioned
- Refactored Slack OAuth integration - Switched from manual HTTP calls to Slack WebClient for OAuth token exchange
- Added automatic Slack user sync - Created background sync to import all Slack workspace users into the database
- Created `slack-utils.ts` - New utility file for Slack user mapping and synchronization functions

### DIFF
--- a/apps/os/app/routes/estate.tsx
+++ b/apps/os/app/routes/estate.tsx
@@ -281,7 +281,7 @@ function EstateContent() {
         estateId: estateId!,
       });
       window.location.href = installationUrl;
-    } catch (_error) {
+    } catch {
       toast.error("Failed to start GitHub connection flow");
     }
   };


apps/os/backend/auth/slack-utils.ts
jonastemplestein, line 1: Do we really need another slack-utils file? I think we must have 4 or something

I hope we can move the other way and over time move all slack related stuff further together

That way, when we add discord support or whatever, we have a blueprint

apps/os/backend/auth/integrations.ts
```diff
@@ -359,61 +338,62 @@ export const integrationsPlugin = () =>
...
-          params.set("redirect_uri", redirectURI);
-
-          const tokenRes = await fetch(`https://slack.com/api/oauth.v2.access`, {
-            method: "POST",
-            body: params.toString(),
-            headers: {
-              Authorization: `Basic ${httpBasicAuth}`,
-              "Content-Type": "application/x-www-form-urlencoded",
-            },
+          const webClient = new WebClient();
```
jonastemplestein:
```suggestion
          const slackAPI = new WebClient();
```

or `slackClient` or something

apps/os/backend/auth/slack-utils.ts
```diff
@@ -0,0 +1,90 @@
...
+      providerMetadata: member,
+    });
+  });
+}
+
+export async function syncSlackUsersInBackground(botToken: string) {
+  const authedWebClient = new WebClient(botToken);
+  const userListResponse = await authedWebClient.users.list({});
+  if (userListResponse.ok && userListResponse.members) {
+    const db = getDb();
```
jonastemplestein: This creates a new connection to hyperdrive (as far as we understand - which is not very well at all)

In other places I've carried around a reference to a `db: DB` client everywhere to make sure we reuse it. But it's possible that this is actually not necessary

Is v hard to say without proper tracing and metrics (or talking to the cloudflare engineers)

apps/os/backend/auth/slack-utils.ts
```diff
@@ -0,0 +1,90 @@
+import { eq, and } from "drizzle-orm";
+import { WebClient, type UsersListResponse } from "@slack/web-api";
+import * as schema from "../db/schema.ts";
+import { getDb } from "../db/client.ts";
+
+export async function saveSlackUserMapping(
```
mmkal: @jonastemplestein @BlankParticle this is the function I was talking about on slack where we're having to do lots of complex logic because we can't just say "hey we have a new provider-profile" (check if mapping exists, if it does update the user table and then the mapping, otherwise check if a user exists, if it does update the user then insert the mapping, otherwise insert the user then the mapping)

mmkal: side note, Zak and I didn't know how to express all this in a single query without dropping down to raw sql, even less so when it's a batch operation with a list of users

zakhardavydov: https://github.com/drizzle-team/drizzle-orm/discussions/2557

zakhardavydov: Can we merge this? It is blocking a lot of things and the current implementation is reasonable

apps/os/app/routes/estate.tsx
```diff
@@ -281,7 +281,7 @@ function EstateContent() {
         estateId: estateId!,
       });
       window.location.href = installationUrl;
-    } catch (error) {
+    } catch (_error) {
```
mmkal:
```suggestion
    } catch {
```

apps/os/backend/agent/agent-core.ts
```diff
@@ -194,7 +194,7 @@ export interface AgentCoreSlice<Spec extends AgentCoreSliceSpec = AgentCoreSlice
   eventSchema: SliceEventSchemaOf<Spec>;
   /** Zod schema for input events (defaults to `eventSchema`) */
   eventInputSchema: SliceEventInputSchemaOf<Spec>;
-  /** Async reducer returning partial state updates */
+  /** Reducer returning partial state updates */
```
mmkal, line 197: 👍

apps/os/backend/agent/slack-agent.ts
```diff
@@ -286,144 +286,136 @@ export class SlackAgent extends IterateAgent<SlackAgentSlices> implements ToolsI
...
-    // ];
+    if (slackUserId === botUserId) {
+      return [];
+    }
+    const currentState = this.agentCore.state;
+
+    if (!currentState.participants) {
+      return [];
+    }
+
```
mmkal:
```suggestion
```

apps/os/backend/agent/slack-agent.ts
```diff
@@ -286,144 +286,136 @@ export class SlackAgent extends IterateAgent<SlackAgentSlices> implements ToolsI
...
+    if (currentState.participants[userMapping.internalUser.id]) {
+      return [];
+    }
+
+    const { internalUser, externalId, providerMetadata } = userMapping;
+
+    return [
+      {
+        type: "CORE:PARTICIPANT_JOINED" as const,
```
mmkal, line 326: you don't need all these `as const` when you have one at the bottom. but you should just use `satisfies AgentCoreEvent` instead anyway, that'll actually give helpful errors in the right place. or a return type of ParticipantJoinedEvent

apps/os/backend/agent/slack-agent.ts
```diff
@@ -286,144 +286,136 @@ export class SlackAgent extends IterateAgent<SlackAgentSlices> implements ToolsI
...
-    //     metadata: {},
-    //   }),
-    // );
+    const mentionedUserIds = getMentionedExternalUserIds(messageText);
+    const currentState = this.agentCore.state;
+
+    if (mentionedUserIds.length === 0) {
+      return [];
+    }
+
```
mmkal, line 362:
```suggestion
```

apps/os/backend/agent/slack-agent.ts
```diff
@@ -286,144 +286,132 @@ export class SlackAgent extends IterateAgent<SlackAgentSlices> implements ToolsI
...
+        internalUser: true,
+      },
+    });
+
+    return userMappings
+      .filter((userMapping) => userMapping.internalUser != null)
+      .map((userMapping): ParticipantMentionedEventInput => {
+        const { internalUser, externalId, providerMetadata } = userMapping;
+        return {
+          type: "CORE:PARTICIPANT_MENTIONED" as const,
```
mmkal, line 396: again let's use satisfies or a proper type

PR #77, opened 2025-09-23
Make DO tool factory typesafe

myst-6:
Resolves [ITE-2918: Make DO tool factory typesafe](https://linear.app/iterate-com/issue/ITE-2918/make-do-tool-factory-typesafe)

@mmkal I was debating between some things:
- Make the class name optional and default to something?
- Not even pass a class name, just have all the iterate agent and slack agent tools.

Would be great to hear your opinion! :)

### DIFF
--- a/apps/os/backend/agent/default-context-rules.ts
+++ b/apps/os/backend/agent/default-context-rules.ts
@@ -7,11 +7,9 @@ import { defineRule, matchers } from "./context.ts";
 import { slackAgentTools } from "./slack-agent-tools.ts";
 import { createDOToolFactory } from "./do-tools.ts";
 import { iterateAgentTools } from "./iterate-agent-tools.ts";
-import type { IterateAgent } from "./iterate-agent.ts";
-import type { SlackAgent } from "./slack-agent.ts";
 
-const iterateAgentTool = createDOToolFactory<IterateAgent>()(iterateAgentTools);
-const slackAgentTool = createDOToolFactory<SlackAgent>()(slackAgentTools);
+const iterateAgentTool = createDOToolFactory(iterateAgentTools);
+const slackAgentTool = createDOToolFactory(slackAgentTools);
 
 const defaultSlackAgentPrompt = dedent`
   You are @iterate, a helpful slackbot made by iterate.com.
@@ -203,7 +201,6 @@ export const defaultContextRules = async () => [
         ),
       }),
       iterateAgentTool.generateImage(),
-      iterateAgentTool.editImage(),
       slackAgentTool.sendSlackMessage({
         overrideInputJSONSchema: z.toJSONSchema(
           slackAgentTools.sendSlackMessage.input.pick({
--- a/apps/os/backend/agent/do-tools.ts
+++ b/apps/os/backend/agent/do-tools.ts
@@ -1,8 +1,6 @@
 import { z } from "zod";
 import type { JSONSchema } from "zod/v4/core";
 import type { AgentDurableObjectToolSpec, ToolSpec } from "./tool-schemas.ts";
-import type { SlackAgent } from "./slack-agent.ts";
-import type { IterateAgent } from "./iterate-agent.ts";
 
 export type RuntimeJsonSchema = {
   type: "query" | "mutation" | "subscription";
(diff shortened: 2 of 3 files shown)


apps/os/backend/agent/do-tools.ts
```diff
@@ -88,3 +91,18 @@ export function doToolToRuntimeJsonSchema(_value: unknown) {
   };
   return runtimeJsonSchema;
 }
+
+export type AgentDOToolMethodName<TClassName extends "SlackAgent" | "IterateAgent"> =
+  TClassName extends "SlackAgent" ? keyof typeof slackAgentTools : keyof typeof iterateAgentTools;
+
+export function agentDOTool<TClassName extends "SlackAgent" | "IterateAgent">(
```
mmkal: How I was thinking of it: a function that takes a record of tool defs and gives you a helper you can use like before 

```ts
const doTools = createThing<FooAgent>('FooAgent', definitions)

doTools.myMethod()
doTools.myMethod({
  description: "abc"
})
```

That'd also give cmd-click ability in cursor if types are implemented right

createThing implementation could be a Proxy or just Object.fromEntries

apps/os/backend/agent/default-context-rules.ts
```diff
@@ -5,6 +5,13 @@ import { z } from "zod/v4";
 import { SearchRequest } from "../default-tools.ts";
 import { defineRule, matchers } from "./context.ts";
 import { slackAgentTools } from "./slack-agent-tools.ts";
+import { createDOToolFactory } from "./do-tools.ts";
+import { iterateAgentTools } from "./iterate-agent-tools.ts";
+import type { IterateAgent } from "./iterate-agent.ts";
+import type { SlackAgent } from "./slack-agent.ts";
+
+const iterateAgentTool = createDOToolFactory<IterateAgent>()(iterateAgentTools);
+const slackAgentTool = createDOToolFactory<SlackAgent>()(slackAgentTools);
```
myst-6, line 15: I had to separate it out into two functions, because it's annoying to pass in `typeof iterateAgentTools` as the second type parameter and typescript doesn't infer the other type parameters when you pass in the first one

apps/os/backend/agent/do-tools.ts
```diff
@@ -88,3 +91,35 @@ export function doToolToRuntimeJsonSchema(_value: unknown) {
...
+            > & {
+              passThroughArgs?: Partial<
+                z.output<NonNullable<NonNullable<(typeof definitions)[K]>["input"]>>
+              >;
+            },
+          ) => ToolSpec
+        : never;
+    };
+  };
+}
```
myst-6, line 125: The `as` is for two things:
- Ctrl clicking taking you to the right place
- Pass through args having the correct type (although overrideInputSchema exists so maybe I should remove it.)

myst-6 (commented):
@jonastemplestein @mmkal thoughts?

I think the requirements are:
1. Ctrl+click will take us to the agent definition (in other words the return type is indexed by keyof TAgent) where TAgent is a type param you pass in
2. It only lets you use tools which are defined on some tool definition record which you pass in
3. It verifies that the tool definitions match up with the keys on the agent which you passed in for the type param

jonastemplestein:
I think adding something like this to `apps/os/sdk/index.ts` would do what I want

```ts
export const tools = {
  ...createDOToolFactory<IterateAgent>()(iterateAgentTools),
  ...createDOToolFactory<SlackAgent>()(slackAgentTools),
}
```

apps/os/backend/agent/do-tools.ts
```diff
@@ -88,3 +91,35 @@ export function doToolToRuntimeJsonSchema(_value: unknown) {
   };
   return runtimeJsonSchema;
 }
+
+export function createDOToolFactory<TAgent extends SlackAgent | IterateAgent>() {
```
mmkal: This function shouldn't know about SlackAgent | IterateAgent

mmkal: Discussed in person - it's extending these to enable cmd-click into the method implementations itself. Let's cmd-click into the tool defs instead

myst-6:
Alright I think I'm happy with how it is now.

apps/os/backend/agent/do-tools.ts
```diff
@@ -88,3 +89,24 @@ export function doToolToRuntimeJsonSchema(_value: unknown) {
   };
   return runtimeJsonSchema;
 }
+
+export function createDOToolFactory<T extends ReturnType<typeof defineDOTools>>(definitions: T) {
```
mmkal, line 93: I'd use the same generic as defineDOTools `T extends Record<string, DOToolDef<any, any>>`

mmkal (approved):
looks good - one comment but you can either not do, do and merge or merge and then do

apps/os/backend/agent/do-tools.ts
myst-6: i was originally going to do that but defineDOTools puts some additional stuff on it which makes the type system crash out so ill just leave it like this

PR #84, opened 2025-09-24
feat: Add untils to interact with github app tokens

BlankParticle:
TLDR  
\- Cloning can be done for private repos with installation tokens  
\- Util to print out the temp access token, so we can use them in local `gh`​ and later pass them into containers

### DIFF
--- a/apps/os/sdk/cli/cli-db.ts
+++ b/apps/os/sdk/cli/cli-db.ts
@@ -1,0 +1,22 @@
+/**
+ * Basically the same as the backend/db/client.ts file, but without cloudflare specific stuff.
+ */
+import { drizzle } from "drizzle-orm/postgres-js";
+import postgres from "postgres";
+import * as schema from "../../backend/db/schema.ts";
+
+if (!process.env.DRIZZLE_RW_POSTGRES_CONNECTION_STRING) {
+  throw new Error("DRIZZLE_RW_POSTGRES_CONNECTION_STRING is not set");
+}
+
+const pg = postgres(process.env.DRIZZLE_RW_POSTGRES_CONNECTION_STRING, {
+  max: 5,
+  // If you are not using array types in your Postgres schema, disable `fetch_types` to avoid an additional round-trip (unnecessary latency)
+  fetch_types: false,
+});
+
+export const db = drizzle(pg, { schema, casing: "snake_case" });
+
+export type DB = typeof db;
+
+export * as schema from "../../backend/db/schema.ts";
--- a/apps/os/sdk/cli/commands/checkout-estate.ts
+++ b/apps/os/sdk/cli/commands/checkout-estate.ts
@@ -3,31 +3,70 @@ import * as path from "path";
 import * as fs from "fs/promises";
 import { z } from "zod";
 import { x as exec } from "tinyexec";
-import { EstateSpecifierFromString } from "../estate-specifier.ts";
 import { t } from "../config.ts";
+import { getRepoAccessToken } from "../github-utils.ts";
 
 export const checkoutEstateCommand = t.procedure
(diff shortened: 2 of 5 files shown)


BlankParticle:
* **#84** <a href="https://app.graphite.dev/github/pr/iterate-com/iterate/84?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/iterate-com/iterate/84?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

apps/os/sdk/cli/github-utils.ts
mmkal, line 1: nit - rename to github-utils.ts

apps/os/sdk/cli/commands/checkout-estate.ts
```diff
@@ -25,7 +27,24 @@ export const checkoutEstateCommand = t.procedure
...
     }
     await fs.mkdir(path.dirname(repoPath), { recursive: true });
-    const cloneResult = await exec("git", ["clone", input.estate.cloneUrl, repoPath]);
+    const githubToken = await getRepoAccessToken(input.estateId);
+
+    if (!input.estate.cloneUrl.includes("https://")) {
+      console.warn(
+        `Cloning is only supported for https URLs for now, found ${input.estate.cloneUrl}`,
+      );
+      return;
```
mmkal: let's just throw? this will mess up scripts that do `iterate estate checkout git:blah && somethingelse`

apps/os/sdk/cli/cli-db.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Basically the same as the backend/db/client.ts file, but without cloudflare specific stuff.
+ */
+import { drizzle } from "drizzle-orm/postgres-js";
+import postgres from "postgres";
+import * as schema from "../../backend/db/schema.ts";
+
+if (!process.env.DRIZZLE_RW_POSTGRES_CONNECTION_STRING) {
```
mmkal, line 8: where's this env var coming from? I don't think we should include the "DRIZZLE_" prefix (unless we're using some drizzle-hosted connection pooler or something??)

apps/os/sdk/cli/commands/gh-commands.ts
```diff
@@ -0,0 +1,16 @@
+import { z } from "zod";
+import { t } from "../config.ts";
+import { getRepoAccessToken } from "../git-utils.ts";
+
+export const ghPrintSetup = t.procedure
```
mmkal, line 5: Minor but I think slightly cleaner to do 

```ts
const ghPrintSetup = t.procedure...

export const githubRouter = t.router({
  setup: ghPrintSetup,
  someOtherCommandWeAddSomeday: otherCommand,
})
```

then keep the root router cleaner

```ts
const router = t.router({
  estate: { checkout: checkoutEstateCommand },
  gh: githubRouter,
});
```

mmkal: I probably should have done this with an `estateRouter`. easy to refactor anytime though so not a big deal

apps/os/sdk/cli/cli-db.ts
BlankParticle: this is used in drizzle kit/studio, to run migrations and stuff
Nick asked me to re use this as its already on doppler

apps/os/sdk/cli/commands/checkout-estate.ts
BlankParticle: sure, I just opened the PR, i am going to change the interface for this command

PR #85, opened 2025-09-24
debug ui: use jsonata instead of loads of tabs

mmkal:
more flexible, less to maintain, better for testing out matchers

https://github.com/user-attachments/assets/c4b18175-301a-44c6-8e21-a80cfb4fa943


mmkal:
found a random shadcn compatible autocomplete component which kinda works:

image

PR #89, opened 2025-09-24
Parse context rules .md front-matter

fnloza:
### TL;DR

Added YAML front matter parsing for context rules in markdown files, enabling custom slugs and matcher expressions.

### What changed?

- Added YAML front matter parsing for markdown context rules
- Created `parseFrontMatter` function to extract YAML metadata from markdown files
- Implemented `parseMatchField` to convert string expressions into matcher objects
- Added `contextRuleFromMarkdownFile` helper to build context rules from a markdown file with front matter
- Updated `contextRulesFromFiles` to use the new helper function `contextRuleFromMarkdownFile`
- Removed verbose logging in pg-estate-bootstrap.ts, and replaced with concise rules summary 

### How to test?

1. Create a markdown file with YAML front matter containing `slug` and `match` fields:
   ```md
   ---
   slug: custom-rule-name
   match: never()
   ---
   This is the rule content
   ```

`pnpm dev -c estates/iterate/iterate.config.ts`
Note you need to update your script keys in package.json in our website app to 
   ``` 
    "dev:website": "doppler run -- vite dev",
    "build:website": "doppler run -- vite build",
```
to get this working locally 

2. Verify the rule is loaded with the custom slug and matcher when using `contextRulesFromFiles` and that they're being applied correctly in the debug UI

### DIFF
--- a/apps/os/backend/agent/context.ts
+++ b/apps/os/backend/agent/context.ts
@@ -2,6 +2,7 @@ import { dirname, join, resolve } from "path";
 import { readFileSync, accessSync } from "fs";
 import { fileURLToPath } from "url";
 import { globSync } from "glob";
+import YAML from "yaml";
 import jsonataLib from "jsonata/sync";
 import type {
   ContextRule,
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: apps/os/backend/agent/context.ts [hunk 1]
```diff
@@ -106,6 +107,101 @@ export const defineRule = <Rule extends ContextRule>(rule: Rule) => rule;
 
 export const defineRules = <Rules extends ContextRule[]>(rules: Rules) => rules;
 
+/**
+ * Parse YAML front matter from a markdown string and return { attrs, body }.
+ * If no front matter is present, attrs is {} and body is the original string.
```
mmkal, line 112: let's mimic the format of the most popular library that does this: https://www.npmjs.com/package/gray-matter

so return type would be `{ content: string; data: Record<string, unknown> }`

unless this is based on something else that's already standard-ish?

apps/os/backend/agent/context.ts
```diff
@@ -106,6 +107,101 @@ export const defineRule = <Rule extends ContextRule>(rule: Rule) => rule;
...
+  //console.log("parseFrontMatter:", { hasFrontMatter: true, keys: Object.keys(attrs) });
+  return { attrs, body };
+}
+
+/**
+ * Convert a front-matter `match` (string or object) to a ContextRuleMatcher or array.
+ * Supports expressions like always(), hasMCPConnection("foo"), and nested calls
+ * using the exported `matchers` object.
+ */
+function parseMatchField(raw: unknown): ContextRule["match"] | undefined {
```
mmkal, line 135: can we not just demand for now that people write

```
match: {type: always}
match: {type: jsonata, expression: agentCoreState.whatever.foo}
```

apps/os/backend/agent/context.ts
```diff
@@ -106,6 +107,101 @@ export const defineRule = <Rule extends ContextRule>(rule: Rule) => rule;
...
+  overrides?: Partial<ContextRule>;
+}): ContextRule {
+  const { attrs, body } = parseFrontMatter(fileContent);
+  const slug =
+    typeof attrs.slug === "string" && attrs.slug.trim() ? (attrs.slug as string).trim() : undefined;
+  const match = parseMatchField((attrs as any)?.match);
+  const key = slug ?? filePath.replace(/\.md$/, "");
+  const matchSummary = Array.isArray(match)
+    ? { kind: "array", types: match.map((m: any) => m?.type) }
+    : { kind: match ? "single" : "none", type: (match as any)?.type };
```
mmkal, line 192: I think we can rely on the existing ContextRule schema here, if we expect them to write plain-old-yaml-objects instead of us implementing a mini parser of javascript expressions

```ts
const {attrs, body} = parseFrontMatter(...) // or const {content, data} maybe
const key = slug ?? filePath.replace(/\.md$/, "");
const rule = ContextRule.parse({
  key,
  ...attrs,
  prompt: body,
})
```

apps/os/backend/agent/context.ts
```diff
@@ -350,16 +446,13 @@ export function contextRulesFromFiles(pattern: string, overrides: Partial<Contex
     if (!configDir) {
       throw new Error("iterate.config.ts not found");
     }
+    //console.log("contextRulesFromFiles:", { pattern, configDir });
     const files = globSync(pattern, { cwd: configDir }) as string[];
+    //console.log("contextRulesFromFiles: files:", files);
```
mmkal, line 451: just a comment to not forget to remove these before merge

apps/os/backend/agent/context.ts
```diff
@@ -350,16 +446,13 @@ export function contextRulesFromFiles(pattern: string, overrides: Partial<Contex
...
     return files.map((filePath: string) => {
       const fileContent = readFileSync(join(configDir, filePath), "utf-8");
-      console.log("fileContent", fileContent);
-      // Get relative path from config directory and remove .md extension
-      return defineRule({
-        key: filePath.replace(/\.md$/, ""),
-        prompt: fileContent,
-        ...overrides,
-      });
+      // Use helper to parse front matter and match expression
```
mmkal, line 454:
```suggestion
```

apps/os/backend/agent/context.ts
```diff
@@ -106,6 +107,101 @@ export const defineRule = <Rule extends ContextRule>(rule: Rule) => rule;
...
+    : { kind: match ? "single" : "none", type: (match as any)?.type };
+  const matchRaw = (attrs as any)?.match;
+  console.log("contextRuleFromMarkdownFile:", {
+    filePath,
+    key,
+    slug,
+    hasMatch: Boolean(match),
+    matchRaw,
+    matchSummary,
+  });
```
mmkal, line 201:
```suggestion
```

apps/os/backend/agent/context.ts
fnloza: why? / what would we gain? 

from a user perspective it's cleaner / more intuitive?

mmkal: We would not have this complex and incomplete partial pseudo-JavaScript parser which `eval`s expressions from a markdown file

apps/os/backend/agent/context.ts
fnloza: plain old yaml objects would mean instead of  e.g hasMCPConnection() 

a user would have to do: 
```
____ 
slug: how-we-use-linear
match:
  type: jsonata
  expression: >-
    $count(
      agentCoreState.mcpConnections.*[
        $contains($string(serverUrl), "https://api.githubcopilot.com/mcp/") or
        $contains($string(serverName), "https://api.githubcopilot.com/mcp/")
      ]
    ) > 0
---
```
Or am I misunderstanding? that would be a lot worse imo..

mmkal: Yeah there would be some usability gaps for sure, but I'd rather plug them with jsonata which does allow custom functions and has a full spec etc.

PR #94, opened 2025-09-24
draft evals w/ evalite (opening for visibility)

mmkal:


.github/workflows/eval.yml
```diff
@@ -0,0 +1,84 @@
...
+          # For some reason the nodejs v24 installation in the Dockerfile fails in CI, but the built-in version of nodejs works for these purposes
+          echo '
+          FROM docker.io/cloudflare/sandbox:0.3.2
+          EXPOSE 3000
+          ' > apps/os/backend/sandbox/Dockerfile
+      - name: run server and test
+        # run the server and test in one big step so logs are visible
+        run: |
+          # for some reason `doppler run -- ...` doesn't inject env vars into the server process, so write to .env
+          doppler run -- printenv > apps/os/.env
```
mmkal, line 49: @NickBlow (and @ Rahul if he were in)

To get `pnpm preview` working I had to do quite a few weird things.

1. delete packages/ngrok/package.json - ideally this would just noop if ngrok isn't found. (or maybe we could [stop using ngrok](https://x.com/samgoodwin89/status/1971350552815730724)?)
2. simplify the sandbox Dockerfile - it was failing to install nodejs 24. After lots of trial and error I looked at the base dockerfile and it already had nodejs so I just deleted the install commands and everything still worked
3. `doppler run -- pnpm preview` didn't seem to actually inject env vars, so I had to use `doppler run -- printenv > .env` which vite automatically picks up, but I know in general we don't like .env files.

These changes are never committed, but still not great to have to mess with source files like this.

apps/os/app/components/impersonate.tsx
```diff
@@ -1,4 +1,3 @@
-"use client";
```
mmkal, line 1: this was causing warnings in `pnpm build`

jonastemplestein (approved):
This looks good, let's merge!

My spidey sense tells me we might tie ourselves in knots with evalite (as opposed to our own thin vitest wrapper). So hopefully we can limit the amount of deeply evalite specific stuff

Using evalite as eval results reporter feels like it might be the worst of both worlds between

1) Using braintrust (or similar) hosted SaaS - turnkey and nothing for us to improve
2) Using our own completely custom viewer that can e.g. also consume serialised "full agent traces" (that can be loaded by our own agents debug UI)

But I don't think we need to change anything because of this feeling. Just wanted to air it :)

mmkal:
> This looks good, let's merge!
> 
> My spidey sense tells me we might tie ourselves in knots with evalite (as opposed to our own thin vitest wrapper). So hopefully we can limit the amount of deeply evalite specific stuff
> 
> Using evalite as eval results reporter feels like it might be the worst of both worlds between
> 
> 1. Using braintrust (or similar) hosted SaaS - turnkey and nothing for us to improve
> 2. Using our own completely custom viewer that can e.g. also consume serialised "full agent traces" (that can be loaded by our own agents debug UI)
> 
> But I don't think we need to change anything because of this feeling. Just wanted to air it :)

Yeah have definitely spent more time fiddling with evalite than writing actual evals. *But*, I would definitely have spent more time and/or have far fewer nice useful eval-debugging features if we were writing our own UI. There's just a decent amount of CRUD boilerplate, database schemas, and UI that Matt's done for us. It also doesn't stop us using braintrust for more long term eval-ing.

Also it wasn't so much fighting against evalite as having to add stuff on to it. It also might be nice if we are the ones to really push evalite forward into being what we (and others) _actually_ need from an eval tool. Getting it to run as a plain vitest wrapper would be a first step I think.

PR #97, opened 2025-09-25
better replicate

jonastemplestein:
I should have just done this directly - turns out our replicate integration need not be more than 20 lines of code and is now much more flexible and easier for the LLM to use


apps/os/backend/agent/iterate-agent.ts
```diff
@@ -1163,52 +1163,63 @@ export class IterateAgent<Slices extends readonly AgentCoreSlice[] = CoreAgentSl
...
-    const result = await replicateIntegration.generateImage(input, {
-      replicateApiToken: this.env.REPLICATE_API_TOKEN,
-      openaiApiKey: this.env.OPENAI_API_KEY,
-      iterateUser: this.env.ITERATE_USER,
-      estateId: this.databaseRecord.estateId,
-      db: this.db,
+    const replicate = new Replicate({ auth: env.REPLICATE_API_TOKEN, useFileOutput: false });
+
+    // Because we set useFileOutput to false above, this will return an array of URLs
+    const replicateResponse = await replicate.run(input.model, {
```
zakhardavydov, line 1197: From https://github.com/replicate/replicate-javascript

> A model that outputs file data returns a FileOutput object by default. This is an implementation of ReadableStream that returns the file contents. It has a .blob() method for accessing a Blob representation and a .url() method that will return the underlying data-source.
> 
> We recommend accessing file data directly either as readable stream or via .blob() as the .url() property may not always return a HTTP URL in future.

Maybe we can read the blob and store it within our storage here?

zakhardavydov: We recommend accessing file data directly either as readable stream or via .blob() as the .url() property may not always return a HTTP URL in future.

I am just concern about this line, but fine otherwise

jonastemplestein: Let's change that when they stop providing the URLs

I prefer it when replicate's SDK provides data, rather than functions, because then the LLM can do something useful with it even when replicate changes their SDK or if other models return other stuff

Because if you JSON stringify a `.blob()` property it disappears

zakhardavydov (approved):

PR #98, opened 2025-09-25
Verify slack webhooks

jonastemplestein:
Best to do this before we're onboarding customers :)

I already updated the stg, prd, and prd_tmp_v2 doppler configs (no idea if the latter is being used)

The code is stolen from `bolt` (a weird but official slack application development framework https://github.com/slackapi/bolt-js/blob/main/src/receivers/verify-request.ts)

### DIFF
--- a/apps/os/backend/integrations/slack/slack.ts
+++ b/apps/os/backend/integrations/slack/slack.ts
@@ -1,3 +1,4 @@
+import { createHmac, timingSafeEqual } from "node:crypto";
 import { Hono } from "hono";
 import { and, eq } from "drizzle-orm";
 import { WebClient, type UsersListResponse } from "@slack/web-api";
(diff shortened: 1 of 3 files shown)


apps/os/backend/integrations/slack/slack.ts
```diff
@@ -472,3 +497,77 @@ export async function syncSlackUsersInBackground(db: DB, botToken: string) {
...
+const verifyErrorPrefix = "Slack request verification";
+
+/**
+ * Verifies the signature of an incoming request from Slack.
+ * If the request is invalid, this method throws an exception with the error details.
+ */
+export function verifySlackRequest(options: SlackRequestVerificationOptions): void {
+  const requestTimestampSec = options.headers["x-slack-request-timestamp"];
+  const signature = options.headers["x-slack-signature"];
+  if (Number.isNaN(requestTimestampSec)) {
```
NickBlow, line 525: nit: I kind of would prefer to return an object like `{valid: false, reason:""} | {valid: true}` instead of throwing

But probably subjective

apps/os/backend/integrations/slack/slack.ts
```diff
@@ -472,3 +497,77 @@ export async function syncSlackUsersInBackground(db: DB, botToken: string) {
...
+  ) {
+    throw new Error(`${verifyErrorPrefix}: signature mismatch`);
+  }
+}
+
+/**
+ * Verifies the signature of an incoming request from Slack.
+ * If the request is invalid, this method returns false.
+ */
+export function isValidSlackRequest(options: SlackRequestVerificationOptions): boolean {
```
mmkal: Let's avoid throwing for our own control flow - let's return a `{success: true} | {success: false; error: string}` or similar

IMO we should be throwing errors as little as possible, and even only catching them when we're relying on some third party library. Pragmatic exceptions like when we have to throw an error deep in the callstack are fine but this isn't one

jonastemplestein: Agree, although I did copy this verbatim from a third party 🤣  Will change

apps/os/backend/integrations/slack/slack.ts
```diff
@@ -38,8 +39,32 @@ async function slackTeamIdToEstateId({ db, teamId }: { db: DB; teamId: string })
...
-  const body = (await c.req.json()) as SlackWebhookPayload;
+
+  // Get raw request body for signature verification
+  const rawBody = await c.req.text();
+  const signature = c.req.header("x-slack-signature");
+  const requestTimestamp = c.req.header("x-slack-request-timestamp");
+  if (!signature || !requestTimestamp) {
+    return c.text("Slack webhook received without required signature headers", 400);
+  }
+  const signingSecret = c.env.SLACK_SIGNING_SECRET;
```
mmkal, line 50: can you add https://iterate-com.slack.com/archives/C06LU7PGK0S/p1758805208202929?thread_ts=1758805180.468249&cid=C06LU7PGK0S to the root readme pls

PR #104, opened 2025-09-25
post internal errors in slack

myst-6:
Resolves [ITE-2931: Post internal error into slack channel](https://linear.app/iterate-com/issue/ITE-2931/post-internal-error-into-slack-channel)


myst-6:
Screenshot 2025-09-25 at 15 14 32

Looks like this

apps/os/backend/agent/slack-agent.ts
```diff
@@ -129,6 +130,16 @@ export class SlackAgent extends IterateAgent<SlackAgentSlices> implements ToolsI
...
             );
             break;
+          case "CORE:INTERNAL_ERROR": {
+            console.error("[SlackAgent] Internal Error:", payload.event);
+            const { data } = payload.event;
+            this.getAgentDebugURL().then((url) =>
+              this.sendSlackMessage({
+                text: `There was an internal error; the debug URL is ${url.debugURL}.\n\n${data.error.slice(0, 256)}${data.error.length > 256 ? "..." : ""}`,
+              }),
+            );
```
mmkal: let's just use `this.ctx.waitUntil` as a rule. Can then make this async. One of these days I'm going to turn on the `no-floating-promises` rule and this will be a violation

myst-6: Oh yeah my bad I forgot to put it there 😭

jonastemplestein: btw you can just import `waitUntil` from `cloudflare:workers`

PR #105, opened 2025-09-25
fix little bugs

fnloza:
### TL;DR

fix small bugs in prompt-land 

### What changed?

- Improved link formatting instructions, requiring agents to use Slack's inline link format `<URL | descriptive text>` instead of raw URLs
- Added instruction for agents to connect to MCP servers before asking for clarification if a task relates to a specific MCP server (as they might get additional context rules on connection) 
- Clarified image generation guidance, specifically when using Slack avatar URLs
- Fixed documentation in the `sendSlackMessage` tool to clarify that `endTurn` should be set to `true` (not "end-turn")
- Enhanced guidance on message repetition, specifying that agents should not repeat error messages if the state hasn't changed

### DIFF
--- a/apps/os/backend/agent/default-context-rules.ts
+++ b/apps/os/backend/agent/default-context-rules.ts
@@ -91,12 +91,13 @@ const defaultSlackAgentPrompt = dedent`
   - If you are addressing the entire team / making an announcement, use "we" to refer to the team.
   - Never make up or guess facts or function tool parameters. Always be honest.  
   - Briefly acknowledge mistakes and correct yourself when you've made a mistake.
-  - Never repeat a message that has already been communicated to the user.
+  - Never repeat a message or update that has already been communicated to the user.
+    - e.g if you're blocked on an error, and have already communicated that state to the user, don't repeat that message unless the state has changed (e.g you are now unblocked). If you keep re-trying and keep hitting the same error, then do it silently. 
 
   Message formatting:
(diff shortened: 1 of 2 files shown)


apps/os/backend/agent/default-context-rules.ts
```diff
@@ -91,12 +91,13 @@ const defaultSlackAgentPrompt = dedent`
   - If you are addressing the entire team / making an announcement, use "we" to refer to the team.
   - Never make up or guess facts or function tool parameters. Always be honest.  
   - Briefly acknowledge mistakes and correct yourself when you've made a mistake.
-  - Never repeat a message that has already been communicated to the user.
+  - Never repeat a message or update that has already been communicated to the user.
+    - e.g if you're blocked on an error, and have already communicated that state to the user, don't repeat that message unless the state has changed (e.g you are now unblocked). If you keep re-trying and keep hitting the same error, then do it silently.
```
jonastemplestein, line 95: Why do you not want the agent to tell the user about errors?

jonastemplestein (approved):

apps/os/backend/agent/default-context-rules.ts
fnloza: I want to avoid this: 
https://iterate-com.slack.com/archives/C08R1SMTZGD/p1758807411689619

PR #106, opened 2025-09-25
prefix mcp connections

zakhardavydov:


apps/os/backend/agent/mcp/mcp-event-hooks.ts
```diff
@@ -54,8 +54,18 @@ interface MCPConnectionResult {
   events: MCPEventHookReturnEvent[];
 }
 
+type MCPManagerCacheKey = string;
```
mmkal: may as well use `Branded<...>` here

mmkal (approved):

PR #114, opened 2025-09-25
Feature: debounced typing indicator

zakhardavydov:
Fixes [ITE-2736: Typing indicator facelift](https://linear.app/iterate-com/issue/ITE-2736/typing-indicator-facelift)

### DIFF
--- a/apps/os/backend/agent/default-context-rules.ts
+++ b/apps/os/backend/agent/default-context-rules.ts
@@ -221,14 +221,21 @@ export const defaultContextRules = async () => [
     key: "using-linear",
     prompt: dedent`
       When using Linear tools:
-      - When displaying Linear issues, use: "<issue.url|issue.identifier>: title".
+      - When displaying Linear issues, use: "<issue.url|issue.identifier>: title (_state_)".
       - For bug tickets, gather: steps to reproduce, expected vs. actual behavior, and environment.
       - For feature tickets, require: user story and success criteria.
       - If information is missing, ask targeted questions; avoid using placeholders like "TBD".
       - Tools that take a "limit" parameter can be slow if limit is > 10 so default to that
-      - When using the list_my_issues tool:
-        - make sure to exclude any issues that have status Canceled, Duplicate or Done
-        - default to limit 10
+      - When using the linear_oauth_proxy_list_issues tool:
+        - Active issues are recently (in the last 7 days) created or updated issues with state 'Up Next', 'In Progress' or 'In Review' 
+        - Inactive issues are issues with state 'Backlog', 'Done', 'Duplicate' or 'Cancelled'
+        - Filtering by state:
+          - The tool accepts one state per call. To filter fetched issues on multiple states, make parallel calls (one per state). To fetch all states, omit the state filter.
+          - Examples:
+            - linear_oauth_proxy_list_issues({ state: 'Up Next' })
+            - linear_oauth_proxy_list_issues({ state: 'In Progress' })
+            - linear_oauth_proxy_list_issues()
+        - Default limit: 10 per call
     `,
     match: matchers.hasMCPConnection("mcp.linear.app"),
   },
--- a/apps/os/backend/auth/integrations.ts
+++ b/apps/os/backend/auth/integrations.ts
@@ -18,8 +18,7 @@ import { syncSlackUsersInBackground } from "../integrations/slack/slack.ts";
 import { MCPOAuthState, SlackBotOAuthState } from "./oauth-state-schemas.ts";
 
 export const SLACK_BOT_SCOPES = [
-  // commented out as it may have broken slack integration
-  // "app_mentions:read",
+  "app_mentions:read",
   "channels:history",
   "channels:join",
   "channels:read",


jonastemplestein (approved):
V clean design - this makes it quite easy to later add an argument `updateStatus : string` to all function tool calls that simply returns an additional event when passed in

apps/os/backend/agent/slack-agent.ts
```diff
@@ -60,6 +61,52 @@ export class SlackAgent extends IterateAgent<SlackAgentSlices> implements ToolsI
...
+
+    await this.slackAPI.assistant.threads
+      .setStatus({
+        channel_id: slackChannelId,
+        thread_ts: slackThreadId,
+        status: status || "",
+      })
+      .catch((error) => {
+        console.error("[SlackAgent] Failed to update typing status:", error);
+      });
```
mmkal:
```suggestion
```

I don't think errors are likely to happen, if they do happen it doesn't affect the rest of our flow, and catching like this just makes them harder to see

apps/os/backend/agent/slack-agent.ts
```diff
@@ -60,6 +61,52 @@ export class SlackAgent extends IterateAgent<SlackAgentSlices> implements ToolsI
...
 
+  private updateSlackStatusDebounced = pDebounce(async (status: string | null) => {
+    const { slackChannelId, slackThreadId } = this.getReducedState() as SlackSliceState;
+    if (!slackChannelId || !slackThreadId) return;
+
+    await this.slackAPI.assistant.threads
+      .setStatus({
+        channel_id: slackChannelId,
+        thread_ts: slackThreadId,
+        status: status || "",
```
mmkal: i'm guessing `""` is how we clear it?

apps/os/backend/agent/slack-agent.ts
```diff
@@ -60,6 +61,52 @@ export class SlackAgent extends IterateAgent<SlackAgentSlices> implements ToolsI
...
+  }, 300);
+
+  private checkAndClearTypingIndicator = pDebounce(async () => {
+    const state = this.agentCore.state as SlackSliceState;
+    const status = state.typingIndicatorStatus;
+
+    if (!status) return;
+
+    if (this.agentCore.llmRequestInProgress()) {
+      this.checkAndClearTypingIndicator();
```
mmkal: this is a floating promise. needs `this.ctx.waitUntil`

apps/os/backend/agent/slack-agent.ts
```diff
@@ -60,6 +61,52 @@ export class SlackAgent extends IterateAgent<SlackAgentSlices> implements ToolsI
...
+      return;
+    }
+
+    this.agentCore.addEvents([
+      {
+        type: "SLACK:UPDATE_TYPING_STATUS",
+        data: { status: null },
+      },
+    ]);
+    this.updateSlackStatusDebounced(null);
```
mmkal: also a floating promise

apps/os/backend/agent/slack-agent.ts
```diff
@@ -60,6 +61,52 @@ export class SlackAgent extends IterateAgent<SlackAgentSlices> implements ToolsI
...
+  }, 15000);
+
+  private syncTypingIndicator() {
+    const state = this.agentCore.state as SlackSliceState;
+    const status = state.typingIndicatorStatus ?? null;
+
+    this.updateSlackStatusDebounced(status);
```
mmkal: also a floating promise

apps/os/backend/agent/slack-agent.ts
```diff
@@ -60,6 +61,52 @@ export class SlackAgent extends IterateAgent<SlackAgentSlices> implements ToolsI
...
+
+    if (status) {
+      this.checkAndClearTypingIndicator();
```
mmkal: also a floating promise

apps/os/backend/agent/slack-agent.ts
```diff
@@ -60,6 +61,52 @@ export class SlackAgent extends IterateAgent<SlackAgentSlices> implements ToolsI
...
+      });
+  }, 300);
+
+  private checkAndClearTypingIndicator = pDebounce(async () => {
+    const state = this.agentCore.state as SlackSliceState;
+    const status = state.typingIndicatorStatus;
+
+    if (!status) return;
+
+    if (this.agentCore.llmRequestInProgress()) {
```
mmkal, line 81: hmm could this logic lead to an infinite typing indicator if we have some bug that makes it look like an LLM request is in progress forever?

apps/os/backend/agent/slack-agent.ts
```diff
@@ -611,9 +647,19 @@ export class SlackAgent extends IterateAgent<SlackAgentSlices> implements ToolsI
...
     });
 
-    // Batch add events
     await this.addEvents(events);
 
+    if (this.agentCore.state.triggerLLMRequest) {
+      this.agentCore.addEvents([
+        {
+          type: "SLACK:UPDATE_TYPING_STATUS",
+          data: { status: "is typing..." },
```
mmkal: how come we have this in two places?

apps/os/backend/agent/slack-slice.ts
```diff
@@ -55,29 +55,51 @@ export const SlackUpdateSliceStateInput = z.object({
...
 export type SlackUpdateSliceState = z.infer<typeof SlackUpdateSliceState>;
+export type SlackUpdateTypingStatus = z.infer<typeof SlackUpdateTypingStatus>;
 export type SlackSliceEvent = z.infer<typeof SlackSliceEvent>;
 export type SlackSliceEventInput = z.input<typeof SlackEventInput>;
 
 export interface SlackSliceState {
   slackThreadId?: string | null;
   slackChannelId?: string | null;
   botUserId?: string;
+  typingIndicatorStatus?: string | null;
```
mmkal, line 102: Let's force ourselves to be explicit

```suggestion
  typingIndicatorStatus: string | null;
```

apps/os/backend/agent/slack-agent.ts
zakhardavydov: Yeah, that is how slack api is requiring it: https://docs.slack.dev/reference/methods/assistant.threads.setStatus

apps/os/backend/agent/slack-agent.ts
zakhardavydov: Yes, so we need to be super careful with marking LLM request as ending

mmkal (approved):

PR #115, opened 2025-09-25
Update linear prompting to retrieve issues correctly

fnloza:


apps/os/backend/agent/default-context-rules.ts
```diff
@@ -214,21 +254,28 @@ export const defaultContextRules = async () => [
...
-      }),
+      },
     ],
   }),
   {
     key: "using-linear",
     prompt: dedent`
       When using Linear tools:
-      - When displaying Linear issues, use: "<issue.url|issue.identifier>: title".
+      - When displaying Linear issues, use: "<issue.url|issue.identifier>: title (_state_)".
```
mmkal, line 264: are `.title` and `.state` props on the issue that come from linear? if so might work better to be consistent in how we refer to each prop like
```suggestion
      - When displaying Linear issues, use: "<issue.url|issue.identifier>: issue.title (_issue.state_)".
```

PR #117, opened 2025-09-26
client fix

zakhardavydov:
- We had multiple clients registered per user
- This PR makes sure we only have one client per user per provider
- In case we hit an unrecoverable MCP error, we clean clients and tokens

### DIFF
--- a/apps/os/backend/agent/mcp/mcp-oauth-provider.ts
+++ b/apps/os/backend/agent/mcp/mcp-oauth-provider.ts
@@ -37,7 +37,6 @@ export class MCPOAuthProvider implements AgentsOAuthProvider {
   serverId: string | undefined;
   authUrl: string | undefined;
 
-  private dbAccountId: string | undefined;
   private baseUrl: string;
 
   constructor(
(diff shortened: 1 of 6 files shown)


apps/os/backend/agent/mcp/mcp-oauth-provider.ts
```diff
@@ -60,26 +60,70 @@ export class MCPOAuthProvider implements AgentsOAuthProvider {
...
+          eq(schema.account.userId, this.params.userId),
           eq(schema.account.providerId, this.providerId),
         ),
-      );
+        orderBy: (account, { desc }) => [desc(account.createdAt)],
+      });
+
+      if (recentAccount) {
+        await this.params.db.delete(schema.account).where(eq(schema.account.id, recentAccount.id));
+      }
```
myst-6: Why can't we just always do this? Or even just delete all of the accounts instead of the first one?

apps/os/backend/agent/mcp/mcp-oauth-provider.ts
```diff
@@ -60,26 +60,70 @@ export class MCPOAuthProvider implements AgentsOAuthProvider {
...
+      } else {
+        return undefined;
+      }
+    }
+
     const account = await this.params.db.query.account.findFirst({
       where: and(
         eq(schema.account.userId, this.params.userId),
         eq(schema.account.providerId, this.providerId),
+        eq(schema.account.accountId, this.clientId),
```
myst-6: is there any reason why we have a one to many relationship between (userId,providerId) and accountId? Why would we want there to be many accounts for a user on a provider?

myst-6: If we dont have to store accountId it would be way better. And it seems like these are actually the same thing anyway? (Or accountId is actually an outdated clientId in some cases..?)

PR #119, opened 2025-09-26
Bring back intelligent link unfurling

myst-6:
Example: https://iterate-com.slack.com/archives/C08R1SMTZGD/p1758885463649179

### DIFF
--- a/apps/os/backend/agent/slack-agent-tools.ts
+++ b/apps/os/backend/agent/slack-agent-tools.ts
@@ -37,9 +37,8 @@ export const slackAgentTools = defineDOTools({
         .enum(["never", "auto", "all"])
         .default("auto")
         .optional()
-        .describe(
-          "Whether to unfurl links and media. If auto, unfurls links and media when and only when there is exactly 1 link in the message.",
-        ),
+        // If auto, unfurls links and media when and only when there is exactly 1 link in the message.
+        .describe("Whether to unfurl links and media."),
       endTurn: z
         .boolean()
         .default(false)


apps/os/backend/agent/slack-agent.ts
```diff
@@ -659,10 +659,18 @@ export class SlackAgent extends IterateAgent<SlackAgentSlices> implements ToolsI
 
     const { endTurn, ...sendInput } = input;
 
+    if (typeof sendInput.unfurl === "undefined") {
```
jonastemplestein: Not v important in this context, but I would prefer `unfurl: "auto" | "never" | "all"` or something, with a well defined default of "auto"

Treating undefined, true, false as options is less clear and can easily cause bugs

And if you must use undefined, why not `sendInput === undefined`?

jonastemplestein (approved):
Weird, I had written a bunch of code on this once that I can't find now. I had also documented the v strange behaviour of the `unfurl_links` and `unfurl_media` options on the slack API call in a bit more detail

The behaviour I'd landed on after playing with it for a bit was that unfurling is turned off if number of links > 2. Somehow two unfurled links or linear tickets still looked fine to me 🤣 

And just for the avoidance of doubt - you _don't_ want the LLM to be able to set this, right? Because it is currently not allowed in the `overrideJSONSchema` of the `sendSlackMessage` tool in `default-context-rules.ts`

apps/os/backend/agent/slack-agent.ts
myst-6: Yes - I don't want the LLM to see it.
It makes sense to have that enum to me!

apps/os/backend/agent/slack-agent-tools.ts
```diff
@@ -33,14 +33,12 @@ export const slackAgentTools = defineDOTools({
...
+      unfurl: z
         .boolean()
         .optional()
-        .describe("Whether to unfurl links. If undefined, uses intelligent heuristics."),
-      unfurlMedia: z
-        .boolean()
-        .optional()
-        .describe("Whether to unfurl media. If undefined, uses intelligent heuristics."),
+        .describe(
+          "Whether to unfurl links and media. If undefined, unfurls links and media iff there is exactly 1 link in the message.",
```
mmkal: "If undefined, unfurls links and media iff there is exactly 1 link in the message" is way less clear than "If undefined, uses intelligent heuristics." to me

If I was an LLM I would be 50/50 on whether "iff" meant "if and only if" or was a typo by the developer. And the "if undefined iff" is even weirder.  AND, it's unnecessary information! Our code is who's apply this rule

apps/os/backend/agent/slack-agent.ts
```diff
@@ -659,10 +659,16 @@ export class SlackAgent extends IterateAgent<SlackAgentSlices> implements ToolsI
 
     const { endTurn, ...sendInput } = input;
 
+    const numLinks = sendInput.text.match(/https?:\/\/[^\s|]+/g)?.length;
+    const doUnfurl = sendInput.unfurl === "all" || (sendInput.unfurl === "auto" && numLinks === 1);
```
mmkal, line 663: this is more of a product decision but I would prefer no-unfurlying unless there's one link that's known to be short-form or unfurl friendly based on a whitelist - e.g. `['x.com', 'twitter.com', 'medium.com', 'github.com'].includes(hostname)` or something

kind of just a personal preference, I hate unfurling most of the time

apps/os/backend/agent/slack-agent-tools.ts
```diff
@@ -33,14 +33,13 @@ export const slackAgentTools = defineDOTools({
...
-        .describe("Whether to unfurl links. If undefined, uses intelligent heuristics."),
-      unfurlMedia: z
-        .boolean()
+      unfurl: z
+        .enum(["never", "auto", "all"])
+        .default("auto")
         .optional()
-        .describe("Whether to unfurl media. If undefined, uses intelligent heuristics."),
+        .describe(
+          "Whether to unfurl links and media. If auto, unfurls links and media iff there is exactly 1 link in the message.",
```
mmkal: reposting:

"If undefined, unfurls links and media iff there is exactly 1 link in the message" is way less clear than "If undefined, uses intelligent heuristics." to me

If I was an LLM I would be 50/50 on whether "iff" meant "if and only if" or was a typo by the developer. And the "if undefined iff" is even weirder. AND, it's unnecessary information! Our code is who's apply this rule

apps/os/backend/agent/slack-agent-tools.ts
myst-6: duplicate

apps/os/backend/agent/slack-agent-tools.ts
myst-6: this is more for internal documentation, not shown to the LLM (but like technically I guess it could be in the future if we decided to let it specify?)
'intelligent heuristics' means nothing and is definitely less clear IMO but i've changed iff to 'when and only when' which should be clearer

apps/os/backend/agent/slack-agent.ts
myst-6: slack has its own whitelist of domains which are unfurled already, so I think it's fine

apps/os/backend/agent/slack-agent-tools.ts
```diff
@@ -33,14 +33,13 @@ export const slackAgentTools = defineDOTools({
...
-        .describe("Whether to unfurl links. If undefined, uses intelligent heuristics."),
-      unfurlMedia: z
-        .boolean()
+      unfurl: z
+        .enum(["never", "auto", "all"])
+        .default("auto")
         .optional()
-        .describe("Whether to unfurl media. If undefined, uses intelligent heuristics."),
+        .describe(
+          "Whether to unfurl links and media. If auto, unfurls links and media when and only when there is exactly 1 link in the message.",
```
mmkal: Again I don't think we should tell the LLM the algorithm that we're going to apply

myst-6: Ok, I moved it out of the prompt into a comment :)

apps/os/backend/agent/slack-agent.ts
mmkal: Sure. It's whitelist is bad though, because it includes linear

PR #124, opened 2025-09-26
Fix: Call syncSlackUsersInBackground when linking Slack on the integrations page

myst-6:
Fixes [ITE-3003: Debug "Estate not found" message in MCP callback](https://linear.app/iterate-com/issue/ITE-3003/debug-estate-not-found-message-in-mcp-callback)

### DIFF
--- a/apps/os/backend/auth/integrations.ts
+++ b/apps/os/backend/auth/integrations.ts
@@ -458,6 +458,7 @@ export const integrationsPlugin = () =>
               return ctx.json({ error: "Can't find the existing user to link to" });
             }
             user = linkedUser.user;
+            waitUntil(syncSlackUsersInBackground(db, tokens.access_token));
           }
 
           if (!user) {


BlankParticle (approved):

apps/os/backend/auth/integrations.ts
```diff
@@ -458,6 +458,7 @@ export const integrationsPlugin = () =>
               return ctx.json({ error: "Can't find the existing user to link to" });
             }
             user = linkedUser.user;
+            waitUntil(syncSlackUsersInBackground(db, tokens.access_token));
```
mmkal, line 461: if not doing this causes the 404, should we also make the 404 more of a hint to try again in a few seconds

myst-6: By the time you get to the 404 it's too late - the user mapping hasn't been injected into the agent for its first message, so the only way to retry would be by starting a new thread. At that point I think it's better to keep the error message as-is, because I think by the time you open slack after linking it, and ping iterate for the first time and add it to the channel, this should have definitely completed, and otherwise there was some kind of error which won't be fixed by just waiting for a bit

myst-6: Anyway, we can follow up in the future to improve the error message if we need to but i'll just quickly merge this in case the customers onboarded at 4:30 try to use google 😂

PR #125, opened 2025-09-26
Feature: create SlackAgent for thread that mentions bot before the bot joined the channel

zakhardavydov:
Fixes [ITE-3001: Join threads after bot is added to the channel](https://linear.app/iterate-com/issue/ITE-3001/join-threads-after-bot-is-added-to-the-channel)

### DIFF
--- a/apps/os/backend/integrations/slack/slack.ts
+++ b/apps/os/backend/integrations/slack/slack.ts
@@ -352,12 +352,7 @@ async function handleBotChannelJoin(params: {
       continue;
     }
 
-    const hasBotMention = threadHistory.messages.some(
-      (m) =>
-        m.type === "message" &&
-        m.type &&
-        isBotMentionedInMessage(m as { user?: string; text?: string; type: string }, botUserId),
-    );
+    const hasBotMention = threadHistory.messages.some((m) => isBotMentionedInMessage(m, botUserId));
 
     if (!hasBotMention) continue;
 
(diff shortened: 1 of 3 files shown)


apps/os/backend/integrations/slack/slack.ts
```diff
@@ -487,6 +310,109 @@ export async function syncSlackUsersInBackground(db: DB, botToken: string) {
...
+
+  if (!history.ok || !history.messages) {
+    console.error("Failed to fetch channel history");
+    return;
+  }
+
+  const validMessages = history.messages.filter((m) => m.ts);
+  const threadsByTs = R.groupBy(validMessages, (m) => m.thread_ts || m.ts!);
+
+  for (const [threadTs] of Object.entries(threadsByTs)) {
```
mmkal: are we doing this one at a time to avoid rate limiting? if so is this definitely enough?

mmkal (approved):
i'm ok with this if it's not risking getting rate limited - i don't think we have any handling for that right now

apps/os/backend/integrations/slack/slack.ts
zakhardavydov: No, rate limit should be fine, not that we are approved app

zakhardavydov: Will change that to fetch those in parallel

PR #128, opened 2025-09-26
fix: Delegate to iterate sdk cli

NickBlow:


apps/os/sdk/cli/index.ts
```diff
@@ -1,14 +1,30 @@
...
+import { resolve, dirname } from "node:path";
+import { fileURLToPath } from "node:url";
 import { createCli } from "trpc-cli";
 import * as prompts from "@clack/prompts";
 import { t } from "./config.ts";
 import { estate } from "./commands/checkout-estate.ts";
 import { gh } from "./commands/gh-commands.ts";
 
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
```
mmkal: i recently discovered [`import.meta.dirname`](https://nodejs.org/api/esm.html#importmetadirname) is a thing now (and import.meta.filename)

apps/os/sdk/cli/index.ts
```diff
@@ -1,14 +1,30 @@
...
+  const result = spawnSync("node", [cliPath], {
+    stdio: "inherit",
+    cwd: process.cwd(),
+  });
+  process.exit(result.status ?? 0);
+}
 
-cli.run({ prompts });
+const cli = createCli({ router });
+if (process.argv) cli.run({ prompts });
```
mmkal: this will always be truthy no?
```suggestion
cli.run({ prompts });
```

apps/os/sdk/cli/index.ts
```diff
@@ -1,14 +1,30 @@
...
 const router = t.router({
   estate,
   gh,
 });
 
-const cli = createCli({ router });
+if (process.argv.length === 2) {
+  console.error("No command provided, assuming you want to run `iterate` sdk cli");
+  // Run the cli.js script from packages/sdk
+  const cliPath = resolve(__dirname, "../../../../packages/sdk/cli.js");
```
mmkal: based on https://github.com/iterate-com/iterate/pull/129 i think this should be flipped

so:

- this file is **the** CLI
- if you run `pnpm iterate whatever` from this repo root it's just going to run `doppler run -- tsx apps/os/sdk/cli/index.ts`
- the `bin` script in the sdk package runs a bundled version of this (so for feature parity we move the print-iterate.config.ts functionality over here)

NickBlow: right now we only run pnpm iterate with no args so if the current cli can handle that, that's perfect.

The iterate cli currently contains internal tools so i dont know how much use it is to bundle it to clients... but if we're gonna add stuff we can use externally im ok with that

Should we merge this in the interim or do that first?

mmkal: If we don't want to provide it to customers then it doesn't belong in the `sdk/` folder? I thought there was some useful stuff like `iterate gh do-stuff-with-github`.

But if this is breaking builds (?) I'm fine to get this in but we should change soon. I want to get in my other PR, we can go from there.

If there's this-repo-internal CLI stuff we want to do, I think those commands belong in a separate trpc router which can easily be composed/merged with this one anyway. But we can do that after merging if you want to get this in to unbreak stuff

PR #129, opened 2025-09-26
smaller sdk publish.

mmkal:
- rm all typescript from packages/sdk
- replace with tsdown which writes to packages/sdk/dist **from** `apps/os`
- tsdown bundles everything except for dependencies used in our exported types - these _can't_ be bundled, so must be put in the sdk package.json
- run tsdown from `prepare` (which means it runs after `pnpm install` - guarantees this repo works in all CI jobs etc.)
- rm `publish.ts` - all we need is a single, normal publish command (`pkg-pr-new` for now)
- remove `@iterate-com/os` as a dependency of the sdk package - the relevant parts of the os package are bundled in
- don't publish `@iterate-com/os` (we still could choose to one day but I don't really see the advantage)
-  run [arethetypeswrong](https://arethetypeswrong.github.io) afore publishing to make sure this continues to work correctly

it works! when customers install will be 99% zod and only around 50kB us:

image

---

image

this is weird. everything interesting goes on in this package. we build the sdk package from here. tsdown from the sdk package struggles to get the types right, because the .ts files live here. it can't handle `export * from '@iterate-com/os'` for now.

we don't need to publish the os package for now.

we want almost everything to be internal, so it gets bundled with the sdk, so that we don't need to worry about keeping the dependencies of the sdk package in sync with the sdk/ folder of this package. the exceptions need to be added as dependencies in the sdk package.json the reason we want these exceptions is that in some cases we depend on third-party types (e.g. zod) unfortunately tsdown can't bundle types right now. other dependencies that are not part of the types of the API surface can be bundled. this is different from tsdown's default behavior, which is to make all prod dependencies external


apps/os/tsconfig.json
```diff
@@ -3,9 +3,9 @@
     "module": "esnext",
     "target": "esnext",
     "types": ["@cloudflare/workers-types", "node", "vite/client"],
-    "sourceMap": true,
+    "sourceMap": false,
```
jonastemplestein, line 6: What does this change?

mmkal: turned source maps off because they point to code in `apps/os` which doesn't make sense in the `packages/sdk context. If we can find a way to avoid this build-from-os setup, we can turn them back on.

mmkal:
>This goes way beyond my level of understanding of non packaging and typescript bundling, but it seems great!
>
>My main question is what, if anything, do we need to consider when adding dependencies to the sdk or os? Why are some dependencies explicitly mentioned in the sdk package.json for example?

The explicitly-mentioned dependencies are the ones whose types are needed for our sdk to work properly. `.js` bundling is a well-solved problem, `.d.ts` not so much. So without lots of tsdown trickery, our `.d.ts` files will contain stuff like `import {z} from 'zod'`. So for the types to work, zod needs to be external, and not bundled. There are some tools that help with this like https://api-extractor.com but I think they're overkill for now.

PR #131, opened 2025-09-29
feat: Add user to estate cli command

NickBlow:


apps/os/sdk/cli/commands/add-to-estate.ts
```diff
@@ -0,0 +1,156 @@
...
+      throw new Error(`Estate with ID ${estateId} not found`);
+    }
+
+    // Check if the user is already a member of the organization
+    const existingMembership = await db.query.organizationUserMembership.findFirst({
+      where: and(
+        eq(schema.organizationUserMembership.userId, user.id),
+        eq(schema.organizationUserMembership.organizationId, estate.organizationId),
+      ),
+    });
```
mmkal: I really think we should start doing upserts instead of constantly doing `const existing = find(); if (existing) update() else insert()`

NickBlow: yep, agreed. added a unique constraint (doesn't make sense to have multiple roles on same org)

PR #135, opened 2025-09-29
a logger of our own

mmkal:
i borrowed from myself https://github.com/mmkal/artifact.ci/blob/cc20e64c9d9baeefcc5452ac1726c28be8a34fc6/src/tag-logger.ts

this adds a "tag logger" which uses async local storage, to keep track of "tags" that can be added to the logger. see logger.test.ts for how it's used/useful:

```ts
import { expect, test } from "vitest";
import { TagLogger } from "./tag-logger.ts";

const createLogger = () => {
  const calls: any[][] = [];
  const mocks = {
    info: (...args: any[]) => calls.push(["info", ...args]),
    warn: (...args: any[]) => calls.push(["warn", ...args]),
    error: (...args: any[]) => calls.push(["error", ...args]),
    debug: (...args: any[]) => calls.push(["debug", ...args]),
  };
  return {
    mocks,
    calls,
    logger: new TagLogger(mocks),
  };
};

test("logger", () => {
  const { logger, calls } = createLogger();
  const one = () => {
    logger.info("one");
  };
  logger.run("numero=uno", one);

  expect(calls).toEqual([["info", "[numero=uno]", "one"]]);
});

test("stores memories", () => {
  const { logger, calls } = createLogger();
  const one = (input: number) => {
    logger.debug("dbg-one", { input });
    logger.run("depth=prettydeep", () => {
      logger.debug("dbg-two", { depth: 1 });
      logger.run("depth=deeper", () => {
        if (input > 0.5) {
          logger.warn("pretty big input");
        }
      });
    });
  };
  logger.run("numero=uno", () => one(0.1));

  expect(calls).toEqual([]); // no logs, because there was no warning, so no need to recall logger.debug(...) calls

  logger.run("numero=dos", () => one(0.9));

  expect(calls).toEqual([
    [
      "warn",
      "[numero=dos][depth=prettydeep][depth=deeper]",
      "pretty big input",
      "memories:",
      [expect.stringMatching(/^2.*/), "debug", "[numero=dos]", "dbg-one", { input: 0.9 }],
      [
        expect.stringMatching(/^2.*/),
        "debug",
        "[numero=dos][depth=prettydeep]",
        "dbg-two",
        { depth: 1 },
      ],
    ],
  ]);
});

test("enterWith", async () => {
  const { logger, calls } = createLogger();
  const foo = () => {
    const one = () => {
      logger.setTag("hello");
      logger.info("one");
      logger.setTag("world");
      logger.info("one point five");
    };
    logger.setTag("foo");
    logger.run("numero=uno", one);
    logger.info("outside one()");
  };
  foo();

  expect(calls).toEqual([
    ["info", "[foo][numero=uno][hello]", "one"],
    ["info", "[foo][numero=uno][hello][world]", "one point five"],
    ["info", "[foo]", "outside one()"],
  ]);
});
```

For now though, we're just using to make sure we only ever log a single arg to cloudflare


jonastemplestein (approved):
I think this is the right direction - let's merge it

But before you merge, please can you make sure AI knows about this rule? Otherwise AI agents will require an extra roundtrip (after eslint feedback) for any code change

You can update llms.ts with a rule

We need to invest way more in our agent rules and documentation than we currently are

jonastemplestein:
Also happy to merge first and add rule after. That's probably better actually as it's less work but has the same result

PR #145, opened 2025-09-30
fix: Defensive checks for orphaned account data

NickBlow:


apps/os/backend/trpc/routers/integrations.ts
```diff
@@ -86,6 +86,7 @@ export const integrationsRouter = router({
 
     // Add estate-wide accounts
     estateAccounts.forEach(({ account: acc_item }) => {
+      // Skip if account is null (orphaned permission record)
```
jonastemplestein, line 89: I would log a warning here so we can grep the logs for it

jonastemplestein: As this is something that _shouldn't_ happen unless there is some bug, right?

jonastemplestein: And why is this called `acc_item` and not `account`?

jonastemplestein (approved):
🚀

jonastemplestein:
Merging this now - we can address the comment separately. Need to unbrick my account 🤣

PR #148, opened 2025-09-30
add testing router to iterate cli

mmkal:
exposes the sometimes-handy testing procedures to (our) CLI

this involved shimming the `cloudflare:workers` and `cloudflare:email` modules since we use verbatimModuleSyntax, and we randomly `import { stuff } from "cloudflare:workers"` in a lot of places. So until adding this shim, it wasn't possible to use most of our routers, but now it is.

we still have the problem of the cli living under the SDK folder which implies it'll be useful to customers, but it's not, so we need to split it in two sometime soon. it's extra confusing because we don't even bundle sdk/cli/index.ts into the sdk package.

https://github.com/user-attachments/assets/0c0bb7f7-fe39-4ec8-8eb2-8645a96561f0

### DIFF
--- a/apps/os/backend/db/client.ts
+++ b/apps/os/backend/db/client.ts
@@ -1,19 +1,10 @@
 import { drizzle } from "drizzle-orm/postgres-js";
 import postgres from "postgres";
+import { env } from "../../env.ts";
 import * as schema from "./schema.ts";
 
-const envOrError = await import("../../env.ts").catch(String);
-const getEnv = () => {
-  if (typeof envOrError === "string") {
-    throw new Error(
-      `${envOrError} - this occurs when trying to use "env" outside of a cloudflare environment`,
-    );
-  }
-  return envOrError.env;
-};
-
-const pg = (connectionString: string) =>
-  postgres(connectionString, {
+const pg = () =>
+  postgres(env.ITERATE_POSTGRES.connectionString, {
     // Use connection pooling with a small max for Cloudflare Workers
     max: 5,
     // If you are not using array types in your Postgres schema, disable `fetch_types` to avoid an additional round-trip (unnecessary latency)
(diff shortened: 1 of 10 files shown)


jonastemplestein:
Why is the CLI not useful to customers? 

We've had lots of ideas flying around like

- it can be used to create and link your estate repo (and interact with other iterate control plane features)
- it can be used by the agent in a container as a script runner (e.g. `pnpm iterate gh` could wrap `gh` or whatever)
- it can be used to get everything that is necessary to "boot up" a local dev environment 
- it can be used to get set up as a contributor to the iterate OSS project (or at least get walked through the setup of necessary oauth clients etc)
- customers might add their own iterate commands in an estate repo like `cli/` that get automatically picked up and can be run by an agent in a container with repo context (this is one of the ways of "adding tools" to the agent that i'm excited about)

I'm not sure we'll do all of these, but I definitely think of the iterate CLI as a very important part of the external interface to our company. Although that doesn't mean it needs to live in the SDK package! Maybe the commands that are available via the CLI just depend on the context its executed in?

jonastemplestein:
And on the PR itself, why is this in a `testing` router? Is this not something we'd want in production? I love interacting with a system I made via CLI - it's way better than a browser based UI in many situations

And how do you think about the difference between this approach (loading the router into the node process) vs what you previously made, where our server publishes its trpc routes and we have a CLI command to interact with all of them?

jonastemplestein:
Makes sense, thanks for explaining! 

I guess my mental model was something like this:

CLI = the entry point for _all_ CLI commands, whether run by us devs or by customers or anyone else

And then depending on the context the CLI is invoked in (i.e. a customer personal computer, a customer cloudflare container, an OS developer personal computer, an iterate repo CI box, etc) it might just load different commands

I wonder if we could even use iterate.config.ts to specify which commands to load. Then we only need one mechanism that can cover all the above scenarios

Perhaps something like 

```ts
defineConfig({
  ...otherArgs,
  cliCommandGlobs: ["cli/**/*.ts"]
})
```

Although one problem with that is that _at the moment_, there is no iterate config at the root of iterate-com/iterate

In any case, this is a great improvement, let's merge

PR #150, opened 2025-09-30
[WIP] Getting evals in braintrust

myst-6:
DO NOT MERGE!
Thoughts:
- This is the minimal lightweight version of what we want. The only problem is that it puts each test case in its own experiment and it's a bit janky
- I would prefer making a custom `IterateEval` function which will run `evalite` with the braintrust logic.
  - It can create a braintrust experiment with the same name as the first parameter,
  - It will then create one eval span in that experiment for each of the test cases, with the same inputs, and we can then pass the exported ids as input into the `task` function,
  - The scorers are automatically wrapped with braintrust.

@mmkal what do you think about my above proposal? I think it would be fairly clean.

### DIFF
--- a/apps/os/evals/helpers.ts
+++ b/apps/os/evals/helpers.ts
@@ -3,13 +3,13 @@ import { expect, vi } from "vitest";
 import { createTRPCClient, httpLink } from "@trpc/client";
 import { createAuthClient } from "better-auth/client";
 import { adminClient } from "better-auth/client/plugins";
+import { Eval, initLogger, type Span } from "braintrust";
 import type { AppRouter } from "../backend/trpc/root.ts";
 import type { AgentCoreEvent } from "../backend/agent/agent-core-schemas.ts";
 import type { MCPEvent } from "../backend/agent/mcp/mcp-slice.ts";
 import { type SlackSliceEvent } from "../backend/agent/slack-slice.ts";
 import type { SlackWebhookPayload } from "../backend/agent/slack.types.ts";
 import { testAdminUser } from "../backend/auth/test-admin.ts";
-import { Eval, initLogger, type Span } from "braintrust";
 
 export * from "./scorer.ts";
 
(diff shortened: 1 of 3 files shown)


apps/os/evals/helpers.ts
```diff
@@ -20,6 +21,31 @@ export const authClient = createAuthClient({
...
+    Eval(projectName, {
+      experimentName,
+      data: [{ input: null }],
+      scores: [],
+      metadata,
+      task: (_input, hooks) => {
+        resolve(hooks.span);
+        return null;
+      },
+    }).then(() => reject(new Error("Failed to create eval span")));
```
mmkal: kinda weird to call `reject` here, the idea being that if we haven't already called `resolve(...)` by now the promise should be rejected? Is this relying on `reject(...)` after `resolve(...)` has been called being a no-op?

apps/os/evals/helpers.ts
```diff
@@ -20,6 +21,31 @@ export const authClient = createAuthClient({
...
+  projectName: string;
+  experimentName: string;
+  metadata: Record<string, unknown>;
+}): Promise<Span> {
+  // this is a hack to get a span inside the 'experiments' tab
+  // instead of the 'logs' tab - it allows us to create better graphs
+  return new Promise((resolve, reject) => {
+    Eval(projectName, {
+      experimentName,
+      data: [{ input: null }],
```
mmkal: I like that we are now doing a much thinner version of using the full-blown braintrust `Eval`, and it's a nice simple function to call, but curious if we now lose any nice UI because our `input` is null, etc.

Would still be interested to dive into the `braintrust` package to see if we can create a span directly

mmkal (commented):
I'd be interested to see what an `IterateEval` looks like - I'm a bit worried about too much abstraction but probably it's worth it. Esp since we are trying to combine two tools which have a decent amount of overlap.

also wrapping `evalite` might cause confusion too, the `evalite` function is basically `test(...)` within a `describe(...)`:

Image

But if you have an idea of how it'd work I think worth opening another PR so we can look at them side-by-side?

apps/os/evals/helpers.ts
myst-6: I think diving deep into the package might yield us something, but it would take soo much time 😭

apps/os/evals/helpers.ts
myst-6: yeh thats true,
better to have a local variable and then reject if not null else resolve, will fix if we decide to use this branch

PR #151, opened 2025-09-30
Getting evals in braintrust

myst-6:
[ITE-3026: Add braintrust into evals](https://linear.app/iterate-com/issue/ITE-3026/add-braintrust-into-evals)

### DIFF
--- a/apps/os/evals/scorer.ts
+++ b/apps/os/evals/scorer.ts
@@ -45,6 +45,22 @@ function _multiTurnScorer(params: MultiTurnScorerParams = {}) {
   const scoreManually = async (newMessages: string[], score: { score: number; reason: string }) => {
     conversation.push(...newMessages);
     scores.push({ ...score, messages: newMessages });
+    const intermediateScoreSpan = startSpan({
+      name: "intermediate-score",
+      parent: params.braintrustSpanExportedId,
+      type: "score",
+    });
+    intermediateScoreSpan.log({
+      input: {
+        conversation,
+      },
+      output: {
+        score: score.score,
+        reason: score.reason,
+      },
+    });
+    intermediateScoreSpan.end();
+    await intermediateScoreSpan.flush();
   };
 
   const scoreTurn = async (newMessages: string[], expectation: string) => {


apps/os/evals/basic-conversation.eval.ts
```diff
@@ -1,19 +1,19 @@
...
 });
 
-evalite("agent knows when to end their turn", {
+IterateEval("agent knows when to end their turn", {
   data: async () => {
     return [
       {
         input: {
-          slug: "multi-turn conversation",
+          slug: "multi-turn fruits conversation",
```
mmkal: nit: i meant for these to be slugs so
```suggestion
          slug: "multi-turn-fruits-conversation",
```

apps/os/evals/helpers.ts
```diff
@@ -213,3 +230,77 @@ export type WaitForEvent = {
     },
   ): Promise<Selection>;
 };
+
+export function IterateEval<TInput, TOutput, TExpected>(
```
mmkal: naming it this would be a missed opportunity to call it `evaliterate`

myst-6: Omg it took me a bit to get it but that's genius

myst-6:
We still need to wrap scorers for them to be put on the span because it doesn't happen inside the task, so I think we would still want this wrapper

mmkal:
Yeah I think this would just mean the implementation of `evaliterate` becomes a little cleaner, we'd probably still want it. In general with these things I prefer doing it non-abstractly at first but I can see the code getting too ugly that way

apps/os/evals/helpers.ts
```diff
@@ -213,3 +230,64 @@ export type WaitForEvent = {
...
+
+  return evalite<TInput, TOutput, TExpected>(experimentName, {
+    data: opts.data,
+    columns: opts.columns,
+    task: async (input) => {
+      const braintrustSpan = experiment.startSpan({ name: "eval", type: "eval" });
+      spanMap[hash(input)] = braintrustSpan;
+      braintrustSpan.log({ input });
+      await braintrustSpan.flush();
+      const braintrustSpanExportedId = await braintrustSpan.export();
```
myst-6: Will this mess up the timing of the tasks reported by evalite? Should we pass in a getter instead?

myst-6: we need the span to be serialised to pass it into the agent via trpc before the first LLM request begins,
and it should be a small fraction of the test time (especially for multi-turn evals),
so it's (1) unavoidable - if we did some annoying stuff we could maybe save like a few ms but it's not worth and (2) not a deal-breaker

mmkal: We could also skip Braintust stuff completely conditionally on an env var so it doesn't affect local

myst-6: 👍 
done

.github/workflows/eval.yml
```diff
@@ -72,6 +72,7 @@ jobs:
           node wait.mjs
 
           # finally, run the tests
+          export PROJECT_NAME=gh-${{ github.head_ref || github.ref_name }}
```
mmkal:
```suggestion
          export PROJECT_NAME=gh-evals-${{ github.head_ref || github.ref_name }}
```

apps/os/backend/agent/iterate-agent.ts
```diff
@@ -410,14 +410,13 @@ export class IterateAgent<Slices extends readonly AgentCoreSlice[] = CoreAgentSl
     });
     const posthogClient = pMemoize(async () => {
       const estate = await getEstate();
-      const environment = getEnvironmentName({
+      const projectName = getProjectName({
         ITERATE_USER: this.env.ITERATE_USER,
-        STAGE__PR_ID: this.env.STAGE__PR_ID,
-        ESTATE_NAME: estate.name,
+        PROJECT_NAME: this.env.PROJECT_NAME,
       });
```
mmkal:
```suggestion
      const projectName = getProjectName(this.env);
```

apps/os/backend/utils/utils.ts
```diff
@@ -25,17 +25,17 @@ export function ensureString(value: unknown): string {
...
-}) {
-  const { STAGE__PR_ID, ITERATE_USER, ESTATE_NAME } = env;
-  if (STAGE__PR_ID) {
-    return `pr-${STAGE__PR_ID}`;
+/**
+ * Get the project name for posthog and braintrust.
+ * To differentiate between different estates, we would add that as metadata to the logs.
+ * For example, in Posthog, we identify ourselves under an estate, and in Braintrust, we add it to the span as metadata.
+ */
+export function getProjectName(env: { PROJECT_NAME?: string; ITERATE_USER: string }) {
```
mmkal: why make `PROJECT_NAME` optional?

apps/os/evals/basic-conversation.eval.ts
```diff
@@ -24,24 +23,41 @@ evalite("agent knows when to end their turn", {
           }),
         },
       },
+      {
+        input: {
+          slug: "multi-turn-vegetables-conversation",
```
mmkal: do we really need this? was it just for debugging?

apps/os/evals/helpers.ts
```diff
@@ -213,3 +231,74 @@ export type WaitForEvent = {
     },
   ): Promise<Selection>;
 };
+
+export function evaliterate<TInput, TOutput, TExpected>(
```
mmkal, line 291: let's add a brief jsdoc explaining why this exists.

apps/os/evals/helpers.ts
```diff
@@ -213,3 +231,74 @@ export type WaitForEvent = {
...
+      const score = await scorerOpts.scorer(result);
+
+      const braintrustSpan = spanMap[hash(result.input)];
+      braintrustSpan?.log({
+        scores: {
+          [scorerOpts.name]: typeof score === "number" ? score : score.score,
+        },
+        metadata: {
+          [scorerOpts.name]: typeof score === "number" ? {} : score.metadata,
+        },
```
mmkal:
```suggestion
        metadata:  typeof score === "number" ? undefined : {
          [scorerOpts.name]: score.metadata,
        },
```

feels like it'll be a little less annoying

apps/os/evals/helpers.ts
```diff
@@ -213,3 +231,74 @@ export type WaitForEvent = {
...
+
+      return score;
+    };
+    return {
+      ...scorerOpts,
+      scorer: wrappedScorerFn,
+    };
+  };
+
+  return evalite<TInput, TOutput, TExpected>(experimentName, {
```
mmkal: nit: isn't `evalite` a void function? can we just call it and not return?

apps/os/env.ts
```diff
@@ -18,6 +18,7 @@ export type CloudflareEnv = Env & {
   EXPIRING_URLS_SIGNING_KEY: string;
   GITHUB_WEBHOOK_SECRET: string;
   STAGE__PR_ID?: string;
+  PROJECT_NAME?: string;
```
mmkal: ?
```suggestion
  PROJECT_NAME: string;
```

mmkal (commented):
i think we can simplify getProjectName more now?

mmkal (approved):
looks great now

apps/os/vitest.config.ts
```diff
@@ -6,5 +6,15 @@ export default defineConfig({
     environment: "node",
     include: ["**/*.test.ts"],
     testTimeout: 120_000, // unfortunately evalite reads from this config and isn't configurable. might be possible to run via vanilla vitest some day.
+    provide: {
+      vitestBatchId: `batch-${Date.now()}`,
```
mmkal, line 10: let's address in a followup but not sure if batch id is the right term for this - also does this "stick" when running in watch mode?

is there any prior art for this kind of thing (like how vitest refer to "a run" in their docs?)?

PR #152, opened 2025-10-01
rm serialised callable stuff

mmkal:


apps/os/backend/utils/type-helpers.ts
```diff
@@ -1,12 +1,8 @@
-import type {
-  AnyRouter,
-  BuiltRouter,
-  MutationProcedure,
-  QueryProcedure,
-} from "@trpc/server/unstable-core-do-not-import";
-import type { AnyTRPCRootTypes, AnyTRPCRouter } from "@trpc/server";
+import type { MutationProcedure, QueryProcedure } from "@trpc/server/unstable-core-do-not-import";
```
jonastemplestein:
```suggestion
```

jonastemplestein (approved):
Never hit approve faster in my life! 

That said, I do think this concept of a "serializable thing that can be called in our system" will likely be quite useful again in the future. But then we can just bring back a more lightweight version of it

apps/os/backend/utils/type-helpers.ts
mmkal: oh damn i guess we have lint warnings off. will turn them on in another pr

PR #153, opened 2025-10-01
tool evals

mmkal:
~not~ ready


apps/os/evals/tools.eval.ts
```diff
@@ -0,0 +1,138 @@
...
+evalite("tool usage", {
+  data: async () => {
+    return [
+      {
+        input: {
+          slug: "reverse-string",
+          messages: [
+            {
+              message: "reverse the string 'supercalifragilisticexpialidocious'",
+              expected: "suoicodilaipxecitsiligarfilacrepus",
```
jonastemplestein: I know this isn't ready, but just a quick note: I've you're trying to test our plumbing, instead of the LLM, (i.e. expecting a 100% pass rate on the eval), it's best not to even show the LLM the string. Otherwise the LLM sometimes just reverses it

I'd use the pass through args mechanism and some version of the flexible test tool to construct a tool that returns something static. I think we can also get rid of the ping pong and reverse tools and consolidate into this test tool

I would probably add an array of tool specs to the input scenarios to make this a bit easier

```ts
          slug: "return-secret",
          tools: [createDOToolFactory(iterateAgentTools).flexibleTestTool({
            overrideName: "test_tool",
            passThroughArgs: {
              behaviour: "return-secret",
              secret: "rumplestiltskin"
            }
          })]
          messages: [
            {
              message:
                'call test_tool and tell me the value of serverSecret',
              expected: "rumplestiltskin",
            },
          ],
```

In general, expressing a more specific tool in terms of the arguments of a more generic tool is quite powerful. This is one way in which we already allow users to "create their own tools"

mmkal: Yeah agreed I don't think the reverse tool is a good test because we'll never really be sure it was the tool call vs the LLM on its own. We can do what we used to do and have a look at the low-level tool call output and make sure the tool was called but passThroughArgs is nicer. The reason I added rumplestiltskin though was as a way of having the equivalent of passthroughArgs but without being dependent on passThroughArgs (I can imagine us breaking passThroughArgs locally if we try to refactor it). Anyway yeah basically I agre with you, I'll keep poking at this.

jonastemplestein: > I can imagine us breaking passThroughArgs locally if we try to refactor it

what do you mean by this? in my mind the ability to create a specific tool from a generic one by setting some or all of it's inputs is one of the key concepts in the tool calling system

if there's a reason not to use the `passThroughArgs` implementation of this idea, we should probably address it rather than avoiding use of the concept. what did you have in mind for a refactor?

i'm not sure if you saw at the time, but i also refactored the old agent e2e tests you'd written (i think for parallel tool calling?) to use passThroughArgs more

mmkal: > what do you mean by this? in my mind the ability to create a specific tool from a generic one by setting some or all of it's inputs is one of the key concepts in the tool calling system

e.g. if we're tinkering locally and we change the zod schema for passThroughArgs itself or the pass-through-args.ts implementation, it could help debug that if we see the only tests that didn't break are the ones that happen not to use it.

>i'm not sure if you saw at the time, but i also refactored the old agent e2e tests you'd written (i think for parallel tool calling?) to use passThroughArgs more

Yep I saw, makes sense. I like passThroughArgs and want to use it! Just wanted a dead-simple test too. For debugging, having less abstraction to figure out will make it easier.

Honestly though this is not a big deal to me, I can remove rumplestiltskin, ping-pong, and reverse tool - I only see this making any difference when debugging a locally-broken test on a branch. I agree we need passThroughArgs functional and shouldn't avoid.

apps/os/backend/agent/iterate-agent-tools.ts
```diff
@@ -12,25 +12,27 @@ export const iterateAgentTools = defineDOTools({
...
-        behaviour: z.literal("raise-error"),
-        error: z.string().describe("Error message to throw"),
-      }),
-      z.object({
-        behaviour: z.literal("return-secret"),
-        secret: z.string().describe("Secret value to return"),
-      }),
-    ]),
+    input: z.object({
+      params: z.discriminatedUnion("behaviour", [
```
mmkal, line 16: I didn't get to the bottom of this but got 400s when using a discriminated union at the root level - @jonastemplestein do you know how this was working before?

jonastemplestein: openai just forbids it for some reason IIRC

jonastemplestein: don't think it was ever working - this might not have come up before because i normally use the flexible test tool with passThroughargs

mmkal: I can open a PR reproducing this after this is in

apps/os/package.json
```diff
@@ -7,8 +7,9 @@
     "cf-typegen": "wrangler types --no-include-runtime",
     "deploy": "doppler run -- pnpm run build && doppler run -- wrangler deploy",
     "dev": "./scripts/dev-wrapper.sh",
-    "evals": "doppler run -- evalite run-once && tsx evals/evalite-exporter.ts",
-    "evals:watch": "doppler run -- evalite watch",
+    "eval": "doppler run -- evalite run-once && tsx evals/evalite-exporter.ts",
```
jonastemplestein, line 10: How often does it run each eval? In my experience with evals you always invoke the runner with some argument to indicate how often you want to run the evals?

Or, in more fancy circumstances, you might say something like "Run this eval as many times as is necessary to gain 98% confidence that you have not made eval scores worse than the reference branch, but no more than 50 times"

apps/os/package.json
```diff
@@ -7,8 +7,9 @@
...
+    "eval:serve": "pnpx serve ignoreme/evalite-ui",
+    "eval:watch": "doppler run -- evalite watch",
```
jonastemplestein, line 12: The inclusion of `evalite watch` is a bit of an orange flag to me with evalite. Evals are not responsive enough for this to be useful IMO - they aren't unit tests. But maybe I'm just thinking about it wrong

jonastemplestein (approved):
Nice! The eval file looks very tidy!

Please can you include some documentation for humans and LLMs about how to run evals in our project? I think this might be sufficiently important for the root README, but I'd also drop something in llms.ts 

For example, I didn't realise I'd have to start the dev server myself

And when running `pnpm eval:serve` from the apps/os folder, a server starts on port 3000 but it just responds with 404 to `GET /`. So I assume I'm doing something wrong

apps/os/package.json
mmkal: it's pretty useful when you're working on evals themselves tho?

apps/os/package.json
mmkal: once, but matt himself plans(?) to add it: https://github.com/mattpocock/evalite/issues/12. tbh I see evalite as just a starting point for us, only a small headstart for our own thing. either as a forkable repo, something that is actively developed that we can contribute to, or just a reference implementation for the CRUD needed for a UI + mini DB.

resolving because this is a legit concern with evalite generally but I don't think addressing it belongs on the `tool evals` PR

PR #155, opened 2025-10-01
Avoid unfurling os.iterate.com authorization links

jonastemplestein:
## Summary
- prevent Slack messages from auto-unfurling when they contain os.iterate.com links
- ensure the sendSlackMessage helper recognises these links even when unfurling is explicitly requested

## Testing
- not run

------
https://chatgpt.com/codex/tasks/task_b_68dbeb2829c08333b84e463ce95cdb0e

### DIFF
--- a/apps/os/backend/agent/slack-agent.ts
+++ b/apps/os/backend/agent/slack-agent.ts
@@ -668,8 +668,17 @@ export class SlackAgent extends IterateAgent<SlackAgentSlices> implements ToolsI
 
     const { endTurn, ...sendInput } = input;
 
-    const numLinks = sendInput.text.match(/https?:\/\/[^\s|]+/g)?.length;
-    const doUnfurl = sendInput.unfurl === "all" || (sendInput.unfurl === "auto" && numLinks === 1);
+    const links = sendInput.text.match(/https?:\/\/[^\s|]+/g) ?? [];
+    const hasOsIterateLink = links.some((link) => {
+      try {
+        return new URL(link).hostname === "os.iterate.com";
+      } catch {
+        return false;
+      }
+    });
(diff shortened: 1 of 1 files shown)


apps/os/backend/agent/slack-agent.ts
```diff
@@ -668,8 +668,17 @@ export class SlackAgent extends IterateAgent<SlackAgentSlices> implements ToolsI
...
+    const hasOsIterateLink = links.some((link) => {
+      try {
+        return new URL(link).hostname === "os.iterate.com";
+      } catch {
+        return false;
+      }
+    });
+    const doUnfurl =
+      !hasOsIterateLink &&
+      (sendInput.unfurl === "all" || (sendInput.unfurl === "auto" && links.length === 1));
```
mmkal: IMO we should have an unfurl whitelist! Start with x.com and uuuh I dunno almost nothing else?

jonastemplestein: I don't agree with this - for example individual linear tickets unfurl really nicely and have useful buttons for interacting with the ticket

But I've asked codex to pull out the logic into an individually testable helper that we can add to over time as we come across annoying examples in agent traces

mmkal (approved):

apps/os/backend/agent/slack-agent.ts
mmkal: sure. I hate unfurling 99% of the time but maybe it's just me

jonastemplestein: i'm also okay with a whitelist if you prefer, but i'd put linear (and anything else that has "rich" unfurling in slack) in there, provided it's only a single link

PR #162, opened 2025-10-01
Fix thread_ts extraction, and make botUserId query suspenseful

myst-6:
Fixes:
- Multiple slack agents being created in an IM thread
- Bot messages being rendered as user messages in the `agents.tsx` file transiently

### DIFF
--- a/apps/os/backend/agent/slack-agent-utils.ts
+++ b/apps/os/backend/agent/slack-agent-utils.ts
@@ -41,8 +41,23 @@ export async function getMessageMetadata(
   switch (slackEvent.type) {
     case "app_mention":
     case "message": {
-      const threadTs =
-        "thread_ts" in slackEvent && slackEvent.thread_ts ? slackEvent.thread_ts : ts;
+      let threadTs: string | undefined;
+      // case 1: we are inside a thread, so thread_ts is specified
+      if ("thread_ts" in slackEvent && slackEvent.thread_ts) {
+        threadTs = slackEvent.thread_ts;
+      }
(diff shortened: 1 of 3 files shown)


apps/os/backend/agent/slack-agent-utils.ts
```diff
@@ -41,8 +41,23 @@ export async function getMessageMetadata(
   switch (slackEvent.type) {
     case "app_mention":
     case "message": {
-      const threadTs =
-        "thread_ts" in slackEvent && slackEvent.thread_ts ? slackEvent.thread_ts : ts;
+      let threadTs: string | undefined;
```
jonastemplestein, line 44: Could you add tests to this function? It would be good to have a test file where we can just add real slack webhook payloads that we encounter to make sure there are never any regressions

jonastemplestein (approved):
Would be good to add a quick test file, but otherwise good to go (and happy for test to be a fast follow)

I do have one question though: do these webhooks get shown to the LLM or not? And if so, do they interrupt LLM requests? 

I think my preference would be that we 1) record all slack webhooks that are clearly about an agent thread as events on that agent and 2) filter out unused events in the reducer

I wouldn't expect these events here to be shown to the LLM at all and wouldn't want them to interrupt LLM requests. But it's still occasionally useful to store them in the event log

apps/os/backend/agent/slack-agent-utils.ts
myst-6: Done!

PR #167, opened 2025-10-01
Ignore unknown users

zakhardavydov:
- Slack workspaces can contain users from external organizations or guests who shouldn't have full access to the estate's agent and integrations.
- Any Slack user could message the bot and trigger LLM requests, and there was no automatic mapping of Slack users to organization memberships, allowing unauthorized access to estate resources and MCP server connections.
- This PR adds multi-table authorization checks that verify Slack users belong to the estate's organization before allowing participation, auto-sync Slack users to organization memberships (with guest/member roles based on Slack restrictions), prevent LLM triggers from unauthorized users, and block guests from connecting MCP servers with graceful error messages.


apps/os/backend/agent/slack-agent.ts
```diff
@@ -358,25 +362,48 @@ export class SlackAgent extends IterateAgent<SlackAgentSlices> implements ToolsI
...
+          eq(providerUserMapping.providerId, "slack-bot"),
+          eq(providerUserMapping.externalId, slackUserId),
+          eq(estate.id, estateId),
+        ),
+      )
+      .limit(1);
 
-    if (!userMapping?.internalUser) {
+    // If no result, user either doesn't exist or doesn't have access to this estate
+    if (!result[0]) {
```
jonastemplestein, line 390: Would it make sense to log something here and also below? Or is this a very common situation?

jonastemplestein:
Sorry, I didn't mean to add an approving review 🤣 I don't know enough about how participant management works - just accidentally clicked approve I think

apps/os/backend/agent/slack-agent.ts
zakhardavydov: Should be rare, will add the logging

apps/os/backend/agent/slack-agent.ts
```diff
@@ -358,36 +364,69 @@ export class SlackAgent extends IterateAgent<SlackAgentSlices> implements ToolsI
...
+      )
+      .limit(1);
+
+    // If no result, user either doesn't exist or doesn't have access to this estate
+    if (!result[0]) {
+      waitUntil(
+        this.slackAPI.chat.postMessage({
+          channel: slackChannelId,
+          thread_ts: slackThreadId,
+          text: `Messaging iterate is not available for you now.`,
```
jonastemplestein: Can we just remove this message? I think the most important thing is that the MCP tool is not called by guests, right? I don't think this here is a good user experience

See this slack message for more context https://iterate-com.slack.com/archives/C08R1SMTZGD/p1759329022030529?thread_ts=1759328429.403659&cid=C08R1SMTZGD

jonastemplestein (changes requested):

apps/os/backend/agent/iterate-agent.ts
```diff
@@ -735,6 +735,31 @@ export class IterateAgent<Slices extends readonly AgentCoreSlice[] = CoreAgentSl
     return dedupedRules;
   }
 
+  /**
+   * Check if a user is a guest in the organization that owns this estate.
+   * Returns true if the user is a guest, false otherwise.
+   */
+  private async isUserGuest(userId: string): Promise<boolean> {
```
mmkal: This will return true if the user isn't found at all is that what we really want? We risk letting nobodies do privileged stuff that way 

Maybe we could just make this a getUserRole fn and do an exhaustive match.

apps/os/backend/agent/slack-agent.ts
```diff
@@ -47,7 +51,7 @@ import { createSlackAPIMock } from "./slack-api-mock.ts";
 // Inherit generic static helpers from IterateAgent
 
 // memorySlice removed for now
-const slackAgentSlices = [...CORE_AGENT_SLICES, slackSlice] as const;
```
mmkal, line 47: Hmm why remove as const? Does this affect type inference?

apps/os/backend/agent/slack-agent.ts
mmkal: Agreed. Maybe a 🙅 reaction if the bot is @ mentioned

apps/os/backend/agent/slack-agent.ts
zakhardavydov: Putting it back

zakhardavydov: I don't think I encountered any issues, but we keep the iterate agent as const, so would be more consistent to keep const here too

mmkal (approved):

PR #170, opened 2025-10-02
Coding sandbox (WIP)

plesiv:
# Current status

Works end-to-end. Needs cleanup.

# Tasks:

- [x] Reuse sandbox setup from `apps/os/backend/sandbox`
- [x] Use a fixed pool of sandbox containers
- [x] Use sandbox sessions for isolation
- [x] Can run shell commands

---

- [x] remove container reuse with sessions (use sessions only for isolation from sandbox code)
- [x] tidy up / deduplicate
- [x] correctly detect if sandbox is running
    - ❗**ISSUES currently: check the comments**
- [x] correctly start it up

---

- [x] exec and return immediately if container is running
- [x] when container boots up, exec and report result
    - ❗**BROKEN currently: check the code**
- [x] use CORE:SET_METADATA for sandboxStatus = starting | attached
    * [x] "context rules" matches this and expands the context for the user

---

- [x] use gh to clone repo
- [x] use gh to list repo contents
- [x] use gh to push to a branch
- [x] use gh to create PR for repo

---

- [ ] load file from sandbox into slack (ephemeral TODO.md)
- [ ] generate image (imagemagick) and load into slack
- [ ] ability to clean the session directory in the sandbox
  * currently persistent for the run of the container
- [ ] correctly handle multiple `exec` state between subsequent commands (e.g. `cd` to a directory)
- [ ] correctly handle multiple `exec` commands in parallel (locking)
- [ ] fix `TODO`s
 
# Potential improvements

- [ ] ensure LLM can do this end-to-end in one go (currently it gets confused):
  1. edit files in the repo,
  2. commit,
  3. push to branch,
  4. and raise a PR (end-to-end)
- [ ] make file editing multi-turn
  * [ ] after each turn load modified file into slack
  * [ ] after each turn ask if we should push to branch + raise a PR

# Research:

- My first attempt at having OpenAI models recognize `local_shell` tool failed (https://x.com/aisdk/status/1973027692942671872)

### DIFF
--- a/apps/os/backend/agent/iterate-agent.ts
+++ b/apps/os/backend/agent/iterate-agent.ts
@@ -1358,7 +1358,7 @@ export class IterateAgent<Slices extends readonly AgentCoreSlice[] = CoreAgentSl
       repository_id: estate.connectedRepoId,
     });
     const githubRepoUrl = repoData.html_url;
-    const branch = estate.connectedRepoRef || "main";
+    const branch = estate.connectedRepoRef || repoData.default_branch || "main";
     const commitHash = undefined; // Use the latest commit on the branch
 
     // ------------------------------------------------------------------------
--- a/apps/os/backend/agent/iterate-agent.ts
+++ b/apps/os/backend/agent/iterate-agent.ts
@@ -1383,6 +1383,18 @@ export class IterateAgent<Slices extends readonly AgentCoreSlice[] = CoreAgentSl
     // Retrieve the sandbox
     const sandbox = getSandbox(env.SANDBOX, sandboxId);
 
+    // Check current state
+    // NOTE: according to the exposed API this should be the correct way to
+    // check if the sandbox is running and start it up if it isnt'. But the logs
+    // are confusing:
+    // ... Sandbox successfully shut down
+    // ... Error checking if container is ready: connect(): Connection refused: container port not found. Make sure you exposed the port in your container definition.
+    // ... Error checking if container is ready: The operation was aborted
+    // ... Port 3000 is ready
+    if ((await sandbox.getState()).status !== "healthy") {
+      await sandbox.startAndWaitForPorts(3000); // default sandbox port
+    }
+
     // Ensure that the session directory exists
     await sandbox.mkdir(sessionDir, { recursive: true });
 
--- a/apps/os/backend/agent/iterate-agent.ts
+++ b/apps/os/backend/agent/iterate-agent.ts
@@ -21,6 +21,7 @@ export type AgentInstanceDatabaseRecord = typeof agentInstance.$inferSelect & {
   iterateConfig: IterateConfig;
 };
 import { makeBraintrustSpan } from "../utils/braintrust-client.ts";
+import { hashStringToRange } from "../utils/utils.ts";
 import { searchWeb, getURLContent } from "../default-tools.ts";
 import { getFilePublicURL, uploadFile, uploadFileFromURL } from "../file-handlers.ts";
 import { tutorialRules } from "../../sdk/tutorial.ts";
@@ -52,6 +53,7 @@ import { defaultContextRules } from "./default-context-rules.ts";
 import { ContextRule } from "./context-schemas.ts";
 import { processPosthogAgentCoreEvent } from "./posthog-event-processor.ts";
 import { triggerEstateRebuild } from "../trpc/routers/estate.ts";
+import { getSandbox } from "@cloudflare/sandbox";
 
 // -----------------------------------------------------------------------------
 // Core slice definition – *always* included for any IterateAgent variant.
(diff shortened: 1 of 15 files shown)


apps/os/backend/sandbox/sandbox-entry.ts
```diff
@@ -11,19 +11,27 @@
 import { spawn } from "child_process";
 import { existsSync } from "fs";
 import { access, constants } from "fs/promises";
+import path from "path";
```
mmkal, line 14: can we use our own dependencies here? or does it have to be vanilla js? if we can use dependencies we should use trpc-cli

apps/os/backend/agent/iterate-agent.ts
```diff
@@ -1286,6 +1280,187 @@ export class IterateAgent<Slices extends readonly AgentCoreSlice[] = CoreAgentSl
...
+    if (!githubInstallation) {
+      throw new Error("No GitHub installation found for this estate");
+    }
+
+    // Get installation token
+    const { getGithubInstallationToken } = await import("../integrations/github/github-utils.ts");
+    const githubToken = await getGithubInstallationToken(githubInstallation.accountId);
+
+    // Fetch repository details from GitHub API
+    const repoResponse = await fetch(
```
mmkal: can we use octokit here?

apps/os/backend/agent/iterate-agent.ts
```diff
@@ -1286,6 +1280,187 @@ export class IterateAgent<Slices extends readonly AgentCoreSlice[] = CoreAgentSl
...
+      },
+    );
+
+    if (!repoResponse.ok) {
+      throw new Error(`Failed to fetch repository details: ${repoResponse.statusText}`);
+    }
+
+    const repoData = (await repoResponse.json()) as { html_url: string };
+    const githubRepoUrl = repoData.html_url;
+    const branch = estate.connectedRepoRef || "main";
```
mmkal: could use github api to get the default branch here (might be master/dev/trunk/canary/next etc.)

apps/os/backend/agent/iterate-agent.ts
```diff
@@ -1286,6 +1280,187 @@ export class IterateAgent<Slices extends readonly AgentCoreSlice[] = CoreAgentSl
...
+      waitUntil(
+        sandbox
+          .startAndWaitForPorts(3000) // default sandbox port
+          .then(execInSandbox)
+          .then((resultExec) => {
+            console.log("TO-REMOVE-LATER: received exec result in callback");
+
+            // Inject a tool call event that will be processed by the agent
+            // TODO: this doesn't work
+            (this as any).addEvent({
```
mmkal, line 1435: do we need the cast to any? does this actually not work or is this copy/paste?

apps/os/backend/agent/iterate-agent.ts
```diff
@@ -1286,6 +1280,187 @@ export class IterateAgent<Slices extends readonly AgentCoreSlice[] = CoreAgentSl
...
+                },
+              },
+              triggerLLMRequest: false,
+            });
+          }),
+      );
+
+      console.log("TO-REMOVE-LATER: sandbox is not yet running, so we wait for it to boot up");
+
+      // TODO: this doesn't work
```
mmkal, line 1458: does it not?

apps/os/backend/sandbox/sandbox-entry.ts
plesiv: Do you mean for CLI argument parsing? My reasoning was that since this script runs in the Docker sandbox and is isolated from the backend itself, that the approach that was already in place (script was previously called `build-script.ts`) was sufficient. But we could upgrade to `trpc-cli` if necessary.

apps/os/backend/agent/iterate-agent.ts
plesiv: OK, ~~I'll fix across the codebase.~~

apps/os/backend/agent/iterate-agent.ts
plesiv: OK, ~~I'll fix across the codebase.~~

apps/os/backend/agent/iterate-agent.ts
plesiv: This is the part that I couldn't get to work, so it's still a bit hacky. The plan is to pair with Jonas to get it working and then fix the types.

apps/os/backend/agent/iterate-agent.ts
plesiv: I'll pair with Jonas on this.

apps/os/backend/agent/iterate-agent.ts
plesiv: @mmkal Done here https://github.com/iterate-com/iterate/pull/170/commits/14ab58c61a3ffdc04ee29a11580731a25adafce5 . I won't look to change other places, since I won't be able to confirm all code paths.

This might be a good issue to test with the newly set up coding agents @mmkal @jonastemplestein

apps/os/backend/agent/iterate-agent.ts
plesiv: Fixed here: https://github.com/iterate-com/iterate/pull/170/commits/578012bdff78732c70591a7c0b1f5a92af9bccb4

PR #171, opened 2025-10-02
allow impersonating by estate id

mmkal:
Opening this on my phone, so draft til I look over what I've done


jonastemplestein (approved):
(I realise its draft but want to make sure you're not blocked on merging later)

This sort of thing you can honestly merge at will as far as I'm concerned. But if you're vibe-coding, I think you can also easily add a autocomplete / typeahead search for users, estates, whatever

or just add a cmd + k menu to the whole app that has a nice impersonate user flow - that's probably what we should do eventually

PR #175, opened 2025-10-02
chore: Add cursor background agent setup

NickBlow:


scripts/setup-background-agent/setup-ubuntu-base.sh
mmkal, line 1: This seems like a looot of setup - and ubuntu is a huge base image. It all looks v standard except for Doppler. Is this the recommended setup?

mmkal: Maybe I just need more context on the why (pr body?)

I am traumatised by spending an hour trying to fix our sandbox dockerfile and finding removing everything from it worked fine

NickBlow: yeah this is the official doppler install script from their website.

I get it - i'll see if there's anything i can trim out :D

mmkal: If there's a link to where this was copy pasted from in a comment that'd be fine

jonastemplestein, line 1: What's the reasoning for putting some stuff here and some in the Dockerfile? 

I think this could also be slimmed down a lot if we first add apt sources and then update once and then install everything we need

jonastemplestein (approved):
It's not pretty but I don't really care so long as it works

We can always make it better later

I think I would prefer for all the involved filenames and comments to make it very clear that this particular cruft is specifically for the cursor agent

scripts/setup-background-agent/setup-ubuntu-base.sh
NickBlow: Some stuff needs to be set up as root, this script runs as the user. i guess we could run it all in the script?

This is the structure that andy had set up

PR #181, opened 2025-10-02
Agent detail export

myst-6:
- [ITE-3018: Agent detail export](https://linear.app/iterate-com/issue/ITE-3018/agent-detail-export)
- [Notion design doc](https://www.notion.so/nustom/Agent-details-export-27f622a338eb80fc8968f0127062ce97)
- The same agent detail component is re-used for both the online and offline versions
- On the online version, you can export the trace which will download an archive.
- On the agents index page, you can navigate to the offline page and import one of these archives. However, mutations are disabled. But there is also a 'view metadata' button where you can look at the metadata of the export.


myst-6:
Looks like I have to rebase as well 😭

apps/os/backend/agent/agent-export-types.ts
```diff
@@ -0,0 +1,62 @@
+import z from "zod";
+import type { AgentCoreEvent, AugmentedCoreReducedState } from "./agent-core-schemas.ts";
+
+// TODO: put these in the right place
+//  and fix the schemas and types
+export interface AgentTraceExportMetadata {
+  estateId: string;
```
jonastemplestein: I'd add owner records here

Might also be easier to just drop in the actual DB rows?

- user
- organisation
- estate
- iterate agent
- maybe even iterate config?

apps/os/backend/agent/agent-export-types.ts
```diff
@@ -0,0 +1,62 @@
...
+});
+
+export const FileMetadataSchema = z.object({
+  filename: z.string().nullable(),
+  mimeType: z.string().nullable(),
+  fileSize: z.number().nullable(),
+});
+
+export const AgentTraceExportSchema = z.object({
+  version: z.string(),
```
jonastemplestein: good shout - unlike agents, i don't really want to brick these things here

apps/os/backend/agent/agent-export-types.ts
myst-6, line 1: Not sure if these types can be moved into a different file or if it's fine to keep here?

apps/os/public/user.svg
jonastemplestein, line 1: can we use lucide icons instead?

jonastemplestein (approved):
One tiny comment (not a merge blocker)

I am supportive of the overall design and integration points, but the actual implementation of the archiving logic and format is a bit above my paygrade. Would be good to review with Misha next week after he's back

I'm really excited to do two things after this 

1. Would be cool if we could make an SPA / static build of an app at traces.iterate.com that can take the URL of a trace archive in a query param, download it and load it

2. Make evals produce viewable traces that are linked from github actions using artifact.ci

PR #182, opened 2025-10-02
feat: add slack web client wrapper

jonastemplestein:
- add a reusable Slack WebClient factory that surfaces non-200 HTTP responses as errors
- update backend Slack integrations to construct clients through the wrapper to ensure consistent error handling

Until we're all-in on effect or something, I prefer this, because it means failed slack API calls come to our attention. At the moment they are just quietly dropped in a few places

### DIFF
--- a/apps/os/backend/integrations/slack/slack-web-client.ts
+++ b/apps/os/backend/integrations/slack/slack-web-client.ts
@@ -1,4 +1,4 @@
-import { ErrorCode, WebClient, type WebClientOptions } from "@slack/web-api";
+import { WebClient, type WebClientOptions } from "@slack/web-api";
 import type { WebAPICallResult } from "@slack/web-api/dist/WebClient";
 import dedent from "dedent";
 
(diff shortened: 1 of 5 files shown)


apps/os/backend/integrations/slack/slack-web-client.ts
```diff
@@ -0,0 +1,55 @@
...
+
+      return result;
+    } catch (error) {
+      if (isSlackHttpError(error)) {
+        const statusMessage =
+          error.statusCode !== undefined
+            ? `HTTP status ${error.statusCode}`
+            : "a non-200 HTTP response";
+
+        throw new Error(`Slack API call to ${method} failed with ${statusMessage}`, {
```
mmkal: What do these errors look like as-is? Just asking because if the error already has the status and method and endpoint this try/catch boilerplate isn't really adding anything?

jonastemplestein: Hm yeah that is unnecessary. The client currently doesn't throw at all even if there's an error 500

jonastemplestein:
turns out this does actually throw, but not always - weird. but not important enough to look into more

PR #183, opened 2025-10-02
feat: Send errors to posthog

NickBlow:


mmkal:
I agree 1. is better - in general we shouldn't use errors for control flow, so we should throw (and catch!) as little as possible and rely on top-level error handlers like this

jonastemplestein (approved):
This is a great start <3 

I'd also like to track client side errors and any CORE:INTERNAL_ERROR in our agent

But that can be separate

apps/os/backend/worker.ts
```diff
@@ -39,6 +40,31 @@ export type Variables = {
...
+// Error tracking with PostHog
+app.onError((err, c) => {
+  const posthog = new PostHog(c.env.POSTHOG_PUBLIC_KEY, {
+    host: "https://eu.i.posthog.com",
+  });
+
+  // Get user ID if available
+  const userId = c.get("session")?.user?.id || "anonymous";
+
+  posthog.captureException(err, userId, {
```
jonastemplestein: Not sure about the semantics of `app.onError`, but should this here maybe be inside `waitUntil()`?

NickBlow: Yes this can go inside a wait until, you are correct - will improve performance here

apps/os/backend/worker.ts
```diff
@@ -100,6 +105,28 @@ app.all("/api/agents/:estateId/:className/:agentInstanceName", async (c) => {
...
+          agentInstanceName,
+          path: c.req.path,
+          method: c.req.method,
+          url: c.req.url,
+        });
+
+        await posthog.shutdown();
+      })(),
+    );
+
```
jonastemplestein: this isn't what i meant - remove it

apps/os/backend/worker.ts
```diff
@@ -122,23 +149,27 @@ app.all("/api/trpc/*", (c) => {
...
-        trpcType: type,
-        trpcCode: error.code,
-        trpcInput: input,
-        method: c.req.method,
-        url: c.req.url,
-      });
-
-      posthog.shutdown();
+      // Track error in PostHog (non-blocking)
+      c.executionCtx.waitUntil(
```
jonastemplestein: you can just import waitUntil from cloudflare:workers

also add a note in vibe-rules/llms.ts named "cloudflare-workerd-stuff" that always attaches to .ts files and just says that we are targeting cloudflare workers

then add a few quick notes to say

- use waitUntil to execute something in the background (with code example of how to import)
- import env from `apps/os/env.ts` (and show concrete code example)

jonastemplestein:
@devin are you there? please fix my comments

jonastemplestein (approved):
ok

PR #184, opened 2025-10-02
A bunch of signup improvements and some admin tools

jonastemplestein:
### Tutorial rules

Added some context rules that are always active in repo-less estates, as well as in estates 

### Random stuff

Add /admin area for admin tools

Added notification that is sent to our iterate slack team whenever somebody signs up (and a page in /admin/slack-notification to test this feature). This mechanism is controlled by a new env var `ITERATE_NOTIFICATION_ESTATE_ID`. If this is set, we'll look up this estate and use its slack access token for the notification

Added some vibe-coded admin tools for deleting a user and all the organisations and estates it owns. Useful for local testing

Store entire `iterateConfig` (not just `contextRules` on `IterateAgent` DO)

### Updates to iterate.com website

Add new button to add directly to slack (and a bit of other tidy up)

CleanShot 2025-10-02 at 16 55 23@2x


jonastemplestein:
* **#184** <a href="https://app.graphite.dev/github/pr/iterate-com/iterate/184?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/iterate-com/iterate/184?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

apps/os/backend/auth/hooks.ts
```diff
@@ -44,6 +48,37 @@ export const createUserOrganizationAndEstate = async (db: DB, userId: string, us
...
 };
+
+async function sendEstateCreatedNotificationToSlack(
+  organization: typeof schema.organization.$inferSelect,
+  estate: typeof schema.estate.$inferSelect,
+) {
+  // Construct the impersonation link
+  const impersonationUrl = `${env.VITE_PUBLIC_URL}/${organization.id}/${estate.id}`;
+
+  const message = `🎉 New user signed up!
```
NickBlow, line 73: *very* minor nitpick, we already have dedent - maybe use that to make this less ugly.

package.json
```diff
@@ -25,6 +25,7 @@
     "db:studio:production": "PSCALE_DATABASE_URL=$(doppler secrets get --plain DRIZZLE_ADMIN_POSTGRES_CONNECTION_STRING) pnpm db:studio",
     "clean": "rm -rf ./**/node_modules",
     "vibe-rules": "vibe-rules install cursor && vibe-rules install codex && vibe-rules install claude-code",
+    "super-reset": "pnpm i && pnpm docker:down && pnpm docker:clean && pnpm docker:up && sleep 3 && pnpm db:migrate && pnpm dev",
```
NickBlow, line 28: lol

apps/os/backend/trpc/routers/admin.ts
```diff
@@ -61,14 +62,151 @@ const getEstateOwner = adminProcedure
...
+    const deletedEstates: string[] = [];
+
+    // Delete all organizations where the user is owner (and their estates)
+    for (const membership of ownerOrganizations) {
+      const org = membership.organization;
+      const estates = org.estates;
+
+      // Delete all estates in the organization
+      for (const estate of estates) {
+        // Delete all related data for each estate
```
mmkal, line 104: if we were writing raw sql i'd say this should use CTEs. With drizzle, I suspect we _could_ write CTEs, but at the very least this should be in a transaction

apps/os/backend/integrations/slack/slack-utils.ts
```diff
@@ -0,0 +1,45 @@
...
+export async function sendNotificationToIterateSlack(
+  message: string,
+  channel = "#building",
+): Promise<void> {
+  const notificationEstateId = env.ITERATE_NOTIFICATION_ESTATE_ID;
+
+  if (!notificationEstateId) {
+    console.warn("ITERATE_NOTIFICATION_ESTATE_ID not configured, skipping notification");
+    return;
+  }
```
mmkal, line 23: Seems weird to tie this to "iterate" when really it could easily be a function that takes an estate id and a channel.

If we really want an iterate-specific helper with this exact signature we could do

```ts
export async function sendNotificationToIterateSlack(message: string, channel = "#building") {
  return sendNotificationToEstateSlack({
    message,
    channel,
    estateId: env.ITERATE_NOTIFICATION_ESTATE_ID,
  })
}
```

where the bulk of this function would be replaced by a more broadly-useful sendNotificationToEstateSlack?

mmkal, line 23: shouldn't we throw here?

apps/os/scripts/pg-estate-bootstrap.ts
```diff
@@ -12,10 +12,31 @@ async function bootstrap() {
   // Check for config path from environment variable
   const iterateConfigPath = process.env.ITERATE_CONFIG_PATH;
 
+  // Connect to the database - use same connection as drizzle config
+  const connectionString = "postgres://postgres:postgres@localhost:5432/iterate";
```
mmkal, line 16: can we not get this from process.env?

apps/os/scripts/pg-estate-bootstrap.ts
```diff
@@ -12,10 +12,31 @@ async function bootstrap() {
...
+
+  const pg = postgres(connectionString, {
+    max: 1,
+    fetch_types: false,
+  });
+
+  const db = drizzle(pg, { schema, casing: "snake_case" });
```
mmkal, line 23: I think we have a `db-cli.ts` module no?

apps/os/sdk/tutorial.ts
mmkal, line 1: separate PRs would make this easier to get through

apps/os/backend/integrations/slack/slack-utils.ts
jonastemplestein: no, the env var controls whether or not the notification gets made

apps/os/backend/trpc/routers/admin.ts
jonastemplestein: we are using cascading deletes now

apps/os/scripts/pg-estate-bootstrap.ts
jonastemplestein: this file doesn't exist on main

apps/os/backend/integrations/slack/slack-utils.ts
jonastemplestein: when do you envision this being used for other estates? we have a separate trpc procedure already that e.g. starts an agent in a slack channel with a given message

i just wanted a quick helper that is _explicitly_ just for us to do slack-ops on our own business

PR #190, opened 2025-10-03
fix: Filter irrelevant TRPC errors

NickBlow:


apps/os/backend/worker.ts
```diff
@@ -135,7 +135,9 @@ app.all("/api/trpc/*", (c) => {
           });
 
           const userId = c.var.session?.user?.id || "anonymous";
-
+          if (error.code !== "INTERNAL_SERVER_ERROR") {
+            return;
+          }
```
mmkal: uhh why is this irrelevant sorry?

jonastemplestein: this PR is in response to [this slack thread](https://iterate-com.slack.com/archives/C06LU7PGK0S/p1759436240086099) from yesterday 

the goal is to filter out 4xx series "errors" so they don't pollute posthog error tracking

but not sure this is the correct way (vs e.g. doing it the other way around and dropping all errors that match a certain condition)

mmkal: yep nick and I spoke and he changed to if <500 return

PR #191, opened 2025-10-03
delete iterate configs when no config is specified in pnpm dev

jonastemplestein:


jonastemplestein:
* **#191** <a href="https://app.graphite.dev/github/pr/iterate-com/iterate/191?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/iterate-com/iterate/191?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

apps/os/sdk/cli/commands/dev.ts
```diff
@@ -129,12 +143,14 @@ const start = t.procedure
...
-      input.config || process.env.ITERATE_CONFIG_PATH || "../../estates/template/iterate.config.ts";
+    const providedConfigPath = input.config || process.env.ITERATE_CONFIG_PATH;
 
-    process.env.ITERATE_CONFIG_PATH = providedConfigPath;
-    console.log(`Using iterate config: ${providedConfigPath}`);
+    if (providedConfigPath) {
+      process.env.ITERATE_CONFIG_PATH = providedConfigPath;
+      console.log(`Using iterate config: ${providedConfigPath}`);
+    } else {
+      console.log("No iterate config provided - will delete all configs");
```
mmkal: I don' tthink we need this, and it risks becoming not-true if we change things elsewhere later
```suggestion
```

apps/os/sdk/cli/commands/dev.ts
```diff
@@ -11,12 +11,35 @@ import { t } from "../config.ts";
...
+      fetch_types: false,
+    });
+
+    const db = drizzle(pg, { schema, casing: "snake_case" });
+
+    // If no config path provided, delete all iterate configs
+    if (!configPath) {
+      console.log("No iterate config path provided, deleting all iterate configs");
+
+      await db.delete(schema.iterateConfig);
```
mmkal: given we insert/update based on estate id, shouldn't we also delete based on it? this truncates the whole table which feels pretty scary as a default action, even under dev

jonastemplestein: the logic for the bootstrap just sets iterate config for all estates - @NickBlow and i thought that is easiest to reason about without having to guess which estate is your "main" estate your developing with etc. but i just saw that the logic we had for that can be simplified a lot

PR #200, opened 2025-10-03
import logger as logger

mmkal:
(no more `import { logger as console }`)

cursor did this

prompt:
>find all the places where we're doing `import { logger as console }` and change to just `import { logger }` then change all the `console.whatever(...)` in affected files to `logger.whatever(...)`

work it did 💀:

https://github.com/user-attachments/assets/8219247d-9c02-47df-8eba-6617c83505fe


jonastemplestein (approved):
Okay let's merge quickly before conflicts

Small pro-tip for stuff like this. Just ask it to use sed to replace in all the files and it'll do it all in one command at a fraction of the cost

PR #201, opened 2025-10-03
multi-trial evals

mmkal:


jonastemplestein (approved):
Let's merge this for now, but I think we're not entirely on the right track here

For example, it doesn't look like our implementation results in this UI experience in braintrust: 

https://www.braintrust.dev/docs/guides/experiments/interpret#viewing-data-across-trials

I also don't understand why we need all this mean of means scoring and the `MultiTrialScorerOutput`

I think it might be helpful to come at this problem from the "other direction" and just run e.g. this [code here](https://www.braintrust.dev/docs/guides/experiments/write#trials) and see what data it results in in braintrust

And my instinct is still that evalite is getting in the way a bit. If we have braintrust as a UI anyway, have to understand braintrust's abstractions, and now have to write custom code to work around evalite shortcomings, then why not just make our own lightweight vitest based runner that just wraps the vanilla braintrust `Eval`?

jonastemplestein (approved):
Let's merge this for now, but I think we're not entirely on the right track here

For example, it doesn't look like our implementation results in this UI experience in braintrust: 

https://www.braintrust.dev/docs/guides/experiments/interpret#viewing-data-across-trials

I also don't understand why we need all this mean of means scoring and the `MultiTrialScorerOutput`

I think it might be helpful to come at this problem from the "other direction" and just run e.g. this [code here](https://www.braintrust.dev/docs/guides/experiments/write#trials) and see what data it results in in braintrust

And my instinct is still that evalite is getting in the way a bit. If we have braintrust as a UI anyway, have to understand braintrust's abstractions, and now have to write custom code to work around evalite shortcomings, then why not just make our own lightweight vitest based runner that just wraps the vanilla braintrust `Eval`?

PR #202, opened 2025-10-03
cli: trpc subcommand

mmkal:
- use "proxify" idea https://github.com/mmkal/trpc-cli/pull/153 - basically, swap a trpc router for a router with the same procedures and inputs, but each procedure's implementation is replaced by a `trpcClient` usage
- use an admin client. tried impersonation but it didn't work. so this is local only for now
- shim a few more cloudflare modules so it's possible to import the router

https://github.com/user-attachments/assets/05b630d1-10be-47c5-988b-9cd86b7adb9d


apps/os/sdk/cli/index.ts
```diff
@@ -2,7 +2,12 @@ import { spawnSync } from "node:child_process";
 import { resolve } from "node:path";
 import { createCli } from "trpc-cli";
 import * as prompts from "@clack/prompts";
+import { proxify } from "trpc-cli/dist/proxify";
+import { createTRPCClient, httpLink } from "@trpc/client";
 import { testingRouter } from "../../backend/trpc/routers/testing.ts";
+import { appRouter } from "../../backend/trpc/root.ts";
```
jonastemplestein, line 8: The appRouter contains more or less our entire backend. Do we really want to import that into our CLI? Now anyone writing any backend code needs to make sure it runs in our nodejs CLI or else `iterate trpc` breaks

jonastemplestein (approved):
This is quite impressive! Let's merge because it's awesome to use this locally straight away

I still think we're building ourselves into a corner a bit by importing the `appRouter`. I liked the design in the `os` repo more where we have a schema discovery endpoint on the server and the CLI client just hits that 

But of course that would be a lot more code to write

As for auth, I'll ask @BlankParticle to make a SERVICE_TOKEN + impersonation header combo early next week. And then the _proper_ solution is for our worker to be an oauth provider one day

apps/os/sdk/cli/index.ts
mmkal: it doesn't really need to run, it just needs to be importable. we never run the procedure `.query(...)`/`.mutation(...)` callbacks

PR #204, opened 2025-10-03
trpc admin ui

mmkal:
@jonastemplestein you might find this useful as an alternative to the `iterate trpc` cli command

it uses the `parseRouter` helper from trpc-cli to send info about procedures to the UI, which then uses react-jsonschema-form to render hideous input forms for each procedure. the ui is copied from an admin ui I vibe coded for postcardo.com! demo:

https://github.com/user-attachments/assets/5b17d81b-6747-4156-adf9-b016bf7604ae

feel free to close while I'm out if too scary/too ugly/rahul hates it


apps/os/worker-configuration.d.ts
```diff
@@ -1,6 +1,10 @@
 /* eslint-disable */
-// Generated by Wrangler by running `wrangler types --no-include-runtime` (hash: 0676b7464628dc424304cbf4aa0c939c)
+// Generated by Wrangler by running `wrangler types --no-include-runtime` (hash: 938d4762cee0d88df975eaf6337a7334)
 declare namespace Cloudflare {
+	interface GlobalProps {
+		mainModule: typeof import("./backend/worker");
+		durableNamespaces: "IterateAgent" | "SlackAgent" | "OrganizationWebSocket" | "Sandbox";
+	}
```
jonastemplestein, line 7: A bit confused why this is here suddenly

jonastemplestein (approved):
Love it!

apps/os/worker-configuration.d.ts
mmkal: me too

PR #208, opened 2025-10-03
hack to make evals not fail

mmkal:
guest/unknown participant stuff broke em


mmkal:
FYI @jonastemplestein @zakhardavydov

mmkal:
merging this so if someone wants to push evals forward while i'm out you cna actually run them

PR #212, opened 2025-10-06
chore: Port garple website

NickBlow:


.github/workflows/deploy-garple.yml
jonastemplestein, line 1: Great! Can we do the same to the iterate website? There should be some ticket for it somewhere - cursor, devin, etc can probably all do it now following this example

jonastemplestein (changes requested):
Why did you delete estates/garple/iterate.config.ts?

I think @mmkal 's idea was that estates/garple contains the same folder contents that _a garple-org/iterate repo would contain if garple was a real customer_

So with that in mind, we need to set it up like this 

- estates/garple/iterate.config.ts - this is the config that feeds the iterate estate in prod
- estates/garple/apps/website - this is where the app that you put directly in estates/garple/ should live

And are you making the os worker change separately that allows one git repo to be connected to multiple estates and trigger multiple builds on change? Or does that "just work"?

estates/garple/wrangler.jsonc
```diff
@@ -0,0 +1,32 @@
+{
+  // this is horrible but we need to do it this way because it doesn't seem to work with symlinks
+  "$schema": "../../../node_modules/wrangler/config-schema.json",
+  "name": "estate-garple-production-garple-website",
+  "compatibility_date": "2025-08-13",
+  "compatibility_flags": ["nodejs_compat_v2"],
```
BlankParticle, line 6: I think we just need `nodejs_compat`
https://developers.cloudflare.com/workers/configuration/compatibility-flags/#nodejs-compatibility-flag

.github/workflows/deploy-garple.yml
NickBlow: yep, was gonna do in a separate ticket

PR #229, opened 2025-10-06
Bring back gmail and google calendar tools

myst-6:
My idea was to make a common `callGoogleAPI` tool. We can then just override their input schema and name, and give some `passThroughArgs` for the URL and method for each tool. It worked for most of the tools, except for the following:
- `deleteCalendarEvent`, `updateCalendarEvent`, `deleteGmailLabel`, `modifyGmailLabels`: some of their parameters are actually path parameters, so I can't override the URL with a static string. Solution: implement some dynamic path resolver in the `callGoogleAPI` tool (this seems fairly reasonable to me!).
- `sendGmail`: you need to format the email into a single string and then base64encode it, which an LLM can't do
- `getGmailMessage`: you need to always base64decode a string, and sometimes the string is split into multiple parts and you have to concatenate them

I reckon that implementing the dynamic path resolver and keeping those two extra methods will be good!

Last point of contention is whether we can define the methods inline on the class or have them separately. Before, when I had 7 methods, I thought it made more sense for them to be separate, but now that we only have 3, I think it makes more sense for them to be inline.

### DIFF
--- a/apps/os/backend/agent/default-context-rules.ts
+++ b/apps/os/backend/agent/default-context-rules.ts
@@ -254,6 +254,406 @@ export const defaultContextRules = defineRules([
     match: matchers.sandboxStatus("starting"),
   },
   {
+    key: "google-gmail-tools",
+    prompt: dedent`
+      You have access to Gmail tools to send, read, reply to, and forward emails.
+
+      When sending emails:
+      - Use simple, plain text by default
+      - Only use HTML formatting when explicitly requested
+      - Keep emails concise and professional
+
+      When replying to emails:
+      - Use sendGmail with threadId and inReplyTo (messageId) from getGmailMessage
+      - This keeps the reply in the same thread
+
+      When forwarding emails:
+      - First use getGmailMessage to retrieve the original email
+      - Then use sendGmail with a formatted body including the original message details
+
+      When organizing emails:
+      - Use modifyGmailLabels to add/remove labels from messages
+      - Common system labels: STARRED (star), UNREAD/INBOX (mark unread/read), TRASH (trash), SPAM (mark as spam), IMPORTANT
+      - Use listGmailLabels to see all available labels including custom ones
+      - Use createGmailLabel to create new organizational labels
+
+      When reading emails:
+      - Use list tools with appropriate filters (e.g., is:unread, from:someone@example.com)
+      - Respect user privacy and only access what's needed
+    `,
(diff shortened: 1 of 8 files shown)


apps/os/backend/agent/default-context-rules.ts
jonastemplestein, line 1: This might be better as a separate file

I personally also prefer expressing the schemas as zod and then converting to json schema. But we can clean that up in a follow up PR because I'm excited to play with this here now :D

apps/os/backend/auth/token-utils.ts
```diff
@@ -101,3 +105,128 @@ export const getGithubUserAccessTokenForEstate = async (db: DB, estateId: string
     installationId: result.installationId,
   };
 };
+
+export const GoogleTokenResponse = z.object({
```
jonastemplestein, line 109: Can we move google specific backend code into backend/integrations/google

Now that we have a monolith, we need to be especially mindful of how we structure our codebase

apps/os/backend/auth/oauth-state-schemas.ts
```diff
@@ -24,8 +26,6 @@ export const MCPOAuthState = BaseOAuthState.extend({
   estateId: z.string(),
   userId: z.string(),
   clientId: z.string(),
-  fullUrl: z.string(),
-  agentDurableObject: AgentDurableObjectInfo,
```
jonastemplestein, line 28: Is this change deliberate or a rebase error?

apps/os/backend/agent/iterate-agent.ts
```diff
@@ -2067,6 +2067,196 @@ export class IterateAgent<Slices extends readonly AgentCoreSlice[] = CoreAgentSl
       __addAgentCoreEvents: [{ type: "CORE:SET_METADATA", data: { sandboxStatus: "attached" } }],
     };
   }
+
+  async callGoogleAPI(input: Inputs["callGoogleAPI"]): Promise<Result<unknown>> {
```
jonastemplestein, line 2071: I think it would be better to just have a one-liner here that calls a library function from backend/integrations/google

apps/os/backend/auth/oauth-state-schemas.ts
myst-6: deliberate, moved it into the base oauth state so that the integrations redirect page isn't MCP-specific

PR #237, opened 2025-10-07
Fix agent init params hydration after durable object alarms

jonastemplestein:
This PR fixes a longstanding issue we had with durable object alarms when using the agents SDK's `this.schedule` method

If the alarm wakes a DO from hibernation, that object would not have the estate, iterateAgent, organization and iterateConfig database records set 

This PR is basically a fancy version of the workaround Sunil suggests [here](https://github.com/cloudflare/partykit/issues/32)

While I was at it, I slightly tidied up and clarified a the flow of the code a bit. It also reduces the number of DB and parent worker <> DO roundtrips

- the weird partykit magic fetch call that eventually triggers `onStart()` is now called from within the new `initIterateAgent(params)` (prev called `initAfterConstructorBeforeOnStart`)
- `AgentInitParams` (i.e. the control plane database records for estate, organization, agent instance and iterate config) are now stored in DO storage; onAlarm() rehydrates via storage and re-runs initIterateAgent
- stub getters now do single joined selects (estate/org/iterateConfig); fewer DB roundtrips; route create path wrapped in transaction
- net effect: alarms wake DOs into fully hydrated agents (all DB records) with onStart executed—same as when created via stub management

### DIFF
--- a/apps/os/backend/agent/iterate-agent.ts
+++ b/apps/os/backend/agent/iterate-agent.ts
@@ -26,6 +26,9 @@ export type AgentInitParams = {
   estate: typeof schema.estate.$inferSelect;
   organization: typeof schema.organization.$inferSelect;
   iterateConfig: IterateConfig;
+  // Optional props forwarded to PartyKit when setting the room name
+  // Used to pass initial metadata for the room/server initialisation
+  props?: Record<string, unknown>;
 };
 import { makeBraintrustSpan } from "../utils/braintrust-client.ts";
 import { searchWeb, getURLContent } from "../default-tools.ts";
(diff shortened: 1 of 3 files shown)


apps/os/backend/agent/iterate-agent.ts
```diff
@@ -332,46 +386,74 @@ export class IterateAgent<Slices extends readonly AgentCoreSlice[] = CoreAgentSl
...
+          estate: estateJoined,
+          organization: estateJoined.organization!,
+          iterateConfig,
+        },
+      });
     }
 
     // No existing agent for this route, create one with route
-    const durableObjectName = `SlackAgent-${route}-${crypto.randomUUID()}`;
+    const durableObjectName = `${this.name}-${route}-${crypto.randomUUID()}`;
```
BlankParticle: `this.name` is name of the DO?
then we don't need the random UUID (also if we need it, lets use `typeid` instead for consistency)

apps/os/backend/agent/iterate-agent.ts
```diff
@@ -384,13 +466,51 @@ export class IterateAgent<Slices extends readonly AgentCoreSlice[] = CoreAgentSl
...
 
-  // This gets run between the synchronous durable object constructor and the asynchronous onStart method of the agents SDK
-  async initAfterConstructorBeforeOnStart(params: AgentInitParams) {
     this._databaseRecord = params.record;
     this._estate = params.estate;
     this._organization = params.organization;
     this._iterateConfig = params.iterateConfig;
+
+    // Store init params so onAlarm re-hydration can re-initialise without DB calls
+    await this.ctx.storage.put("agent-init-params", params);
```
BlankParticle: use the SyncKV, the old one is considered legacy now

```suggestion
    this.ctx.storage.kv.put("agent-init-params", params);
```

apps/os/backend/agent/iterate-agent.ts
```diff
@@ -384,13 +466,51 @@ export class IterateAgent<Slices extends readonly AgentCoreSlice[] = CoreAgentSl
...
+    } catch (e) {
+      logger.error("Could not set server name:", e);
+    }
+
+    this._isInitialized = true;
+  }
+
+  override async onAlarm() {
+    logger.info("onAlarm - pulling init params from storage");
+    const params = await this.ctx.storage.get<AgentInitParams>("agent-init-params");
```
BlankParticle:
```suggestion
    const params = this.ctx.storage.kv.get<AgentInitParams>("agent-init-params");
```

apps/os/backend/agent/iterate-agent.ts
jonastemplestein: The route key contains useful info for visually debugging as a human (e.g. slack thread ts)

And this _becomes_ this.name inside the DO. Here we are in a static method on the class

BlankParticle: ok, then lets use `typeid` for consistency sake
else is fine

jonastemplestein: let me call you - i may not understand how typeid works in cases where I want `${route}` to be present in the ID

PR #242, opened 2025-10-07
fix global connection cache

zakhardavydov:
Fixes [ITE-3115: Fix the module-level connection cache](https://linear.app/iterate-com/issue/ITE-3115/fix-the-module-level-connection-cache)

Fixes a critical Cloudflare Workers error where MCP manager I/O objects (streams, connections) were being shared across Durable Object boundaries, causing `"Cannot perform I/O on behalf of a different Durable Object"` errors. Additionally fixes MCP OAuth reconnection to use the correct server ID.

## Problem

When MCP tools executed tool calls to MCP server, ReadableStreams and other I/O objects created in one Durable Object context were being accessed from another DO's context. This violates Cloudflare Workers' isolation model, which requires I/O objects to remain within their originating execution context for performance and correctness.

The root cause was a module-level `mcpManagerCache` shared across all DO instances in the same isolate. When DO-A cached a manager with I/O objects, DO-B would retrieve and attempt to use those objects, triggering the error.

Additionally, the MCP OAuth reconnection flow was using a hardcoded placeholder `"cloudflare-requires-id"` instead of the actual server ID from the initial connection attempt, preventing proper reconnection after OAuth authorization.

### Instance-Level Cache Architecture
- Created `MCPManagerCache` interface with factory function `createMCPManagerCache()`
- Added `mcpManagerCache` instance property to `IterateAgent` and `SlackAgent`
- Updated all MCP functions (`handleMCPConnectRequest`, `handleMCPDisconnectRequest`, `getOrCreateMCPConnection`, `rehydrateExistingMCPConnection`) to accept cache as a parameter
- Updated `LazyConnectionDeps` interface to include `cache` property for tool execution
- Simplified cache key from `durableObjectId--connectionKey` to just `connectionKey` since each DO now has its own cache

**Result**: Each Durable Object maintains its own isolated cache of MCP managers with I/O objects bound to its execution context, eliminating cross-boundary access.

### OAuth Server ID Preservation
- Added `serverId` field to `MCPOAuthState` schema
- Added `serverId` property to `MCPOAuthProvider` class, set after `manager.connect()` returns
- Updated `MCP:OAUTH_REQUIRED` event to include `serverId`
- Updated `reconnect` parameter schema to include `id` field
- Modified OAuth callback to extract and use `state.serverId` for reconnection
- Updated MCP slice reducer to store actual `serverId` instead of empty string

**Result**: OAuth reconnection now uses the correct server ID from the initial connection attempt, enabling proper session resumption.

### DIFF
--- a/apps/os/backend/agent/iterate-agent.ts
+++ b/apps/os/backend/agent/iterate-agent.ts
@@ -46,7 +46,12 @@ import {
   type AugmentedCoreReducedState,
 } from "./agent-core-schemas.ts";
 import type { DOToolDefinitions } from "./do-tools.ts";
-import { runMCPEventHooks, getOrCreateMCPConnection } from "./mcp/mcp-event-hooks.ts";
+import {
+  runMCPEventHooks,
+  getOrCreateMCPConnection,
+  createMCPManagerCache,
+  type MCPManagerCache,
+} from "./mcp/mcp-event-hooks.ts";
 import { mcpSlice, getConnectionKey } from "./mcp/mcp-slice.ts";
 import { MCPConnectRequestEventInput } from "./mcp/mcp-slice.ts";
 import { iterateAgentTools } from "./iterate-agent-tools.ts";
@@ -371,6 +376,8 @@ export class IterateAgent<Slices extends readonly AgentCoreSlice[] = CoreAgentSl
   _organization?: typeof schema.organization.$inferSelect;
   _iterateConfig?: IterateConfig;
 
+  protected mcpManagerCache: MCPManagerCache;
+
   // This gets run between the synchronous durable object constructor and the asynchronous onStart method of the agents SDK
   async initAfterConstructorBeforeOnStart(params: AgentInitParams) {
     this._databaseRecord = params.record;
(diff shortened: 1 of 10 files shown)


apps/os/backend/agent/iterate-agent.ts
```diff
@@ -408,6 +418,9 @@ export class IterateAgent<Slices extends readonly AgentCoreSlice[] = CoreAgentSl
   constructor(ctx: DurableObjectState, env: CloudflareEnv) {
     super(ctx, env);
     this.db = getDb();
+    // Initialize instance-level MCP manager cache and connection queues
+    this.mcpManagerCache = createMCPManagerCache();
+    this.mcpConnectionQueues = createMCPConnectionQueues();
```
myst-6, line 423: isn't it true that if you do this in the constructor, the initialization will not happen after the DO wakes up for a second time, and they will be undefined?

zakhardavydov: No, that's not correct. The constructor is called every time the DO wakes up from hibernation/eviction. So initialization in the constructor will happen on the first wake up, second wake up, and every subsequent wake up.

The constructor runs:
- On first access
- After eviction (10s for hibernatable, 70-140s otherwise)
- When woken by alarms/WebSocket messages
- After code deployments

So variables initialized in the constructor won't be undefined on subsequent wake-ups—they'll be re-initialized each time the constructor runs.

myst-6: oh yeh sorry

PR #243, opened 2025-10-07
Correct OIDC Slack scopes + fix sign-in with Slack user sync

zakhardavydov:
Fixes [ITE-3114: Ditch identity scopes from slack flow](https://linear.app/iterate-com/issue/ITE-3114/ditch-identity-scopes-from-slack-flow)

**1. Updated to OpenID Connect scopes**
- Replaced legacy `identity.*` scopes with modern OIDC: `openid`, `profile`, `email`
- Updated API calls from `users.identity()` to `openid.connect.userInfo()`
- Updated response parsing to use standard OIDC claims

**2. Fixed Slack user sync on sign-in**
- Added auto-connect logic in `redirect.tsx` when users sign in with Slack
- Automatically connects Slack bot to new estate and syncs workspace users
- Creates `providerEstateMapping` with team info from Slack API

### DIFF
--- a/apps/os/app/routes/redirect.tsx
+++ b/apps/os/app/routes/redirect.tsx
@@ -1,12 +1,15 @@
 import { redirect } from "react-router";
-import { eq } from "drizzle-orm";
+import { eq, and } from "drizzle-orm";
 import { waitUntil } from "cloudflare:workers";
+import { WebClient } from "@slack/web-api";
 import { GlobalLoading } from "../components/global-loading.tsx";
 import { getDb } from "../../backend/db/client.ts";
 import { getAuth } from "../../backend/auth/auth.ts";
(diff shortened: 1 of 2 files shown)


apps/os/app/routes/redirect.tsx
jonastemplestein, line 1: @NickBlow will be better placed to advise on this, but to me it feels like we're increasingly abusing this file

It started with moving the org/estate generation from better auth hooks in here. Then we added stripe customer creation. And now this...

I would have thought the purpose of this file is to redirect from `/` to `/:orgId/:estateId` and not much else

@NickBlow do you have any ideas for a better design?

jonastemplestein (approved):
One comment but not a merge blocker

PR #246, opened 2025-10-07
Remove all waitUntil inside durable objects

myst-6:
[ITE-3112: get rid of unnecessary waitUntil in durable objects](https://linear.app/iterate-com/issue/ITE-3112/get-rid-of-unnecessary-waituntil-in-durable-objects)

### DIFF
--- a/apps/os/backend/agent/agent-core.ts
+++ b/apps/os/backend/agent/agent-core.ts
@@ -131,7 +131,7 @@ export type CheckDepsConflict<T> =
 export interface AgentCoreDeps {
   /** Persist the full event array whenever it changes – safe to store by ref */
   storeEvents(events: ReadonlyArray<AgentCoreEvent>): void;
-  /** Run a background task (e.g. DurableObject ctx.waitUntil) */
+  /** Run a background task */
   background(fn: () => Promise<void>): void;
   /** Lazily construct an OpenAI client */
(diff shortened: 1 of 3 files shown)


apps/os/backend/agent/agent-core.ts
```diff
@@ -131,7 +131,7 @@ export type CheckDepsConflict<T> =
 export interface AgentCoreDeps {
   /** Persist the full event array whenever it changes – safe to store by ref */
   storeEvents(events: ReadonlyArray<AgentCoreEvent>): void;
-  /** Run a background task (e.g. DurableObject ctx.waitUntil) */
+  /** Run a background task */
   background(fn: () => Promise<void>): void;
```
myst-6, line 135: i suppose its possible that we run agent core from a different context outside of durable objects so I think it should still be dependency injected

jonastemplestein (approved):
Thanks for picking this up! I think these are all floating promises now so when Misha turns the "no floating promises" rule on, we'll be in trouble

But I think you can explicitly void them, which also makes the code clearer

`void someFunctionThatReturnsAPromise()` 

I'd change that but feel free to merge straight after

mmkal:
note to me so i'll see on git blame:

we should add these all back! `waitUntil` is now wrapped and errors go to posthog

(having a global rejected promise handler is an alternative but depends on more framework magic)

PR #247, opened 2025-10-07
external users, slack rules

zakhardavydov:
Fixes:
1. [ITE-3065: Create "slack channel name" cache, store slack channel cache in slack slice state and make it possible to match on slack channel name in context rules](https://linear.app/iterate-com/issue/ITE-3065/create-slack-channel-name-cache-store-slack-channel-cache-in-slack)
2. [ITE-3117: Make sure slack connect users are treated as guests](https://linear.app/iterate-com/issue/ITE-3117/make-sure-slack-connect-users-are-treated-as-guests)

### DIFF
--- a/.github/workflows/pkg-pr.yml
+++ b/.github/workflows/pkg-pr.yml
@@ -76,7 +76,7 @@ jobs:
           echo "Using commit hash: $COMMIT_HASH"
 
           # Replace workspace:* with pkg.pr.new URL
-          sed -i "s|\"@iterate-com/sdk\": \"workspace:\*\"|\"@iterate-com/sdk\": \"https://pkg.pr.new/iterate-com/iterate/@iterate-com/sdk@$COMMIT_HASH\"|g" package.json
+          sed -i "s|\"@iterate-com/sdk\": \"workspace:\*\"|\"@iterate-com/sdk\": \"https://pkg.pr.new/iterate/iterate/@iterate-com/sdk@$COMMIT_HASH\"|g" package.json
 
           # Show the change
           echo "Updated package.json:"
@@ -97,6 +97,6 @@ jobs:
           if git diff --staged --quiet; then
             echo "No changes to sync"
           else
-            git commit -m "Sync from iterate-com/iterate@$COMMIT_HASH"
+            git commit -m "Sync from iterate/iterate@$COMMIT_HASH"
             git push
           fi
--- a/apps/os/app/routes/org/estate/integrations/index.tsx
+++ b/apps/os/app/routes/org/estate/integrations/index.tsx
@@ -45,6 +45,7 @@ import {
 import { useTRPC } from "../../../../lib/trpc.ts";
 import { useEstateId } from "../../../../hooks/use-estate.ts";
 import { useSlackConnection } from "../../../../hooks/use-slack-connection.ts";
+import { authClient } from "../../../../lib/auth-client.ts";
 import type { Route } from "./+types/index.ts";
 
 export function meta(_args: Route.MetaArgs) {
@@ -158,6 +159,11 @@ export default function Integrations() {
     } else if (integrationId === "slack-bot") {
       // Use the shared Slack connection logic
       await connectSlackBot("/integrations");
+    } else if (integrationId === "google") {
+      const result = await authClient.integrations.link.google({
+        callbackURL: window.location.pathname,
+      });
+      window.location.href = result.url.toString();
     }
     // TODO: Implement OAuth flow for other integrations
     // This would redirect to the auth provider's OAuth URL
--- a/apps/os/app/routes/org/estate/integrations/redirect.tsx
+++ b/apps/os/app/routes/org/estate/integrations/redirect.tsx
@@ -1,7 +1,7 @@
 import { redirect } from "react-router";
 import { eq } from "drizzle-orm";
 import { getDb, schema } from "../../../../../backend/db/client.ts";
-import { MCPOAuthState } from "../../../../../backend/auth/oauth-state-schemas.ts";
+import { BaseOAuthState } from "../../../../../backend/auth/oauth-state-schemas.ts";
 import type { Route } from "./+types/redirect.ts";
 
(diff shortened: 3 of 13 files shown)


apps/os/backend/db/schema.ts
```diff
@@ -4,7 +4,7 @@ import { relations } from "drizzle-orm";
 import type { SlackEvent } from "@slack/web-api";
 import type { DynamicClientInfo } from "../auth/oauth-state-schemas.ts";
 
-export const UserRole = ["member", "admin", "owner", "guest"] as const;
+export const UserRole = ["member", "admin", "owner", "guest", "external"] as const;
```
jonastemplestein, line 16: What does this represent? `user.role` or is this the role wrt an organization? This could definitely do with a comment

apps/os/backend/db/schema.ts
```diff
@@ -247,6 +248,38 @@ export const providerEstateMappingRelations = relations(providerEstateMapping, (
   }),
 }));
 
+export const providerChannelMapping = pgTable(
```
jonastemplestein: This is too abstract. Just call the table `slack_channel` . Even with very similar providers like discord the semantics would be so different that this would be a v leaky abstraction

apps/os/backend/db/schema.ts
```diff
@@ -247,6 +248,38 @@ export const providerEstateMappingRelations = relations(providerEstateMapping, (
...
+  "provider_channel_mapping",
+  (t) => ({
+    id: iterateId("pcm"),
+    providerId: t.text().notNull(),
+    internalEstateId: t
```
jonastemplestein:
```suggestion
    estateId: t
```

There are no "non internal" estates

apps/os/backend/trpc/routers/admin.ts
```diff
@@ -333,7 +333,7 @@ export const adminRouter = router({
         continue;
       }
 
-      waitUntil(syncSlackUsersInBackground(ctx.db, slackToken, estate.id));
+      waitUntil(syncSlackForEstateInBackground(ctx.db, slackToken, estate.id));
```
jonastemplestein, line 334: How long do you expect the syncing to take? Given this is an admin procedure, my preference would be not to use waitUntil, so we can return the syncing result or an error directly

estates/iterate/iterate.config.ts
```diff
@@ -23,6 +23,33 @@ const config = defineConfig({
       // this is the #agents-with-sandboxes channel
       match: matchers.slackChannel("C09JH97Q0RL"),
     },
+    // Example: Match by Slack channel name
+    {
+      key: "speak-like-an-alien",
+      prompt: dedent`
+        Talk like an alien.
+      `,
+      match: {
```
jonastemplestein: Can we extend the factory method we have to support both names and IDs as you suggested?

estates/iterate/iterate.config.ts
```diff
@@ -23,6 +23,33 @@ const config = defineConfig({
...
+      },
+    },
+    // Example: Different behavior for Slack Connect (shared) channels
+    {
+      key: "external-users-present",
+      prompt: dedent`
+        ⚠️ External Slack Connect users may be present in this conversation.
+
+        - Avoid sharing internal company information
+        - Be professional and use formal tone
```
jonastemplestein:
```suggestion
```

estates/iterate/iterate.config.ts
```diff
@@ -23,6 +23,33 @@ const config = defineConfig({
...
+        - Don't discuss internal metrics, roadmaps, or confidential matters
+        - Treat all discussions as public-facing
```
jonastemplestein:
```suggestion
        - Treat all discussions as public-facing
        - If a true member of the slack workspace tells you it's okay to do something for an external user, then you can do it
```

apps/os/backend/agent/slack-agent.ts
```diff
@@ -142,6 +143,22 @@ export class SlackAgent extends IterateAgent<SlackAgentSlices> implements ToolsI
 
   protected getExtraDependencies(deps: AgentCoreDeps) {
     return {
+      getRuleMatchData: (state: CoreReducedState) => {
+        const slackState = state as SlackSliceState;
+
+        return {
+          agentCoreState: state,
+          durableObjectClassName: this.constructor.name,
+          slack: {
```
jonastemplestein: Why is all this here and not reduced state? we also want to show the current slack channel name to the LLM so need this anyway

apps/os/backend/agent/slack-agent.ts
```diff
@@ -142,6 +143,22 @@ export class SlackAgent extends IterateAgent<SlackAgentSlices> implements ToolsI
...
+            channelId: slackState.slackChannelId || null,
+            channelName: slackState.slackChannel?.name || null,
+            threadId: slackState.slackThreadId || null,
+            isSharedChannel: slackState.slackChannel
```
jonastemplestein:
```suggestion
            channelHasExternalUsers: slackState.slackChannel
```

apps/os/backend/agent/slack-agent.ts
```diff
@@ -142,6 +143,22 @@ export class SlackAgent extends IterateAgent<SlackAgentSlices> implements ToolsI
...
   protected getExtraDependencies(deps: AgentCoreDeps) {
     return {
+      getRuleMatchData: (state: CoreReducedState) => {
+        const slackState = state as SlackSliceState;
+
+        return {
+          agentCoreState: state,
+          durableObjectClassName: this.constructor.name,
+          slack: {
+            channelId: slackState.slackChannelId || null,
```
jonastemplestein: https://github.com/iterate-com/iterate/blob/main/apps/os/backend/agent/context.ts#L36-L41

we already have this on reduced state

apps/os/backend/trpc/routers/admin.ts
zakhardavydov: Seconds

PR #253, opened 2025-10-08
Include files in slack thread history

myst-6:
[ITE-3093: Mentioning the bot on threads with existing images doesn't work as expect](https://linear.app/iterate-com/issue/ITE-3093/mentioning-the-bot-on-threads-with-existing-images-doesnt-work-as)

### DIFF
--- a/apps/os/backend/agent/slack-agent.ts
+++ b/apps/os/backend/agent/slack-agent.ts
@@ -28,7 +28,6 @@ import { shouldIncludeEventInConversation, shouldUnfurlSlackMessage } from "./sl
 import type {
   AgentCoreEvent,
   CoreReducedState,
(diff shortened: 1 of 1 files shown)


jonastemplestein (approved):
Thanks <3 I love how you just pick up stuff from the backlog and make everything better

I think we have some other backlog ticket that might make sense to tackle at the same time 

Instead of the custom rendering of the webhook events into the single `context` prompt fragment, we could refactor all this to take all the webhooks and render them exactly as they would be rendered if they were received by a live agent

That way the images appear in the correct places and the agent gets the same context in this "late mention" case as they would in the normal case

It also prepares us for "mention-less" chiming in better

So I'd probably make that change now, too, but it can totally be a separate PR

PR #261, opened 2025-10-08
wip: Service auth

NickBlow:


apps/os/backend/worker.ts
```diff
@@ -90,9 +90,64 @@ app.onError((err, c) => {
...
   const auth = getAuth(db);
-  const session = await auth.api.getSession({
-    headers: c.req.raw.headers,
-  });
+
+  let session: AuthSession | null = null;
+
+  // Check for service auth token for impersonation
+  const serviceAuthToken = c.req.header("X-Service-Auth-Token");
+  const impersonateUserId = c.req.header("X-Impersonate-User-Id");
```
jonastemplestein, line 98: I'm not sure if this is easy with the better auth admin plugin or not

But I was wondering if we could do something like this

- introduce a column in the users table to indicate whether the user is a human user or a service account
- create a service account user with `role: admin` 
- the X-Service-Auth-Token could then be used to authenticate as that service user
- then use the normal better auth impersonation mechanism to impersonate


@BlankParticle would probably know more

PR #1276, opened 2026-04-21
chore(deps): pin Cloudflare packages to latest via pnpm catalog

jonastemplestein:
## Summary

Pin `apps/os` to older Cloudflare tooling (`@cloudflare/vite-plugin@1.15.3` → `miniflare@4.20251125.0`) while keeping `catalog:cloudflare` at the latest versions for every other app (`apps/agents`, `apps/events`, `apps/codemode`, …). `wrangler` is overridden repo-wide to `^4.84.1` so the older vite-plugin still parses the modern `wrangler.jsonc` shape.

## Why

`miniflare@4.20260421.0` ships two regressions that OOM the Vite/Node host during `apps/os` dev + Playwright spec runs within 2–5 minutes:

1. `DispatchFetchDispatcher.dispatch()` sets `options.reset = true` per runtime request — no keep-alive reuse, TCP churn.
2. Internal `undici.Pool` is constructed with `headersTimeout: 0, bodyTimeout: 0` — timeouts fully disabled, slow/stalled responses never free their buffers.

Combined with the `apps/os` local cron simulator (fires `/cdn-cgi/handler/scheduled` every few seconds) and concurrent spec workers, sockets/buffers outpace GC, the Node host heap exceeds ~1.4 GB, and V8 aborts with "JavaScript heap out of memory". Bumping `--max-old-space-size` only delays it.

`apps/agents` doesn't hit this path (single WebSocket turn, no cron loop) and **genuinely needs** the newer `@cloudflare/vite-plugin ^1.33.x` for Node.js `require()` interop used by its Workers AI + codemode flows. Hence the split.

## Changes

- `apps/os/package.json`: direct pin to `@cloudflare/vite-plugin@1.15.3`, `@cloudflare/workers-types@^4.20251128.0`, `wrangler@^4.84.1`.
- `pnpm-workspace.yaml`: `catalogs.cloudflare` unchanged at latest — only used by non-os apps now.
- `package.json`: removed `miniflare`/`workerd` from `pnpm.overrides` so each app can resolve its own transitive versions; added `wrangler: ^4.84.1` override so both vite-plugin versions agree on the CLI/config format.
- `apps/agents/package.json`: point `EVENTS_BASE_URL` default from the intermittently-522 `events-preview-1.iterate.com` to the live `events.iterate.com`.
- New doc `docs/cloudflare-deps-split-versions.md` + pointer from `AGENTS.md` explaining the split and the revert criteria (flip `apps/os` back to `catalog:cloudflare` once miniflare restores keep-alive + sane timeouts).

Verify the split with:

```bash
readlink apps/os/node_modules/@cloudflare/vite-plugin        # …vite-plugin@1.15.3…
readlink apps/agents/node_modules/@cloudflare/vite-plugin    # …vite-plugin@1.33.1…
```

## Test plan

- [x] `pnpm typecheck` — all 20 packages pass.
- [x] `pnpm -r build` for `@iterate-com/os` + `@iterate-com/agents` — both build.
- [x] `pnpm dev` starts cleanly, no `CorePaths`/config errors.
- [x] Full OS spec subset (`email-onboarding`, `env-vars-secrets`, `login-page`, `machine-provider-ready`, `naming-defaults`, `org-creation`, `org-invites`) runs ~2.4m: **29 passed, 4 pre-existing "Leave Test Org" strict-mode flake** (same behavior as `main`, unrelated to miniflare). Zero OOM in dev logs.
- [x] `cd apps/agents && pnpm test:e2e:agent-loop` against `events.iterate.com`: **1 passed in 28s** — Workers AI kimi-k2.5 loop replies "Paris".

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes dependency resolution for Cloudflare build/dev tooling across many workspaces and removes repo-wide `miniflare`/`workerd` overrides, which can affect local dev/CI installs and worker bundling behavior.
> 
> **Overview**
> **Cloudflare dependency management is reworked to use `pnpm` catalogs across the repo**, switching most apps’ Cloudflare packages (e.g. `@cloudflare/vite-plugin`, `@cloudflare/workers-types`, `wrangler`, and related Cloudflare deps) to `catalog:cloudflare` instead of per-app version ranges.
> 
> **`apps/os` is explicitly pinned to older Cloudflare tooling** (not using the catalog) while the root `package.json` drops repo-wide `miniflare`/`workerd` overrides and instead enforces a single `wrangler` version via `pnpm.overrides` for compatibility.
> 
> Updates `apps/agents` e2e scripts to default `EVENTS_BASE_URL` to `https://events.iterate.com`, and adds documentation (`docs/cloudflare-deps-split-versions.md`) with a pointer from `AGENTS.md` explaining the version split and when to revert it.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cead449a5ac0acc8a96c299d1394220cf41215ec. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jonastemplestein:
Re: Devin's concern about `apps/os` resolving `miniflare@4.20260421.0` via `@hono/vite-dev-server` — that's technically present in the lockfile but inert. Verified:

- `apps/os/node_modules/@cloudflare/vite-plugin` → `vite-plugin@1.15.3`, which has `miniflare@4.20251125.0` in its own `node_modules` (the "good" version).
- `@hono/vite-dev-server` is declared in `apps/os/package.json` but **never imported** in any `apps/os/**` source (checked with ripgrep — only appears in `package.json`). Its transitive `miniflare@4.20260421.0` is dead code at runtime.
- The Vite dev server's worker is driven exclusively by `@cloudflare/vite-plugin@1.15.3`, so the only miniflare actually loaded into the Node host process is the good one.
- Empirical confirmation: specs pass locally and in CI on this commit.

Resolving as inaccurate-in-practice. Will clean up the unused `@hono/vite-dev-server` dep in a follow-up.

PR #1282, opened 2026-04-27
Add composable Durable Object core, lifecycle, catalog, alarm, scheduler, keepalive, and inspector mixins

jonastemplestein:
## Motivation

This PR adds a small experimental Durable Object utility layer for building Agent-like classes from explicit mixins, without adopting the full Cloudflare Agents SDK as the root class.

The design goal is deliberately narrow:

- Keep raw `DurableObject` usable as the root, but wrap it once with `withDurableObjectCore()` so later mixins do not repeatedly poke at `ctx.storage` through casts.
- Compose behavior with nested `withX(...)` calls, not `pipe`, `flow`, builders, or codegen.
- Preserve the Cloudflare-style `class Room extends Base<Env> {}` ergonomics after every mixin.
- Keep public APIs small; most mutation/handler APIs are `protected` so concrete DO classes expose domain-specific public RPC methods instead.
- Make lifecycle, cataloging, alarm, scheduling, and heartbeat behavior durable enough to survive eviction, hibernation wakeups, and deployed Worker alarm delivery.

The main problem this solves is that raw Durable Objects give us powerful primitives, but they are intentionally low-level: one platform alarm slot, local SQLite/KV storage, a generic `env`, and public methods that become RPC. These mixins package the common patterns we kept discussing into reusable, documented pieces while keeping the type surface explainable.

## The Composition Shape

A consumer composes a base class, then extends it with the final Worker `Env`:

```ts
import { DurableObject } from "cloudflare:workers";
import { withD1ObjectCatalog } from "@iterate-com/shared/durable-object-utils/mixins/with-d1-object-catalog";
import { withDurableObjectCore } from "@iterate-com/shared/durable-object-utils/mixins/with-durable-object-core";
import { withLifecycleHooks } from "@iterate-com/shared/durable-object-utils/mixins/with-lifecycle-hooks";
import { withMultiplexedAlarms } from "@iterate-com/shared/durable-object-utils/mixins/with-multiplexed-alarms";
import { withScheduler } from "@iterate-com/shared/durable-object-utils/mixins/with-scheduler";

type RoomInit = {
  name: string;
  ownerUserId: string;
};

type NeedsCatalog = {
  DO_CATALOG: D1Database;
};

type Env = NeedsCatalog & {
  OTHER_BINDING: Fetcher;
};

const DurableObjectCore = withDurableObjectCore(DurableObject);

const RoomBase = withD1ObjectCatalog<RoomInit, NeedsCatalog>({
  className: "Room",
  getDatabase(env) {
    return env.DO_CATALOG;
  },
  indexes: {
    ownerUserId(params) {
      return params.ownerUserId;
    },
  },
})(
  withScheduler<RoomInit>()(
    withMultiplexedAlarms<RoomInit>()(
      withLifecycleHooks<RoomInit>()(DurableObjectCore),
    ),
  ),
);

export class Room extends RoomBase<Env> {
  async enableDailySummary() {
    await this.schedule({
      key: "daily-summary",
      method: "sendDailySummary",
      payload: { room: this.initParams.name },
      recurrence: {
        type: "cron",
        expression: "0 9 * * *",
      },
    });
  }

  protected async sendDailySummary(payload: unknown) {
    console.log("daily summary", payload);
  }
}
```

The important type invariant is that every mixin republishes a fresh generic class value. This remains legal:

```ts
const Base = withLifecycleHooks<RoomInit>()(
  withDurableObjectCore(DurableObject),
);

class Room extends Base<Env> {}
```

That is the same broad ergonomic target as Cloudflare's `withVoice(Agent)` style. The implementation keeps the type expressions documented in code and in `packages/shared/src/durable-object-utils/README.md` so the casts are not magic.

### `withDurableObjectCore()`

Adds the small shared capability layer every other storage-owning mixin expects:

```ts
const DurableObjectCore = withDurableObjectCore(DurableObject);
```

It exposes protected helpers for local DO SQLite/KV access and a synchronous transaction helper:

```ts
this.useDurableObjectSql((sql) => sql.exec("SELECT 1"));
this.useDurableObjectKv((kv) => [...kv.list()]);
this.transactionSync(() => {
  // multiple local SQLite/KV operations that should be atomic together
});
```

The callback helpers are for one-off storage operations, such as the debug inspectors. Raw protected storage handles still exist for state-owning mixins like alarms and scheduler that need several related operations across helper methods. This keeps the repeated `ctx.storage` casts in one root mixin rather than scattering them through every feature mixin.

### `withLifecycleHooks()`

Adds named initialization, a protected `initParams` getter, and two hook registration points:

- `registerOnFirstInitialize(fn)` runs after params are created for the first time.
- `registerOnStart(fn)` runs once per Durable Object activation after params exist.

Default stub acquisition is through the free helper:

```ts
const room = await getOrInitializeDoStub({
  namespace: env.ROOMS,
  name: "room-a",
  initParams: {
    ownerUserId: "user-a",
  },
});
```

`getOrInitializeDoStub()` always calls `initialize()`. If the init shape is only `{ name: string }`, `initParams` is optional and defaults to `{ name }`. If the init shape has more fields, TypeScript requires `initParams`.

Inside subclasses, the protected getter is the intended path:

```ts
class Room extends RoomBase<Env> {
  sendMessage(text: string) {
    const { name, ownerUserId } = this.initParams;

    return {
      room: name,
      ownerUserId,
      text,
    };
  }
}
```

`this.initParams` throws `NotInitializedError` synchronously if the object has not been initialized. External callers cannot access it because it is protected; their public API is `initialize(params)`, `ensureStarted()`, and `assertInitialized()`.

Lifecycle hooks are awaited by `initialize()` and `ensureStarted()`. If a hook wants best-effort work, it must explicitly detach and catch that promise so hook failure does not make initialization fail. `withD1ObjectCatalog()` uses that pattern for its external D1 writes.

The PR includes regression coverage for important lifecycle edge cases, including retry after hook failure and JavaScript's odd but real `throw undefined` case.

### `withD1ObjectCatalog()`

Adds a best-effort D1 catalog of initialized objects. This replaces the vague `withExternalListing` naming with a purpose-specific API: it is a D1-backed object catalog/index, not source-of-truth state.

The mixin is implemented in terms of lifecycle hooks: it registers `onFirstInitialize` and `onStart` hooks and detaches the D1 writes as best-effort promises. That keeps `withLifecycleHooks()` as the primitive and keeps cataloging as one flexible consumer of lifecycle events.

The object table is keyed by `(class, name)` and stores:

```txt
class, name, id, init_params_json, created_at, last_started_at
```

The optional index table stores derived init-param indexes without dynamic D1 columns:

```txt
class, index_name, index_value, name
```

Example:

```ts
const CatalogedRoomBase = withD1ObjectCatalog<RoomInit, NeedsCatalog>({
  className: "Room",
  getDatabase(env) {
    return env.DO_CATALOG;
  },
  indexes: {
    ownerUserId: (params) => params.ownerUserId,
    projectId: (params) => params.projectId,
    memberUserIds: (params) => params.memberUserIds,
  },
})(withLifecycleHooks<RoomInit>()(withDurableObjectCore(DurableObject)));
```

Free helpers support lookup without waking a specific Durable Object:

```ts
await getD1ObjectCatalogRecord<RoomInit>(env.DO_CATALOG, {
  className: "Room",
  name: "room-a",
});

await listD1ObjectCatalogRecordsByIndex<RoomInit>(env.DO_CATALOG, {
  className: "Room",
  indexName: "ownerUserId",
  indexValue: "user-a",
});
```

The D1 write is best-effort and idempotent. The mixin sends `CREATE TABLE IF NOT EXISTS` for both tables in the same D1 batch as each upsert, so construction does not block on D1 setup. `getD1ObjectCatalogRecord()` returns `null` when the local DO is uninitialized, the background D1 write has not completed, or the tables do not exist yet.

### `withMultiplexedAlarms()`

Cloudflare gives each Durable Object one platform alarm slot. This mixin turns that single slot into a queue of logical one-shot alarms stored in the object's local SQLite database.

```ts
const RoomBase = withMultiplexedAlarms<RoomInit>()(
  withLifecycleHooks<RoomInit>()(withDurableObjectCore(DurableObject)),
);

class Room extends RoomBase<Env> {
  async scheduleSummary() {
    await this.scheduleMultiplexedAlarm({
      key: "daily-summary",
      runAt: Date.now() + 60_000,
      method: "sendDailySummary",
      payload: { room: this.initParams.name },
    });
  }

  protected async sendDailySummary(payload: unknown) {
    console.log("running daily summary", payload);
  }
}
```

Important runtime semantics:

- `key` is an idempotency key. Scheduling the same key replaces the existing logical alarm.
- Target methods are `protected` instance methods, not a public registry.
- Payloads must be JSON-serializable because rows must survive eviction, deploys, and alarm wakeups.
- Rows are deleted only after the target method succeeds.
- If a target throws, `alarm()` throws and the row remains, preserving Cloudflare's normal alarm retry semantics.
- If a subclass or later mixin overrides `alarm()`, it must call `super.alarm?.()`.

### `withScheduler()`

Adds key-based schedules on top of multiplexed alarms.

```ts
const RoomBase = withScheduler<RoomInit>()(
  withMultiplexedAlarms<RoomInit>()(
    withLifecycleHooks<RoomInit>()(withDurableObjectCore(DurableObject)),
  ),
);

class Room extends RoomBase<Env> {
  async enablePolling() {
    await this.schedule({
      key: "poll-api",
      method: "pollAPI",
      recurrence: {
        type: "interval",
        everyMs: 30_000,
      },
    });
  }

  protected async pollAPI(payload: unknown, schedule: SchedulerRecord) {
    console.log("running schedule", schedule.key, payload);
  }
}
```

The recurrence API is intentionally a tagged union rather than Cloudflare Agents-style positional overloads:

```ts
type SchedulerRecurrence =
  | { type: "once"; runAt: Date | number }
  | { type: "delayed"; delayMs: number }
  | { type: "interval"; everyMs: number }
  | { type: "cron"; expression: string; timezone?: string }
  | { type: "rrule"; rrule: string; timezone?: string; dtstart?: Date | number };
```

Reasons for this shape:

- It avoids unit guessing. `everyMs` and `delayMs` are explicit milliseconds.
- Stored rows preserve the original recurrence tag, which makes future bugfixes and migrations easier.
- Cron and RRULE are separate concepts instead of one overloaded string.
- `key` makes schedules idempotent by default, so scheduling from `registerOnStart()` does not create duplicates.

Cron uses `croner`; RRULE uses `rrule`. If `timezone` is omitted, schedules evaluate in UTC. RRULE input is deliberately narrow: pass a bare rule body such as `FREQ=MONTHLY;BYDAY=MO;BYSETPOS=1`; use explicit `dtstart` and `timezone` fields instead of embedded iCalendar snippets.

Failure policy is split intentionally:

- One-shot and delayed schedules keep their row due when the callback throws, so platform alarm retry can retry the work.
- Recurring schedules log the failure and advance, so one broken recurring run does not wedge the object forever.
- Interval schedules have overlap/hung-run handling so they do not pile up concurrent executions.

The latest scheduler fix handles callbacks that replace their own schedule key while running. Recurring advancement and finite-RRULE cleanup now snapshot-match the row that started the callback; if the callback wrote a replacement row, the stale advance/delete touches zero rows and the replacement's own multiplexed alarm remains authoritative.

### `withKeepAlive()`

Adds a tiny heartbeat primitive on top of `withScheduler()`.

This is deliberately not a second scheduler. It only names a common pattern: while some runtime condition should be rechecked, keep a keyed recurring callback scheduled so the object is woken periodically. That is useful for cases like a Discord connection DO where the durable state says "this connection should be connected", but the in-memory socket/client must be rebuilt after restart from `env` and DO storage.

The README explains the tradeoff explicitly: keepalive can wake and reconcile; it cannot make an in-memory promise, socket, or API client durable. Long-running work still needs app-owned durable state if it must survive process loss.

### Debug Inspectors

Adds two explicit unsafe debug mixins:

```ts
const InspectorBase = withKvInspector({
  unsafe: "I_UNDERSTAND_THIS_EXPOSES_KV",
})(
  withOuterbase({
    unsafe: "I_UNDERSTAND_THIS_EXPOSES_SQL",
  })(withDurableObjectCore(DurableObject)),
);
```

Routes:

- `GET /__kv` renders a simple KV inspector page.
- `GET /__kv/json` returns local Durable Object KV entries for tools/tests.
- `GET /__outerbase` renders an embedded libSQL Studio page.
- `POST /__outerbase/sql` executes SQL against `ctx.storage.sql`.

These are for local/test workers or explicitly gated development routes. The `unsafe` option is an acknowledgement at composition time, not runtime authentication.

## Why Not More Abstraction?

This PR intentionally does not add a `pipe()`, `flow()`, builder, root class, decorator DSL, or giant environment registry.

Nested composition gives better TypeScript diagnostics and is easier to explain. Capability dependencies stay on the mixin that needs them:

```ts
withScheduler(
  withMultiplexedAlarms(
    withLifecycleHooks(withDurableObjectCore(DurableObject)),
  ),
);
```

Env requirements stay as structural lower bounds on the returned generic class. For example, `withD1ObjectCatalog()` requires enough `Env` to retrieve D1, but the final application `Env` can still have additional bindings.

## Production and Runtime Notes

- Init params are persistent identity/config, not dependency injection. They must be serializable enough to cross DO RPC and persist in storage. Rebuild clients from `env` on each activation.
- `withMultiplexedAlarms()` and `withScheduler()` require SQLite-backed Durable Objects because they use `ctx.storage.sql`.
- Alarm and scheduler payloads must be JSON-serializable. The code validates this at schedule time and throws mixin-specific errors.
- D1 object cataloging is a best-effort discoverability index. Local DO storage remains the source of truth.
- Start hooks are retryable and at-least-once if a later hook fails. External side effects in hooks must be idempotent.
- The inspector routes are unsafe by design and should only be mounted behind local/dev/test routing.

## Testing and Tooling

Adds a dedicated test harness under `packages/shared/src/durable-object-utils`:

- Type tests using Vitest `expectTypeOf`, including examples that prove protected members and generic `Base<Env>` composition still work.
- Worker-pool unit tests using `cloudflare:test`, real DO stubs, RPC methods, SQLite-backed storage, D1 binding, fetch wrappers, lifecycle hooks, alarms, scheduler behavior, keepalive behavior, and D1 catalog indexing.
- Deployed Worker E2E tests using an Alchemy/Wrangler deployment path, including natural platform alarm delivery for multiplexed alarms and scheduler.
- A shared fronting Worker test entrypoint so unit and deployed E2E tests exercise the same composition shape.

Also adds `.agents/skills/adding-a-new-durable-object-mixin/SKILL.md`, so future mixins follow the same process: spec the API first, keep types documented, add type/unit/E2E coverage where relevant, and avoid new composition abstractions unless explicitly designed.

`packages/shared/src/durable-object-utils/AGENTS.md` remains a symlink to `README.md`, so humans and agents read the same durable-object-utils guidance.

## Verification

Latest local verification after merging `origin/main`:

```bash
pnpm --dir packages/shared test:durable-object-utils:type
pnpm --dir packages/shared test:durable-object-utils:unit -- --runInBand
pnpm exec oxlint packages/shared/src/durable-object-utils --threads 1 --deny-warnings
pnpm exec oxfmt --check packages/shared/src/durable-object-utils .agents/skills/adding-a-new-durable-object-mixin/SKILL.md README.md packages/shared/package.json
doppler run -- pnpm --dir packages/shared test:durable-object-utils:e2e:deploy
```

Result: deployed E2E passed with 9 tests, including natural platform alarm delivery.

## Review Focus

The most important things to review are:

- Whether the lifecycle semantics are clear enough: first-initialize once per object, start once per activation, and `ensureStarted()` as the readiness boundary.
- Whether the D1 catalog/index shape is flexible enough without becoming a dynamic-schema system.
- Whether scheduler and keepalive failure policy match what we want before wider adoption.
- Whether the protected/public surface is strict enough for DO RPC exposure.
- Whether the README comments and inline comments make the type incantations explainable rather than magical.
- Whether the unsafe inspector opt-ins are loud enough for future production use.

---

> [!NOTE]
> **Medium Risk**
> Adds new Durable Object mixins that manage initialization, D1 writes, alarms, and scheduling, plus new worker-pool and deployed E2E test/deploy tooling; runtime behavior changes are non-trivial and could impact future DO consumers if adopted.
> 
> **Overview**
> Adds a new experimental `packages/shared/src/durable-object-utils` library that composes Durable Objects via mixins: `withDurableObjectCore` (protected wrappers around `ctx.storage`/alarms), `withLifecycleHooks` (named init + start hooks + `getOrInitializeDoStub`), `withD1ObjectCatalog` (best-effort D1 mirroring + secondary indexes), `withMultiplexedAlarms` (logical alarms stored in DO SQLite behind the single platform alarm), `withScheduler` (cron/interval/RRULE schedules projected onto multiplexed alarms), and fetch-wrapping debug inspectors `withKvInspector`/`withOuterbase`.
> 
> Extends `@iterate-com/shared` package exports and scripts to ship these mixins and run new type/unit suites, adds dependencies needed for scheduling (`croner`, `rrule`) and comparisons (`dequal`), and introduces a shared fronting-worker test harness plus an Alchemy/Wrangler-backed deployed E2E workflow (deploy/run/destroy) to validate real Cloudflare alarm delivery and D1 behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e9742f972e71eaf2ed0825d1c33a82f5cb2a39f3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/shared/src/durable-object-utils/mixins/with-kv-inspector.ts
```diff
@@ -0,0 +1,94 @@
...
+export type WithKvInspectorResult<TBase extends Constructor> = TBase & Constructor<FetchBase>;
+
+/**
+ * Debug-only KV inspector.
+ *
+ * This exposes every key/value pair in the Durable Object's SQLite-backed KV
+ * storage. The `unsafe` option is intentionally noisy so call sites have to
+ * acknowledge that this mixin must sit behind a development-only or otherwise
+ * authenticated route.
+ */
```
jonastemplestein, line 26: This is a terrible comment - it doesn't even explain what it does?!  I.e. render a KV debug page for a certain path etc

packages/shared/src/durable-object-utils/mixins/with-outerbase.ts
```diff
@@ -0,0 +1,175 @@
...
+export type WithOuterbaseResult<TBase extends Constructor> = TBase & Constructor<FetchBase>;
+
+/**
+ * Debug-only SQL inspector.
+ *
+ * This exposes arbitrary SQL execution against the Durable Object's embedded
+ * SQLite storage. The `unsafe` option is intentionally noisy so call sites have
+ * to acknowledge that this mixin must sit behind a development-only or
+ * otherwise authenticated route.
+ */
```
jonastemplestein, line 29: Terrible comment - doesn't expose what this does

packages/shared/src/durable-object-utils/test-harness/initialize-fronting-worker.ts
```diff
@@ -0,0 +1,249 @@
+/// <reference types="@cloudflare/workers-types" />
```
jonastemplestein, line 1: This file doesn't explain what it's there for or how it's used at all?!

packages/shared/src/durable-object-utils/mixins/with-lifecycle-hooks.type.test.ts
```diff
@@ -0,0 +1,142 @@
+import { DurableObject } from "cloudflare:workers";
```
jonastemplestein, line 1: should be a proper vitest test with expect-types (research it)

packages/shared/src/durable-object-utils/mixins/with-kv-inspector.ts
jonastemplestein: Addressed. The JSDoc now says this is a debug-only KV inspector for the Durable Object's embedded KV storage and explicitly documents the owned routes: `GET /__kv` renders the HTML page and `GET /__kv/json` returns the entries for tools/tests.

packages/shared/src/durable-object-utils/test-harness/initialize-fronting-worker.ts
jonastemplestein: Addressed. The file now starts with a module-level comment explaining that this is the shared test-only fronting Worker used by both worker-pool unit tests and deployed E2E tests, so both paths exercise the same DO stubs, RPC, fetch wrappers, D1 binding, and SQLite-backed storage.

packages/shared/src/durable-object-utils/mixins/with-outerbase.ts
jonastemplestein: Addressed. The JSDoc now says this is a debug-only SQL inspector for the Durable Object's embedded SQLite storage and explicitly documents the owned routes: `GET /__outerbase` renders the embedded libSQL Studio page and `POST /__outerbase/sql` executes SQL against `ctx.storage.sql`.

packages/shared/src/durable-object-utils/mixins/with-lifecycle-hooks.type.test.ts
jonastemplestein: Addressed. This is now `initialize.type.test.ts`, runs through Vitest with `describe`/`it` and `expectTypeOf`, and `vitest.type.config.ts` explicitly includes `*.type.test.ts` for `--typecheck.only`. Verified with `pnpm --dir packages/shared test:durable-object-utils`.

packages/shared/src/durable-object-utils/mixins/kv-inspector.ts
```diff
@@ -0,0 +1,151 @@
...
+  ctx: DurableObjectState;
+};
+
+type FetchBase = {
+  fetch(request: Request): Response | Promise<Response>;
+};
+
+type OptionalFetchBase = {
+  fetch?(request: Request): Response | Promise<Response>;
+};
```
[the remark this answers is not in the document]
jonastemplestein: Addressed in `f1dc01f89`. I extracted the duplicated generic DO constructor types, fetch result type, base-fetch delegation, and protected `ctx` access into `mixins/fetch-mixin-utils.ts`, with comments explaining the constructor/static split and the protected-field escape hatch. `kv-inspector.ts` and `outerbase.ts` now share that helper.

packages/shared/src/durable-object-utils/mixins/external-listing.ts
```diff
@@ -0,0 +1,276 @@
+/// <reference types="@cloudflare/workers-types" />
+
+import { DurableObject } from "cloudflare:workers";
+import type { InitializeMembers, NamedInit } from "./initialize.ts";
+
+type Constructor<T = object> = abstract new (...args: any[]) => T;
```
jonastemplestein: Why is that defined in external-listing.ts?

packages/shared/src/durable-object-utils/mixins/initialize.ts
```diff
@@ -0,0 +1,317 @@
...
+// A plain constructor type for "this mixin adds these instance members".
+//
+// Example:
+//
+//   Constructor<InitializeMembers<RoomInit>>
+//
+// means "instances constructed by this class have initialize/assertInitialized".
+// We intersect that with TBase below so the returned class keeps the original
+// static side and generic DurableObject constructor shape.
+type Constructor<T = object> = abstract new (...args: any[]) => T;
```
jonastemplestein: why is this defined here again?

packages/shared/src/durable-object-utils/mixins/with-initialize.ts
```diff
@@ -0,0 +1,317 @@
...
+// Example:
+//
+//   Constructor<InitializeMembers<RoomInit>>
+//
+// means "instances constructed by this class have initialize/assertInitialized".
+// We intersect that with TBase below so the returned class keeps the original
+// static side and generic DurableObject constructor shape.
+type Constructor<T = object> = abstract new (...args: any[]) => T;
+
+export type NamedInit = {
```
jonastemplestein: should be BaseInitParams or something

packages/shared/src/durable-object-utils/mixins/with-external-listing.ts
jonastemplestein, line 1: file names should be with-external-listing.ts and with-initialize-params.ts etc

packages/shared/src/durable-object-utils/mixins/with-kv-inspector.ts
jonastemplestein, line 1: with-kv-inspector.ts

packages/shared/src/durable-object-utils/mixins/with-outerbase.ts
jonastemplestein, line 1: with-outerbase.ts

packages/shared/src/durable-object-utils/mixins/initialize.ts
```diff
@@ -0,0 +1,317 @@
...
+  // call the public `initialize()` RPC with the name-filled params.
+  await (stub as unknown as InitializeMembers<InitParamsOf<TInstance>>).initialize({
+    ...input,
+    name,
+  } as unknown as InitParamsOf<TInstance>);
+
+  return stub;
+}
+
+function areInitializeParamsEqual<InitParams extends NamedInit>(
```
jonastemplestein: this seems unnecessarily fancy

packages/shared/src/durable-object-utils/mixins/with-initialize.ts
```diff
@@ -0,0 +1,317 @@
...
+      protected get initParams(): InitParams {
+        return this.assertInitialized();
+      }
+
+      async initialize(params: InitParams): Promise<InitParams> {
+        if (!params.name) {
+          throw new Error("initialize(params) requires a non-empty params.name.");
+        }
+
+        const objectName = this.ctx.id.name;
```
jonastemplestein: FYI this.ctx.id.name doesn't exist in miniflare - i hope that works anyway

packages/shared/src/durable-object-utils/mixins/initialize.ts
jonastemplestein: as in - we can inline it and use a deepEqual that i'm sure we have in a popular and simple vendor package in our project

packages/shared/src/durable-object-utils/mixins/external-listing.ts
jonastemplestein: Addressed in `7530c0306`. The shared `Constructor` / `DurableObjectConstructor` types now live in `mixins/mixin-types.ts` instead of being defined inside `with-external-listing.ts`.

packages/shared/src/durable-object-utils/mixins/initialize.ts
jonastemplestein: Addressed in `7530c0306`. The shared constructor types now live in `mixins/mixin-types.ts` and `with-initialize.ts` imports them.

packages/shared/src/durable-object-utils/mixins/with-initialize.ts
jonastemplestein: Addressed. `InitializeInput` now uses `BaseInitParams` for the generic parameter and the explanatory comment names that same type parameter.

packages/shared/src/durable-object-utils/mixins/external-listing.ts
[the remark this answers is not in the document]
jonastemplestein: Addressed in `7530c0306`. Added `mixins/mixin-types.ts` for the shared plain constructor and Durable Object constructor constraints, then imported those from `with-initialize.ts` and `with-external-listing.ts`.

packages/shared/src/durable-object-utils/test-harness/initialize-fronting-worker.ts
```diff
@@ -0,0 +1,256 @@
...
+    name: error instanceof Error ? error.name : "UnknownError",
+    message: error instanceof Error ? error.message : String(error),
+  };
+}
+
+function isCaughtErrorResult(
+  value: SendMessageResult | CaughtErrorResult,
+): value is CaughtErrorResult {
+  return "name" in value && "message" in value;
+}
```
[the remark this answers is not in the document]
jonastemplestein: Addressed in `7530c0306`. `CaughtErrorResult` now has an explicit `kind: "error"` discriminant and `isCaughtErrorResult()` checks that instead of relying on `name` / `message` property existence.

packages/shared/src/durable-object-utils/mixins/with-external-listing.ts
jonastemplestein: Addressed in `7530c0306`. Renamed the mixin implementation files to `with-*`: `with-external-listing.ts`, `with-initialize.ts`, `with-kv-inspector.ts`, and `with-outerbase.ts`. I used `with-initialize.ts` to match the exported `withInitialize()` function name rather than changing the public API to `withInitializeParams()`.

packages/shared/src/durable-object-utils/mixins/with-kv-inspector.ts
jonastemplestein: Addressed in `7530c0306`: renamed to `with-kv-inspector.ts` and updated package exports/imports/docs.

packages/shared/src/durable-object-utils/mixins/with-outerbase.ts
jonastemplestein: Addressed in `7530c0306`: renamed to `with-outerbase.ts` and updated package exports/imports/docs.

packages/shared/src/durable-object-utils/mixins/with-initialize.ts
jonastemplestein: Addressed with a code comment and test coverage. `withInitialize` now explicitly documents that `ctx.id.name` is the production invariant for `getByName` / `idFromName`, and the worker-pool Miniflare tests exercise both successful initialization and the mismatched-name branch. If Miniflare stopped exposing it, those tests would fail.

packages/shared/src/durable-object-utils/mixins/initialize.ts
jonastemplestein: Addressed in `7530c0306`. Removed the bespoke recursive `deepEqual` implementation and inlined the check as `dequal(existing, params)` using the small `dequal` package, now declared directly in `packages/shared`.

packages/shared/src/durable-object-utils/mixins/with-scheduler.ts
```diff
@@ -0,0 +1,751 @@
...
+            await this.advanceRecurringSchedule(row, recurrence);
+            return;
+          }
+
+          throw error;
+        }
+
+        if (isRecurring) {
+          await this.advanceRecurringSchedule(row, recurrence);
+          return;
```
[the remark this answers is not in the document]
jonastemplestein: Addressed in `d22437bef`. I added a regression that seeds the persisted shape of an exhausted finite RRULE schedule, proves the prior behavior throws `NoNextScheduleOccurrenceError`, and verifies the fixed path runs the final callback, deletes the schedule row, and lets the multiplexed alarm owner delete the logical alarm. The scheduler now treats `NoNextScheduleOccurrenceError` during recurrence advance as terminal success rather than an alarm failure, with an explanatory comment in `advanceRecurringSchedule()`.

packages/shared/src/durable-object-utils/mixins/with-scheduler.ts
```diff
@@ -0,0 +1,776 @@
...
+      next = new Cron(recurrence.expression, {
+        paused: true,
+        timezone: recurrence.timezone ?? "UTC",
+      }).nextRun(after);
+      break;
+    case "rrule":
+      next = rrulestr(recurrence.rrule, {
+        dtstart: new Date(recurrence.dtstartMs),
+        tzid: recurrence.timezone ?? "UTC",
+      }).after(after, false);
```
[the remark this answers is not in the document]
jonastemplestein: Addressed in `f6c8494cd`. I added a unit regression for `FREQ=DAILY;COUNT=1` with the default `dtstart`; it failed before the fix with `NoNextScheduleOccurrenceError`. Initial RRULE scheduling now asks for the occurrence at-or-after `dtstart`, because `dtstart` is itself a valid first occurrence. Recurrence advancement remains exclusive, so completed finite RRULEs still terminate and clean up instead of repeating the final occurrence.

packages/shared/src/durable-object-utils/mixins/with-scheduler.ts
```diff
@@ -0,0 +1,776 @@
...
+
+    if (json === undefined) {
+      throw new Error("JSON.stringify returned undefined.");
+    }
+
+    return json;
+  } catch (error) {
+    throw new SchedulerPayloadSerializationError(error);
+  }
+}
```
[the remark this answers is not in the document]
jonastemplestein: Addressed in `f6c8494cd`. Extracted the duplicated persisted-payload serialization into `mixins/json-payload.ts`. Both mixins now share the same JSON/null/circular-object handling while still passing their own public error class (`SchedulerPayloadSerializationError` or `MultiplexedAlarmPayloadSerializationError`), so callers keep the precise error surface.

packages/shared/src/durable-object-utils/mixins/with-lifecycle-hooks.ts
```diff
@@ -0,0 +1,417 @@
...
+          // Do not throw from inside blockConcurrencyWhile. Cloudflare documents
+          // that thrown errors reset the Durable Object, which is correct for
+          // fatal constructor setup but too harsh for retryable lifecycle hooks.
+          //
+          // Capture and re-throw after the gate so initialize()/ensureStarted()
+          // callers can catch and retry normally.
+          // https://developers.cloudflare.com/durable-objects/api/state/#blockconcurrencywhile
+          if (startupError !== undefined) {
+            throw startupError;
+          }
```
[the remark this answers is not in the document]
jonastemplestein: Addressed in `d88d36db6`. I added a regression where a registered start hook deliberately does `throw undefined`; before the fix, `initialize()` incorrectly resolved with params and later `ensureStarted()` could observe a resolved startup promise even though hooks never completed. The lifecycle startup path now tracks `hasStartupError` separately from the captured `startupError`, with comments explaining why JavaScript arbitrary thrown values make `undefined` unsafe as the sentinel. Verified with `pnpm --dir packages/shared test:durable-object-utils`, `pnpm --dir packages/shared typecheck`, `pnpm exec oxlint packages/shared/src/durable-object-utils --threads 1 --deny-warnings`, and `pnpm exec oxfmt --check packages/shared/src/durable-object-utils`.

packages/shared/src/durable-object-utils/mixins/with-scheduler.ts
```diff
@@ -0,0 +1,774 @@
...
+        const nowMs = Date.now();
+
+        this.ctx.storage.sql.exec(
+          `UPDATE ${SCHEDULER_TABLE}
+           SET next_run_at_ms = ?, running = 0, execution_started_at_ms = NULL, updated_at_ms = ?
+           WHERE key = ?`,
+          nextRunAtMs,
+          nowMs,
+          row.key,
+        );
```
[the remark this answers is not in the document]
jonastemplestein: Addressed in `71c13b118`. I added failing regressions for both self-replacement cases:

- a recurring interval callback that replaces its own key with a delayed schedule must keep the replacement's near-term `nextRunAt`, rather than advancing using the stale interval row;
- a final finite RRULE callback that replaces its own key must not have the replacement deleted by terminal cleanup.

The fix snapshot-matches the pre-callback row for both recurring advancement and finite-RRULE deletion. If the callback replaced the row, the stale update/delete writes zero rows and returns because the replacement schedule already armed its own multiplexed alarm.

packages/shared/src/durable-object-utils/e2e/initialize.e2e.test.ts
```diff
@@ -0,0 +1,356 @@
...
+    const response = await fetch(input, init);
+    if (![404, 500, 503].includes(response.status)) {
+      return response;
+    }
+
+    await new Promise((resolve) => setTimeout(resolve, 250));
+  }
+
+  return await fetch(input, init);
+}
```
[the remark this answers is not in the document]
jonastemplestein: Addressed in `4e259d2fa`. I changed the deployed E2E route retry helper to retry only routing/propagation statuses (`404` and `503`). A `500` now returns immediately because it means the Worker code ran and produced an application error, which is exactly what this NotInitializedError test is asserting. I also added a comment beside the helper explaining the distinction. Verified with `pnpm exec oxfmt --check packages/shared/src/durable-object-utils/e2e/initialize.e2e.test.ts` and `doppler run -- pnpm --dir packages/shared test:durable-object-utils:e2e:deploy`; deployed E2E passed with 9 tests and the run time dropped from ~21s to ~9.5s.

doppler.yaml
```diff
@@ -6,6 +6,8 @@
 # `doppler run` resolves `project` from the current working directory against these `path` entries
 # (see `doppler setup`). Scripts should not hard-code `--project` unless there is a special reason.
 setup:
+  - project: _shared
+    path: packages/shared/
```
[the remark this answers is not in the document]
jonastemplestein: This is a false positive. The new `_shared` mapping is indented at the same sequence level as the existing entries:

```yaml
setup:
  - project: _shared
    path: packages/shared/
  - project: os
    path: ./
```

I verified it locally by parsing `doppler.yaml` with Ruby's YAML parser and by running `doppler setup --config dev_test --project os --no-interactive`; both loaded the setup list correctly, including `_shared`, `os`, and the subsequent app mappings.

PR #1295, opened 2026-04-29
Revert "Disable automatic specs workflow runs"

mmkal:
Reverts iterate/iterate#1292

will be broken until our login spec bug is fixed

## Environment Config Lease


mmkal:
already reverted

PR #1331, opened 2026-05-14
[codex] Use worker bundler for project config workers

jonastemplestein:
## Summary

- Bundle project `iterate-config` workers with `@cloudflare/worker-bundler` when the repo contains both `package.json` and `worker.ts`.
- Cache the built Worker Loader code object in the Project Durable Object checkout storage, while keeping the raw `worker.ts` path for repos without `package.json`.
- Return a clear `This worker is currently being built.` response while the first build is running and no cached worker is available.
- Seed new iterate-config repos with a minimal `package.json` and extend the project ingress test harness to exercise the package-based path.

## Validation

- `pnpm --dir apps/os2 typecheck`
- `OS2_PROJECT_INGRESS_TEST_APP_ROOT=$PWD/apps/os2 pnpm --dir packages/shared exec vitest run --config ../../apps/os2/src/durable-objects/project-ingress.vitest.config.ts`

## Notes

The test uses an override seam for the bundler call because `@cloudflare/worker-bundler` dynamically imports `esbuild.wasm`, which the Workers Vitest Durable Object harness does not reliably resolve. The fixture writes both `package.json` and `worker.ts`; the raw worker would return a different response, and the test asserts the bundled response.

## Environment Config Lease

### OS
Status: released
Commit: `61f8d6f`
Preview: https://os2.iterate-preview-3.com
[Workflow run](https://github.com/iterate/iterate/actions/runs/25860667541)
Updated: 2026-05-14T13:01:39.750Z

---

> [!NOTE]
> **High Risk**
> Touches critical project provisioning and ingress dispatch paths, adds asynchronous worker builds with new failure modes (503/building, fallback routing), and introduces Cloudflare DNS mutations plus a DB migration dropping `projects.metadata`. These changes can impact project creation, routing, and deployment stability if bundling/DNS/config assumptions are wrong.
> 
> **Overview**
> **Project provisioning and routing are revamped**: `projects.create` now writes to D1 immediately and *asynchronously bootstraps* the Project Durable Object (via `waitUntil`) so API creation returns faster, while the Project DO defers heavy setup (config worker build + agents root rule) and returns `503` with `retry-after` when a request arrives before the first build has produced a cached worker.
> 
> **Project config workers can now be bundled**: when an `iterate-config` repo includes `package.json`, the Project DO reads all repo files, bundles with `@cloudflare/worker-bundler`, and stores the resulting Worker Loader `code` object (modules/mainModule/compat flags) in DO storage; repos without `package.json` keep the simple single-file `worker.ts` path. The base seed repo and test harness were updated to include `package.json` plus `apps/app1`/`apps/app2` workers, and ingress now forwards an `x-iterate-app-slug` header derived from platform/custom-host subdomains.
> 
> **DNS automation and schema cleanup**: adds optional `cloudflare.apiToken` config and provisions `*.<project-slug>.<base>` CNAMEs in Cloudflare (copying the existing `*.<base>` target) with lifecycle events for created/failed, and drops `projects.metadata` across the contract, router inputs, UI, D1 schema, and migrations (including updated generated SQL queries/types).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 61f8d6fde6a776ec9ab10028fb0d84cefba038eb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


apps/os2/iterate-config-repo/apps/app1/worker.ts
```diff
@@ -0,0 +1,13 @@
+// @ts-nocheck
+import { WorkerEntrypoint } from "cloudflare:workers";
+
+export class AppOne extends WorkerEntrypoint {
+  async fetch() {
+    return new Response("hello from app one", {
```
jonastemplestein: Remove the dumb headers - also in ../worker.ts

apps/os2/iterate-config-repo/worker.ts
```diff
@@ -17,3 +30,10 @@ export default class Project extends WorkerEntrypoint {
     console.log("Project config worker afterAppend", event.type);
   }
 }
+
+function appSlugFromHostname(hostname) {
```
jonastemplestein: Make more generic - appSlugFromHostname

jonastemplestein: Though I think maybe better to model this as "each app has a partial fetch function" and then you just iterate over all the apps and say "does your fetch return anything? if so, return that response. otherwise go to next"

Then you can put the helper function in a file lib/sdk.ts in the iterate-config-repo

apps/os2/iterate-config-repo/worker.ts
```diff
@@ -1,10 +1,23 @@
 // @ts-nocheck
 import { WorkerEntrypoint } from "cloudflare:workers";
 
+export { AppOne } from "./apps/app1/worker.ts";
+export { AppTwo } from "./apps/app2/worker.ts";
```
jonastemplestein: Don't need to export

I though you could just map from ["app1", "app2"] to import statements to fetch methods 

And then in the fetch function we just go through the fetchers until we find one that returns something

An app is just a partial fetch function

apps/os2/iterate-config-repo/worker.ts
```diff
@@ -1,10 +1,23 @@
...
+
 export default class Project extends WorkerEntrypoint {
   async fetch(request) {
     const url = new URL(request.url);
     const hostname = request.headers.get("x-iterate-ingress-hostname") ?? url.hostname;
```
jonastemplestein: I don't want any of this ugly host malarkey here

jonastemplestein: worker.ts should be mega clean and easy to reason about

PR #1349, opened 2026-05-19
Rebuild voice agent processor on stream host

jonastemplestein:
## What changed

Stream-based realtime voice agent on the stream-processor host, now as a **single unified `voice-agent` processor**:

- One processor contract + implementation for Gemini Live, OpenAI Realtime, and Grok Realtime; the backend is selected from `setup-configured` state. Provider differences live in one endpoint table (URL, headers, session setup message, message handler).
- Six generic `provider-*` audit/status events with a `provider` payload field replace the previous 27 per-provider event types. `consumes` is trimmed to the four events the processor acts on, so its own audit appends no longer round-trip through ingest.
- Provider message audit events redact large strings (base64 PCM), so audio is stored exactly once per direction. Per-connection promise chains serialize message handling and appends, guaranteeing output frames land in stream order.
- Gemini `goAway` surfaces as a `going-away` status event; provider error messages are recorded without tearing down the session; `agent/input-added` is referenced via `processorDeps` instead of being redeclared.
- Gemini "required" messageAgent tool choice is enforced through the system instruction — Gemini Live v1beta rejects `toolConfig` anywhere in `BidiGenerateContentSetup` (close 1007), so the previous API-level approach silently killed sessions.
- Retired `voice-agent/<provider>` slugs map to a no-op processor so pre-existing stream subscriptions don't error; new streams subscribe only the unified slug.
- Browser console: resubscribes with backoff from the last seen offset, never awaits playback or `AudioContext.resume()` inside the subscription loop, batches mic appends behind a bounded drop-oldest queue, and tracks played offsets with a monotonic counter.
- Audio worklets: allocation-free ring buffers, box low-pass filtering for mic downsampling (no more aliasing decimation), and underruns counted only when audio resumes shortly after a drain.
- Voice code agent uses the standard agent workspace and chat tool; it replies into the voice stream by appending `input-text-appended` events from codemode.

## Why

Proves that a voice client can append PCM input frames to a stream, have a realtime voice backend process them, and receive PCM output frames back through the same stream — with one canonical contract instead of per-provider forks.

## Validation

- `pnpm typecheck` (repo-wide), `pnpm lint`, `pnpm format`
- `packages/shared` test suite incl. new coverage: per-provider audio forwarding, audit redaction, output frame ordering, speaker-buffer-clear, `goAway`, resampling, reducers (19 tests)
- apps/os unit tests (27 files)
- **Live smoke test against all three real provider APIs** (text in → PCM out, in-memory stream): connected/ready, speech-level audio (RMS 1700–3900), correct transcripts, ordered frames, redacted audits with no audio leakage; full messageAgent handoff loop verified on Gemini (tool call → `agent/input-added` → code-agent reply → spoken audio)

## Environment Config Lease

Lease: `preview-6`
Doppler config: `preview_6`
Type: `environment-config-lease`
Leased until: 2026-06-10T12:05:46.980Z

### OS
Status: deployed
Commit: `2a894e2`
Preview: https://os.iterate-preview-6.com
[Workflow run](https://github.com/iterate/iterate/actions/runs/27271812546)
Updated: 2026-06-10T11:08:10.672Z

### Semaphore
Status: deployed
Commit: `2a894e2`
Preview: https://semaphore.iterate-preview-6.com
[Workflow run](https://github.com/iterate/iterate/actions/runs/27271812546)
Updated: 2026-06-10T11:08:05.650Z

### DIFF
--- a/apps/os/alchemy.run.ts
+++ b/apps/os/alchemy.run.ts
@@ -13,7 +13,7 @@ import type { AgentDurableObject } from "./src/domains/agents/durable-objects/ag
 import type { RepoDurableObject } from "./src/domains/repos/durable-objects/repo-durable-object.ts";
 import type { SlackAgentDurableObject } from "./src/domains/slack/durable-objects/slack-agent-durable-object.ts";
 import type { SlackIntegrationDurableObject } from "./src/domains/slack/durable-objects/slack-integration-durable-object.ts";
-import type { VoiceAgentDurableObject } from "./src/domains/voice-agents/durable-objects/voice-agent-durable-object.ts";
+import type { StreamProcessorDurableObject } from "./src/domains/stream-processors/durable-objects/stream-processor-durable-object.ts";
 import type { WorkspaceDurableObject } from "./src/domains/workspaces/durable-objects/workspace-durable-object.ts";
 import type { OutboundMcpFromOurClientCapability } from "./src/domains/outbound-mcp-client/entrypoints/outbound-mcp-from-our-client-capability.ts";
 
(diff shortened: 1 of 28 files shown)


jonastemplestein:
Rebuilt this PR from current origin/main and folded in the follow-up from #1351. The voice provider adapter now runs as hosted processors on AgentDurableObject via the new stream processor host instead of the old standalone stream processor DO split.\n\nVerified locally:\n- pnpm --dir apps/os typecheck\n- pnpm --dir apps/os exec vitest run src/domains/agents/stream-processors/voice-agent/implementation.test.ts

PR #1382, opened 2026-06-08
[codex] Support chunked stream event storage

jonastemplestein:
## What changed

- Split committed stream event JSON into ordered `event_chunks` rows while keeping `events` as the offset/idempotency metadata table.
- Moved the core reduced stream state into a SQL `reduced_state` table.
- Added a Vitest e2e that appends an 8 MiB event, reads it back by offset and range, and verifies a follow-up append still works.

## Why

Large event rows should not depend on storing the entire event JSON in one SQLite text cell. Chunking keeps each stored row bounded while preserving existing RPC read behavior.

This intentionally does not include a legacy `raw_json` migration because production stream data will be deleted before this deploy.

## Review notes

A sub-agent reviewed the first pass for simplicity. I applied its concrete suggestions: use a conservative chunk size, make chunking a generator, drop the speculative `(type, offset)` index, and add an ownership foreign key from chunks to events.

## Validation

- `pnpm --dir packages/streams typecheck`
- `pnpm --dir packages/streams test`
- `STREAM_STAGING_E2E=true WORKER_URL=http://127.0.0.1:5173 pnpm --dir packages/streams/example-app vitest e2e/vitest/stream-capnweb.test.ts --run -t "appends, reads, and keeps running after event rows larger than 2 MiB"`

---

> [!NOTE]
> **Medium Risk**
> Touches core append/read/replay paths for all stream events; no legacy `raw_json` migration in this diff (assumes fresh or wiped production data).
> 
> **Overview**
> Stream Durable Object storage no longer keeps full event JSON in a single `events.raw_json` column. **`events`** now holds only metadata (offset, type, timestamps, idempotency key), and committed JSON is split into ordered **`event_chunks`** rows (512 KiB UTF-8 blobs) with a foreign key and JS-side encode/decode so each SQLite row stays under Cloudflare’s ~2 MiB limit.
> 
> **Reads and replay** (`getEvent`, idempotency lookup, `getEvents` ranges, log recovery) load chunk rows and reassemble JSON in JS instead of parsing one oversized text cell; range queries join the offset window to chunks and group by offset.
> 
> **E2E coverage** adds a staging test that appends a ~2.25 MiB payload, reads it by offset and range, and confirms a follow-up append still works, plus an `it.fails` case documenting the separate **32 MiB inbound WebSocket** cap on client appends.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8ae02f7bb3a5a562765f550bbe96b948bcd56914. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/streams/src/workers/durable-objects/stream.ts
+++ b/packages/streams/src/workers/durable-objects/stream.ts
@@ -25,6 +25,13 @@ import { disposeIgnoredRpcResult, retainProcessEventBatch } from "../rpc-lifecyc
  * HTTP/WebSocket Cap'n Web termination belongs at the fronting Worker, which
  * exposes this DO through `StreamRpcTarget`.
  */
+
+// Cloudflare Durable Objects cap each SQLite string/blob/table row at 2 MB.
+// BLOB columns do not raise that ceiling, and SQL-side substr(?) chunking would
+// still require binding the oversized value first, so event JSON is chunked in JS.
+const EVENT_CHUNK_SIZE = 512 * 1024;
+const textEncoder = new TextEncoder();
+
 export class Stream extends DurableObject<Env> implements StreamRpc {
   #coreProcessorState: StreamCoreProcessorState;
   #coreProcessor = coreProcessor.build({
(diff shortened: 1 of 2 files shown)


packages/streams/src/workers/durable-objects/stream.ts
```diff
@@ -356,6 +445,14 @@ export class Stream extends DurableObject<Env> implements StreamRpc {
     return events;
   }
 
+  #persistAppend(args: {
```
jonastemplestein: this doesn't need to be a separate method! inline it

PR #1424, opened 2026-06-10
perf(os): scope the pretty-JSON interceptor to curl/CLI traffic instead of the browser transport

jonastemplestein:
## Summary

`prettyJsonInterceptor` in `apps/os/src/orpc/handler.ts` re-stringified every JSON response from the OpenAPI handler with 2-space indentation for curl ergonomics. But that handler at `/api` is also the production browser transport — `src/orpc/client.ts` points `OpenAPILink` at `window.location.origin + /api` — so every dashboard API response was buffered, re-serialized, and inflated with whitespace.

Changes:

- Gate the prettification on the request's `User-Agent` matching a command-line HTTP client (`curl`, `HTTPie`, `wget`). Browser traffic (and anything else) now returns the upstream result untouched — the body is never consumed, so streaming responses pass through.
- Extract the interceptor into `apps/os/src/orpc/pretty-json-interceptor.ts` so it can be unit-tested without importing the full `appRouter`.
- Add `pretty-json-interceptor.test.ts` (7 tests): pretty-printing for curl/HTTPie/wget, pass-through for browser UAs and missing UA (asserting the original `Response` instance is returned with `bodyUsed === false`), SSE untouched, unmatched results untouched.

Curl ergonomics are preserved: `curl https://os.iterate.com/api/...` still gets indented JSON.

Note: the `iterate-app-cli` (`pnpm cli rpc ...`) uses `RPCLink` against `orpcRpcHandler` at `/api/orpc/`, which never had this interceptor, so CLI behavior is unchanged.

## Skipped items

None — the single finding was verified accurate and fixed.

## Flags for reviewers

- The UA allowlist is `curl`/`httpie`/`wget` (case-insensitive, word-boundary match). If other terminal tools are used against `/api` (e.g. `xh`), they'll now get compact JSON — easy to add to the regex if desired.
- An `Accept`-based check was considered, but curl sends `Accept: */*` by default, so User-Agent is the only signal that works without extra flags.

## Checks

- `pnpm typecheck` — pass
- `pnpm lint` — pass
- `pnpm exec vitest run src/orpc/pretty-json-interceptor.test.ts` (apps/os) — 7/7 pass
- oxfmt on changed files

---

> [!NOTE]
> **Low Risk**
> Response-shaping change on the OpenAPI adapter only; behavior for browsers improves (pass-through), with curl ergonomics preserved via UA allowlist.
> 
> **Overview**
> Limits **pretty-printed JSON** on the OpenAPI `/api` handler to requests whose `User-Agent` looks like **curl, HTTPie, or wget**. Browser dashboard traffic (and anything else) now skips the interceptor entirely so responses are not buffered or re-serialized.
> 
> The interceptor moves from `handler.ts` into **`pretty-json-interceptor.ts`**, with **Vitest** coverage for CLI pretty-printing, browser/missing-UA pass-through (same `Response`, `bodyUsed` unchanged), SSE, and unmatched routes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1c7745d6dfa8209ea5e6597ee160c90b6e619ac1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Environment Config Lease

### DIFF
--- a/apps/os/src/orpc/handler.ts
+++ b/apps/os/src/orpc/handler.ts
@@ -2,36 +2,18 @@ import { EvlogHandlerPlugin } from "@iterate-com/shared/evlog/orpc-plugin";
 import { OpenAPIHandler } from "@orpc/openapi/fetch";
 import { OpenAPIReferencePlugin } from "@orpc/openapi/plugins";
 import { ZodToJsonSchemaConverter } from "@orpc/zod/zod4";
-import { onError, type Context } from "@orpc/server";
+import { onError } from "@orpc/server";
 import { RPCHandler } from "@orpc/server/fetch";
-import type { FetchHandleResult, FetchHandlerInterceptorOptions } from "@orpc/server/fetch";
 import { experimental_RPCHandler as WebSocketRPCHandler } from "@orpc/server/crossws";
 import { CORSPlugin } from "@orpc/server/plugins";
 import packageJson from "../../package.json" with { type: "json" };
 import type { RequestContext } from "~/request-context.ts";
 import { appRouter } from "~/orpc/root.ts";
+import { prettyJsonInterceptor } from "~/orpc/pretty-json-interceptor.ts";
 import { prettifyStandardSchemaError } from "~/standard-schema/errors.ts";
 import { looksLikeStandardSchemaFailure } from "~/standard-schema/utils.ts";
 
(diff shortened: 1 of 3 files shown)


jonastemplestein:
dumb PR - we will rip out orpc

PR #1458, opened 2026-06-10
fix(streams): Stage 1 — failed-batch desync, checkpoint ordering, RPC surface (C1/C2/C3)

jonastemplestein:
## What

**Stage 1** of the `packages/streams` review (plan in `tasks/streams-review-fixes.md`). Fixes the three CRITICAL findings. Flips the Stage 0 ratchets **T1, T1b, T2, T3** from `it.fails` → `it`.

**Stacked on #1455** (Stage 0 tests) — base is `streams-review-stage0-tests`; retarget to `main` once that merges.

## C1 — a failed batch no longer drops events (data loss)

The delivery pump is fire-and-forget and advances its cursor before delivery, so a swallowed ingest failure left a permanent gap in the subscriber's state. Per the agreed approach (consumer resubscribes from checkpoint):

- On ingest failure the host re-handshakes from the durable checkpoint; the stream replays the batch (replay is idempotent).
- **Continued-delivery race:** because the pump keeps delivering during recovery, a *later* batch's successful ingest would advance the checkpoint past the gap. Each subscription is now tagged with a `generation`, and ingest runs through a per-processor serial chain that re-checks the generation **between** batches — batches from the superseded connection are dropped, and the post-recovery replay is the single source of truth.
- **Poison policy (owner's call):** after `MAX_CONSECUTIVE_INGEST_FAILURES` (3), the host appends a `stream/error-occurred` event and disconnects rather than hot-looping; recovery is then up to the subscriber/processor.

Covered by **T1** (transient recovery under continued delivery) + **T1b** (poison → error + disconnect).

> Note: the browser-mirror consumer (`stream-browser-store.ts`) has the same swallow-and-wedge bug. Left for Stage 4 with the other browser-runtime fixes (tracked in the task).

## C2 — checkpoint ordering (silent data loss on write failure)

`StreamProcessor.#ingest` advanced `#state`/`#checkpointOffset` *before* `writeState`, so a failed durable write left the in-memory checkpoint ahead and the retried batch silently no-oped. Now it persists the snapshot **first**, then advances. (Inlined the single-use `#saveSnapshot`.) Covered by **T2**.

## C3 — `PublicStreamRpcTarget` method leak (security)

`makeRpcTargetClass` copied every prototype method except a two-item denylist, so the `protected` `readCoreProcessorState` / `writeCoreProcessorState` were proxied onto the unauthenticated public target — `writeCoreProcessorState` could inject an attacker-chosen subscription callable that the DO later dispatches with full `env`. Added an `include` allowlist; both Stream targets now allowlist the explicit `StreamRpc` surface (`satisfies readonly (keyof StreamRpc)[]` so it can't drift). Covered by **T3**.

## Verification

```
node pool:    61 passed | 2 expected-fail   (T5/T6 remain → Stage 3)
workers pool:  7 passed | 1 expected-fail   (T4 remains → Stage 3)
typecheck: clean   lint: clean
```

---

> [!NOTE]
> **High Risk**
> Changes durable ingest/checkpoint behavior and closes a security hole on unauthenticated Stream RPC; incorrect recovery or ordering could still cause data loss or wedge delivery.
> 
> **Overview**
> **Stage 1** closes three CRITICAL streams review findings (C1/C2/C3) and turns regression tests T1, T1b, T1c, T2, and T3 from `it.fails` into passing `it`.
> 
> **C1 — failed batches no longer lose events (hosted path).** `createStreamProcessorHost` no longer swallows ingest errors. On failure it re-handshakes from the durable checkpoint so the stream replays the batch. A **generation** gate plus a per-processor **ingest chain** drops batches from superseded connections while recovery runs (avoids advancing the checkpoint past a gap). After three consecutive failures it appends `stream/error-occurred` and disconnects instead of hot-looping. Regression coverage adds a replay-capable `fakeStream`, T1b (poison), and T1c (recovery throw must not wedge the chain).
> 
> **C2 — checkpoint vs durable write.** `StreamProcessor.#ingest` now **`await`s `writeState` before** updating in-memory `#state` / `#checkpointOffset`, so a failed write leaves the batch retryable instead of a silent no-op.
> 
> **C3 — public RPC surface.** `makeRpcTargetClass` gains an **`include` allowlist**; `PublicStreamRpcTarget` / `StreamRpcTarget` only expose explicit `StreamRpc` methods so protected `readCoreProcessorState` / `writeCoreProcessorState` are not callable on the unauthenticated public target.
> 
> Task doc and test headers are updated to mark C1/C2/C3 fixed on the hosted path; browser mirror recovery remains Stage 4.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e5aeaac53f822dc90a70ba3e92ece0bd81112b1e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Environment Config Lease

### OS
Status: released
Commit: `a9d4a4d`
[Workflow run](https://github.com/iterate/iterate/actions/runs/27285949769)
Updated: 2026-06-10T15:34:34.983Z

### DIFF
--- a/packages/streams/src/shared/rpc-target.ts
+++ b/packages/streams/src/shared/rpc-target.ts
@@ -26,19 +26,24 @@ export function makeRpcTargetClass<
   const TExcluded extends keyof TSource = never,
 >(
   sourceClass: { prototype: TSource },
-  options?: { exclude?: readonly TExcluded[] },
+  options?: { exclude?: readonly TExcluded[]; include?: readonly PropertyKey[] },
 ): RpcTargetClass<RpcMethods<TSource, TExcluded>, TSource>;
 
 export function makeRpcTargetClass<TApi extends object, TSource extends object = TApi>(
   sourceClass: { prototype: TSource },
-  options?: { exclude?: readonly PropertyKey[] },
+  options?: { exclude?: readonly PropertyKey[]; include?: readonly PropertyKey[] },
 ): RpcTargetClass<TApi, TSource>;
 
 export function makeRpcTargetClass<TSource extends object>(
   sourceClass: { prototype: TSource },
-  options: { exclude?: readonly PropertyKey[] } = {},
+  options: { exclude?: readonly PropertyKey[]; include?: readonly PropertyKey[] } = {},
 ): RpcTargetClass<object, TSource> {
   const exclude = new Set<PropertyKey>(["constructor", ...(options.exclude ?? [])]);
+  // When `include` is given it is an allowlist: only those methods are proxied.
+  // Prefer it for classes with internal (e.g. `protected`) methods — `protected`
+  // does not exist at runtime, so a denylist silently leaks any method not
+  // explicitly excluded.
+  const include = options.include === undefined ? undefined : new Set<PropertyKey>(options.include);
 
   class GeneratedRpcTarget extends RpcTarget {
     constructor(readonly source: TSource) {
@@ -50,6 +55,9 @@ export function makeRpcTargetClass<TSource extends object>(
     Object.getOwnPropertyDescriptors(sourceClass.prototype),
   )) {
     if (exclude.has(key)) {
+      continue;
+    }
+    if (include !== undefined && !include.has(key)) {
       continue;
     }
 
--- a/packages/streams/src/stream-processor.ts
+++ b/packages/streams/src/stream-processor.ts
@@ -366,14 +366,21 @@ export abstract class StreamProcessor<
       await Promise.all(blockingWork);
     } catch (error) {
       // A failed batch must still settle work it already registered so nothing
-      // rejects unobserved. The checkpoint is not written; the batch retries.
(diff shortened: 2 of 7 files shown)


jonastemplestein:
Addressed the Bugbot finding in d89a4a9: the ingest chain now has a trailing `.catch` (logs + resolves) so a throwing recovery can't leave it permanently rejected and wedge later batches — mirroring `StreamProcessor.ingest`'s own `#processing` guard. Added regression test **T1c** (verified failing without the catch). Stage 3 (#1459) rebased on top.

jonastemplestein:
Addressed the second Bugbot finding (generation invalidation) in e5aeaac: `requestStreamSubscription` now bumps `entry.generation` before re-opening, same as `recoverFromIngestFailure`, so a prior connection's in-flight batches are dropped by the gate and the new connection's replay is authoritative. #1459 rebased on top.

PR #1466, opened 2026-06-10
itx: project DO sheds its hats — event-sourced creation, stateless ingress, fetch = egress

jonastemplestein:
## What

The Project Durable Object was four products in one 1473-line class. After this PR it has exactly three documented jobs (see the new file header): **itx registry supervisor**, **worker source-of-truth**, and **egress authority**. Net diff: +567 / −1628.

### Ingress leaves the DO
`ProjectIngressEntrypoint` (stateless) now serves *all* project-host traffic: one light `getWorkerVersion()` RPC (freshness TTL + deduped background rebuild semantics live in the DO), then it loads the worker itself via `env.LOADER` with the shared cache key — the code payload only crosses RPC inside the loader's miss callback. Landing/building fallbacks and app-slug parsing moved with it. The DO is just where the source code lives.

### fetch on the project IS egress
The worker's `fetch` is the project's homepage; the project's `fetch` is egress — matching itx vocabulary where `itx.fetch` ≡ project egress. `ingressFetch` is deleted. The intercept tunnel's WebSocket handshake stays on the DO fetch path (upgrades can't cross RPC); the header documents the endgame: stateless egress capability + captun replaced by a live egress-shadowing cap over capnweb-with-WebSockets.

### Creation is event-sourced (`ProjectProcessor`, slug `project`)
```
project/create-requested    { projectId, slug }        appended by createProject
project/created             { …, defaultHost, hosts }  registered, hosts assigned
project/config-worker-built { commitOid, … }           build fact (string unchanged)
project/create-completed    { projectId }              registration done
```
`processEvent` drives the idempotent setup steps (example secret, agents root + rule, created/completed events); the worker build runs in the background and never gates creation — ingress self-heals failed builds. `reduce` now projects **all** contract events into `{ phase, project, worker }`. Deleted: the DO's `afterAppend`, `getProjectLifecycleRunnerState`, and the fake legacy runner shape. The project worker's event hook is renamed `afterAppend` → `processEvent` (base seed updated).

### Storage teardown
- DO SQLite tables `project_state` and `project_ingress_routes` deleted; summary derives from the processor snapshot with a pure-compute D1 fallback (`projectHosts()` is a pure function of projectId/slug/config). Only the itx registry table and the checkout KV cache remain.
- D1 `ingress_routes` table **destroyed** (migration `0015_drop_ingress_routes.sql`) — platform hosts already resolve dynamically from the `projects` table; the materialized rows were stale duplicates. `lookup.ts` flattens to one priority chain: itx cap host → platform host (slug/projectId) → custom hostname.

### "config worker" is dead in code
New `durable-objects/worker.ts`: `WorkerHost` (clone → bundle → validate → cached checkout, 10s TTL) plus the pure load helpers shared with the ingress entrypoint. `callWorkerFunction` / `getWorkerVersion` / `getWorkerCheckout`; `itx.worker.fetch` routes through the ingress entrypoint. Durable strings (event type, storage keys, loader cache key) unchanged.

## Compatibility

**None, deliberately.** Existing DOs hold stream subscriptions pointing at processor name `project-lifecycle` and first-init won't re-run, so already-created projects won't pump events into the new processor (summary still works via the D1 fallback). Assumes the usual stage destroy/redeploy; a one-off resubscribe script would cover prod if ever needed.

## Testing

- `test:project-ingress` (6/6) — covers stateless ingress dispatch incl. async loader `getCode`, app slugs, custom hostnames, creation events, processor snapshot, egress secret substitution, and the intercept tunnel.
- `test:itx-stream-subscribe` (9/9).
- `test:codemode-session` (1 failure) and `test:project-mcp-server-connection` (3 failures) fail **identically on clean origin/main** — pre-existing.
- Not run: itx e2e (needs a deployed worker); `sqlfu:check` (needs a materialized alchemy config in this fresh worktree — the migration is a one-line DROP TABLE).

---

> [!NOTE]
> **High Risk**
> Large architectural change to project routing, creation, and ingress/egress boundaries; existing DOs may need redeploy/resubscribe for stream processor migration.
> 
> **Overview**
> The **Project Durable Object** is narrowed to three roles (registry, worker source-of-truth, egress). **Ingress** moves to the stateless `ProjectIngressEntrypoint`, which calls `getWorkerVersion()` on the DO and loads the dynamic worker via `env.LOADER` locally instead of routing through the DO.
> 
> **`project.fetch` is egress** (worker `fetch` = homepage); `itx.worker.fetch` goes through the ingress entrypoint. **Creation** becomes event-driven: `createProject` appends `project/create-requested` and returns immediately while the new **`ProjectProcessor`** runs idempotent setup steps and exposes `phase` / `processor.snapshot()` for progress.
> 
> **Storage cleanup:** D1 `ingress_routes` and related queries are removed; host lookup uses the `projects` table only. DO-local `project_state` / ingress route tables go away in favor of processor state + D1 fallback. Worker build logic is extracted to **`WorkerHost`** in `worker.ts`; **`callConfigWorkerFunction`** → **`callWorkerFunction`**.
> 
> **itx:** `itx.project` is wrapped in a path proxy so `itx.project.processor.snapshot()` works in one expression; fallthrough proxy no longer binds getter results (fixes `bind` path-segment bug). Docs add §9 egress-as-capability roadmap and mark worker rename as shipped.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 731e5c68d10ba47df5640b6ec16e3bd979989e42. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Environment Config Lease

### OS
Status: released
Commit: `731e5c6`
Preview: https://os.iterate-preview-5.com
[Workflow run](https://github.com/iterate/iterate/actions/runs/27307697912)
Updated: 2026-06-10T21:37:38.888Z

### Semaphore
Status: released
Commit: `ceb03f8`
Preview: https://semaphore.iterate-preview-5.com
[Workflow run](https://github.com/iterate/iterate/actions/runs/27305690873)
Updated: 2026-06-10T21:37:19.430Z


apps/os/src/domains/projects/stream-processors/project-processor.ts
```diff
@@ -0,0 +1,218 @@
...
+
+  protected override async processEventBatch(
+    args: Parameters<StreamProcessor<ProjectProcessorContract>["processEventBatch"]>[0],
+  ): Promise<void> {
+    await super.processEventBatch(args);
+    // Forward EVERY live event (consumed or not) to the project worker.
+    for (const event of args.events) {
+      if (event.offset <= args.sideEffectsAfterOffset) continue;
+      args.runInBackground(() => this.deps.forwardEventToWorker(event));
+    }
```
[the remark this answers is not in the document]
jonastemplestein: Fixed in c92b958cb: the contract now consumes `"*"" so the subscription delivers unfiltered batches to `processEventBatch`. The not-ready early return is now documented as the deliberate contract: forwarding is best-effort and begins once a worker build exists (same semantics as the old `afterAppend` ready-flag gate); pre-build events are not replayed.

apps/os/src/domains/projects/entrypoints/project-ingress-entrypoint.ts
```diff
@@ -17,10 +41,120 @@ export class ProjectIngressEntrypoint extends WorkerEntrypoint<
...
+      const worker = this.env.LOADER.get(
+        workerCacheKey({ commitOid: version.commitOid, projectId: summary.id }),
+        async () => {
+          const checkout = await project.getWorkerCheckout();
+          return withWorkerEnv({
+            exports: readLoopbackExports(this.ctx.exports),
+            projectId: summary.id,
+            workerCode: checkout.workerCode,
+          });
+        },
```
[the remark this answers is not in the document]
jonastemplestein: Acknowledged as benign and documented in c92b958cb: if a rebuild lands between the version RPC and the cold-isolate miss callback, newer code loads under the stale key for that one request; the next request keys the new commit and converges, and the stale isolate ages out. No correctness impact (the code served is newer, never older).

apps/os/src/domains/projects/durable-objects/project-durable-object.ts
```diff
@@ -251,145 +182,128 @@ const ProjectLifecycleBase = createIterateDurableObjectBase<
...
+          buildWorker: async (project) => {
+            const checkout = await this.workerHost.buildFresh({
+              id: project.projectId,
+              slug: project.slug,
+            });
+            return {
+              commitOid: checkout.commitOid,
+              mainModule: checkout.workerCode.mainModule,
+            };
+          },
```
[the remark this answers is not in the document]
jonastemplestein: Fixed in c92b958cb: `WorkerHost.buildFresh` now serializes every build through a single promise chain, so creation-step builds, ingress background rebuilds, and `callWorkerFunction` builds can no longer interleave clone/read cycles in the shared workspace checkout dir.

apps/os/src/domains/projects/durable-objects/project-durable-object.ts
```diff
@@ -78,6 +108,8 @@ const ProjectStructuredName = z.object({
   projectId: z.string(),
 });
 
+const STREAM_SUBSCRIPTION_CONFIGURED_TYPE = "events.iterate.com/stream/subscription-configured";
```
jonastemplestein: Don't like these - put the strings inline

apps/os/src/domains/projects/durable-objects/project-durable-object.ts
```diff
@@ -134,112 +146,12 @@ type ProjectEnv = {
...
-type ProjectConfigWorkspaceStub = {
-  cloudflareShellGit(): Promise<unknown>;
-  cloudflareShellState(): Promise<Record<string, unknown>>;
-  hasFile(path: string): Promise<boolean>;
-  initialize(input: { name: string }): Promise<unknown>;
-  removePath(input: { force: boolean; path: string; recursive: boolean }): Promise<void>;
-};
-
-const ProjectLifecycleBase = createIterateDurableObjectBase<
+const ProjectDurableObjectBase = createIterateDurableObjectBase<
```
jonastemplestein: Let's not use this mixin - it's too complicated now that this is the _only_ domain object. just have a single D1 binding in the apps/os worker and make it use super vanilla sqlfu and then make the project processor write to it as a side-effect. it should also cross-post project creation requested events to /projects in the "global" namespace

apps/os/src/domains/projects/durable-objects/project-durable-object.ts
```diff
@@ -251,145 +163,141 @@ const ProjectLifecycleBase = createIterateDurableObjectBase<
...
+      }),
+    cloneRepo: (input) => this.cloneWorkerRepo(input),
+    bundle: (files) => this.bundleWorkerCode(files),
+  });
+  projectProcessor = this.host.add(
+    ProjectProcessorContract.slug,
+    (deps) =>
+      new ProjectProcessor({
+        ...deps,
+        creation: {
```
jonastemplestein: The deps seems sketch - why not move a lot of the logic from these functions _into_ the processor and have higher level deps like workerHost or env bindings? it's fine - especially given these methods should still be accessible on itx via itx.project.projectProcessor.someMethod (because projectProcessor is a public property and also an rpc target? or not?)

apps/os/src/domains/projects/durable-objects/project-durable-object.ts
```diff
@@ -251,145 +163,141 @@ const ProjectLifecycleBase = createIterateDurableObjectBase<
...
+              id: project.projectId,
+              slug: project.slug,
+            });
+            return {
+              commitOid: checkout.commitOid,
+              mainModule: checkout.workerCode.mainModule,
+            };
+          },
+        },
+        forwardEventToWorker: (event) => this.forwardEventToWorker(event),
```
jonastemplestein: why doesn't the processor just use its itx dep for this? itx.project.worker.processEvent() or sth like that? not sure exactly

apps/os/src/domains/projects/durable-objects/project-durable-object.ts
```diff
@@ -251,145 +163,141 @@ const ProjectLifecycleBase = createIterateDurableObjectBase<
...
-      priority INTEGER NOT NULL,
-      notes TEXT,
-      callable_json TEXT NOT NULL,
-      created_at_ms INTEGER NOT NULL,
-      updated_at_ms INTEGER NOT NULL
-    )`);
-    sql.exec(`CREATE INDEX IF NOT EXISTS idx_project_ingress_routes_host
-      ON project_ingress_routes (host)`);
-
     this.registerOnFirstInitialize(async (params) => {
```
jonastemplestein: Can we do without this lifecycle mixin for the moment? i think it's getting in the way and i think this should potentially all be code that lives in the _processor_ anyway

apps/os/src/domains/projects/durable-objects/project-durable-object.ts
jonastemplestein: Done in bad4740a6 — the `STREAM_SUBSCRIPTION_CONFIGURED_TYPE` const and all four `*_EVENT_TYPE` consts are gone; event type strings are inline at every use site (contract events map, reduce cases, appends, test assertions), matching the slack contract style. The `ProjectStructuredName` type + zod schema died with the mixin too.

apps/os/src/domains/projects/durable-objects/project-durable-object.ts
jonastemplestein: Done in bad4740a6. The DO now extends plain `DurableObject` — no mixin, no DO_CATALOG, no initialize/ensureStarted anywhere. Project DOs are addressed by the PLAIN project id (`getProjectDurableObjectName` is identity; `ctx.id.name` reads it back). The D1 `projects` upsert is now the processor's first creation step (vanilla `env.DB.prepare` on the one binding) — still safe for create-then-immediately-update flows because `createProject` waits for create-completed. And create-requested is cross-posted to `/projects` in the `global` namespace (asserted in the ingress test) — this also answers itx-next's open question 4 (global stream as the project-lifecycle audit surface) with a yes.

apps/os/src/domains/projects/durable-objects/project-durable-object.ts
jonastemplestein: Done in bad4740a6 — deps are now exactly what you suggested: `{ env, exports, workerHost, appConfig }`, and ALL step logic (D1 projection, secret, agents root, global cross-post, build, forwarding) lives in the processor. Split into `stream-processors/project/contract.ts` + `implementation.ts` per the house pattern, because event-docs imports the contract from Node and the implementation now touches cloudflare:workers.

On `itx.project.projectProcessor.someMethod`: `projectProcessor` IS a public property, but `StreamProcessor` extends **capnweb's** RpcTarget, not cloudflare:workers' — so the property doesn't cross the Workers-RPC stub that `itx.project` hands out. Today the snapshot is exposed via `getProjectState()`. If we want processors directly traversable, they'd need to extend (or be wrapped in) the workers RpcTarget — happy to do that as a follow-up rather than smuggle it in here.

apps/os/src/domains/projects/durable-objects/project-durable-object.ts
jonastemplestein: Moved the forwarding INTO the processor in bad4740a6 (no more bespoke closure dep), but via `workerHost` rather than `itx.worker.processEvent`: the itx.worker path replays through `callWorkerFunction`, which builds a FRESH checkout per call — right for tool calls (always latest pushed config), absurd per forwarded event. Forwarding wants the warm cached isolate, which is exactly what `workerHost.load(cachedCheckout)` gives. There's a comment at the forwarding site saying this; if we later make callWorkerFunction reuse the fresh-with-TTL dispatch semantics, the itx.worker spelling becomes viable.

apps/os/src/domains/projects/durable-objects/project-durable-object.ts
jonastemplestein: Done in bad4740a6 — the mixin is gone entirely (this thread + the one above), and there's no constructor left at all: the subscription bootstrap moved into `createProject` as an idempotent append. That's the ONE creation step that can't live in the processor — the processor only sees events after the subscription exists (bootstrap), so the method that mints the first event also pins the subscription. Everything else creation-related is processor code now.

apps/os/src/domains/projects/durable-objects/project-durable-object.ts
jonastemplestein: That's actually another thing. I don't think `createProject` should wait for `create` completed. I think it should be like this, and it's okay to rip out some stuff here. I think:
1. The project ID should come from the auth service.
2. We should not wait for the `created` event; we should just immediately redirect to that project page.

apps/os/src/domains/projects/durable-objects/project-durable-object.ts
jonastemplestein: Are you sure about this? I thought in the case of cap, if you thought a captain web RpcTarget is the same as a workers RpcTarget when loading in workers Rpc. I would like them to be directly and transparently traversable on itx.

apps/os/src/domains/projects/durable-objects/project-durable-object.ts
jonastemplestein: But just to be clear, we want to do as much as possible in the actual processor, so if there's any stuff that needs to be created, like repos, artefacts, or whatever, it should all happen in the processor and leave a trail of events.

apps/os/src/domains/projects/durable-objects/project-durable-object.ts
jonastemplestein: Done in 0f4cf51ec — `createProject` is fire-and-return now: append create-requested, return the purely computed summary, no waiting. Callers redirect immediately and can watch `projectProcessor.snapshot()` (phase: creating → ready) for progress. The egress e2e now polls readiness instead of assuming born-with guarantees, and the ingress test entry inserts the projects row before dialing the DO — exactly what the production callers (directory, itx.projects.create) already do.

On the project ID: that's already the model — auth is the canonical minter (`prj_` via `generateId` in apps/auth), and OS adopts the id auth returns for user creates. The only places OS mints (`mintProjectId`, same `prj_` format) are the operator/recovery paths where no auth org exists to own the project — documented in `src/domains/projects/project-id.ts`. If you want even those to round-trip through auth, auth needs an org-less mint endpoint; say the word and I'll spec it as a follow-up.

apps/os/src/domains/projects/durable-objects/project-durable-object.ts
jonastemplestein: You were right, I was wrong — capnweb's workerd build does `RpcTarget = workersModule.RpcTarget`, so processors ARE workers-RPC traversable. Verified empirically in the workers tests and shipped in 0f4cf51ec: `projectProcessor` is exposed as a prototype getter (own instance fields don't cross Workers RPC — that part of the visibility rules is real), `getProjectState()` is deleted, and the orpc handler + tests + the egress e2e all traverse `itx.project.projectProcessor` directly.

One workerd quirk found while testing, documented at the getter: calls don't pipeline through property accesses in a single expression (`stub.projectProcessor.snapshot()` → "receiver does not implement"), but awaiting the property first works perfectly: `const p = await itx.project.projectProcessor; await p.snapshot()`.

apps/os/src/domains/projects/durable-objects/project-durable-object.ts
jonastemplestein: Agreed and extended in 0f4cf51ec: the iterate-config repo is now an explicit creation step in the processor (it was previously only materialized lazily by the worker build), and it leaves its own fact on the stream — `project/repo-initialized` { projectId, repoSlug, defaultBranch }. The full trail for a create is now: create-requested → created → repo-initialized → create-completed (+ config-worker-built whenever the build lands), with create-requested also cross-posted to global /projects. Anything else a project grows at birth should follow the same pattern: a processor step + an event.

apps/os/src/domains/projects/durable-objects/project-durable-object.ts
jonastemplestein: this is lame ...   maybe do #processor = and have a `get processor()` around it or sth? would that help

jonastemplestein: Solved properly in 395ed3346 — the quirk only applies to PROPERTY accesses; workerd happily pipelines calls on CALL results. So it's now a method: `await itx.project.processor().snapshot()` works in one expression (verified in the workers tests — the project-state endpoint and the egress e2e readiness poll both use the single-expression form). The `#projectProcessor = …` private field + public `processor()` shape is exactly your suggestion, just with parens instead of a getter.

jonastemplestein: Even better in 5f052fb14: the bare property spelling `await itx.project.processor.snapshot()` now works in one expression. `itx.project` wraps the DO stub in the existing PathProxyRpcTarget → replayPathCall, which awaits each intermediate property segment — exactly the boundary workerd needs. D17's whole-surface posture is unchanged (the proxy is generic, zero hand-maintained forwarders), and reserved/prototype segments stay gated in replayPathCall. The DO is back to `get processor()` per your suggestion. Only code holding RAW Workers stubs (orpc handler, test entries) still awaits the property first, with comments saying why.

apps/os/src/domains/projects/durable-objects/project-durable-object.ts
```diff
@@ -130,267 +113,127 @@ type ProjectEnv = {
...
-  LOADER: ProjectDynamicWorkerLoader;
+  LOADER: WorkerLoaderBinding;
   WORKSPACE: DurableObjectNamespace;
 };
 
-type ProjectConfigCheckout = {
-  commitOid: string;
-  workerCode: ProjectDynamicWorkerCode;
-};
-
```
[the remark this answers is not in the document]
jonastemplestein: Fixed in 867b18ebc — `createProject` now refuses loudly when `input.projectId` doesn't match the DO's own name (which IS the project id). All real callers derive the stub and the input from the same id, so this is a pure tripwire against miswired dials.

apps/os/src/itx/handle.ts
```diff
@@ -164,37 +180,43 @@ export class Itx extends RpcTarget {
...
+   * does not pipeline calls through property accesses, so on a raw stub
+   * `stub.processor.snapshot()` throws. The proxy accumulates the path and
+   * replayPathCall awaits each intermediate segment, so deep traversal works
+   * in one expression: `await itx.project.processor.snapshot()`. Reserved/
+   * prototype path segments stay gated inside replayPathCall.
    */
   get project(): ProjectStub {
-    return this.#projectStub();
+    const stub = this.#projectStub();
+    return new PathProxyRpcTarget((call) => replayPathCall(stub, call)) as unknown as ProjectStub;
```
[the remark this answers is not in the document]
jonastemplestein: Empirically refuted — this exact path runs green against live deployments. The egress e2e's readiness poll does `projectItx.project.processor.snapshot()` from Node over a `connectItx` Cap'n Web WebSocket on every test (itx-egress.e2e.test.ts: waitForProjectReady), and the Preview/e2e check passed on both runs since the path proxy landed (716568980 and the merge commit ceb03f804). Cap'n Web marshals Functions by reference (its serializer treats `value instanceof Function` like an RpcTarget), and PathProxyRpcTarget is a callable Proxy — so property gets pipeline back to the server where replayPathCall (and its reserved-segment gate) runs, same as itx.worker has always worked over capnweb. The `value.bind is not a function` failure cited here was real but was the previous commit's bug (the handle's fallthrough proxy binding getter results), fixed in 716568980 with a workers-pool regression test (`/__test/itx-project-processor-phase`).

apps/os/src/domains/projects/stream-processors/project/implementation.ts
```diff
@@ -0,0 +1,297 @@
...
+          idempotencyKey: `project-config-worker-built:${projectId}:${checkout.commitOid}`,
+          payload: {
+            commitOid: checkout.commitOid,
+            mainModule: checkout.workerCode.mainModule,
+            projectId,
+            repoSlug: ITERATE_CONFIG_REPO_SLUG,
+          },
+        },
+      });
+    });
```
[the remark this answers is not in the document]
jonastemplestein: Fixed in 4ce594703 — and it was worth fixing properly: anyone holding a project's itx handle can append arbitrary events to its streams, so a crafted create-requested on project A's stream could have run creation side effects for an arbitrary id (including overwriting another project's slug via the D1 upsert). The processor now checks every creation event's payload.projectId against the hosting DO's own id (`#ownEvent`): `reduce` ignores foreign events, `processEvent` warns and skips. Covered by a test that appends a spoofed create-requested, waits for the checkpoint to pass its offset, and asserts no reduced-state change and no D1 row for the foreign id.

apps/os/src/domains/projects/stream-processors/project/implementation.ts
```diff
@@ -0,0 +1,321 @@
...
```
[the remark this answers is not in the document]
jonastemplestein: Fixed in 913e4c543 — WorkerHost now takes an `onBuilt` hook and the hosting DO appends the config-worker-built fact after EVERY successful build (ingress staleness rebuilds and callWorkerFunction builds included), with the per-commit idempotency key deduping same-commit rebuilds. The processor's creation step just kicks the build and appends nothing, so there's exactly one source of the event and processor state tracks what dispatch actually serves.

PR #1467, opened 2026-06-10
feat(streams): lazy stream initialization (Stage 2)

jonastemplestein:
## What

**Stage 2** of the `packages/streams` review: a stream is no longer initialized eagerly on first touch — it initializes lazily on its first append.

- **Connect to a stream that doesn't exist → empty.** Subscribing/reading a never-appended stream creates no storage and leaves no trace (not in any parent's `childPaths`).
- **First append initializes it:** prepends `created`@1 + `woken`@2, then your event@3. Both lifecycle events are kept; each later DO incarnation still appends its own `woken` on wake. State *after* the first append is identical to the old eager behavior — only the open→first-append window is now empty.

## Implementation (Stream DO)

- Constructor does no writes for an uninitialized stream; only detects existing storage (`#eventsTableExists()` → `#storageReady`). An initialized stream still appends `woken` + reconciles on wake.
- `#appendBatchHere` ensures the schema and prepends `[created, woken]` on the first append, stripped from the input-aligned return.
- Read paths short-circuit when storage isn't ready; `readCoreProcessorState` only replays the log when it exists.
- **Path resolution** uses the DO's own name (`#streamIdentity()`), not reduced state (the `"uninitialized"` placeholder pre-init would misroute relative/child appends).

## Tests — verified locally

Always-on DO tests (vitest-pool-workers): empty fresh stream, first-append created@1/woken@2/event@3, woken-per-incarnation, relative-child resolution on a never-initialized parent. **node 67 + workers 12 green.**

example-app **e2e** rewritten for empty-on-visit and **run locally** against a `vite dev` worker (Playwright auto-starts the dev server; chromium only; no Doppler): **Playwright 26/26, capnweb 18 + 1 expected-fail.** Running it caught two real bugs (fixed here): relative-path misrouting on a lazy parent, and the example app naming the circuit-breaker runner DO from the lazy `core.path` (→ shared runner that never tripped) — now uses the real `streamView.path`.

## ⚠️ Needs rebase onto #1460

`#1460` (subscriber-presence facts + reconciler homogenization) landed on main after this branch and significantly reworked the Stream DO (CORE_STATE_VERSION→4, core processor owns reconciliation, `processor-registered` → `subscriber-connected`). This branch now conflicts and needs a deliberate rebase that re-applies lazy init on the new reconciler model — not a mechanical merge.

### DIFF
--- a/packages/streams/example-app/e2e/playwright/stream-browser.spec.ts
+++ b/packages/streams/example-app/e2e/playwright/stream-browser.spec.ts
@@ -34,8 +34,6 @@ test("stream page appends through the shared browser mirror", async ({ page }) =
   const streamPath = `/e2e/${crypto.randomUUID()}`;
   await page.goto(streamRoute({ path: streamPath }));
 
-  await expect(eventMeta(page, "events.iterate.com/stream/created").first()).toBeVisible();
-
   const type = "events.iterate.com/debug/playwright-single";
   await appendComposerEvent(page, {
     type,
@@ -48,7 +46,6 @@ test("stream page appends through the shared browser mirror", async ({ page }) =
 test("sidebar circuit breaker config runs the hosted built-in processor", async ({ page }) => {
   const streamPath = `/e2e/${crypto.randomUUID()}`;
   await page.goto(streamRoute({ path: streamPath }));
-  await expect(eventMeta(page, "events.iterate.com/stream/created").first()).toBeVisible();
 
   await page.getByTestId("circuit-breaker-burst-capacity").fill("1");
   await page.getByTestId("circuit-breaker-refill-rate").fill("1");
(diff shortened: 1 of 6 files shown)


jonastemplestein:
Backing out Stage 2 (lazy initialization). Keeping eager init.

Reason: #1460 (subscriber presence facts) landed and makes **subscribing append a `subscriber-connected` event**. Under lazy init, that means merely *connecting* to a stream would initialize it — which is the exact opposite of the goal ("connect without initializing; leave no trace unless you append"). Reconciling the two would require reworking #1460's just-merged presence model so presence isn't recorded on uninitialized streams, which isn't worth the complexity. Eager init (created + woken on first touch, presence on connect) stays.

Stages 0/1/3 of the streams review (the bug fixes) are already merged; only this lazy-init stage is abandoned.

PR #1471, opened 2026-06-10
fix(streams+agent): re-dial dead outbound subscribers, resurrect scheduled LLM turns after DO eviction

jonastemplestein:
## Context

Root-cause fixes from today's prd Slack agent outage (second incident of the day). A deploy landed at 16:02:09 UTC — three seconds after a Slack message in `#...C08R1SMTZGD` — evicted the agent host DO mid-run, and the agent never responded to that message or any later one. Diagnosis trail: thread stream `/agents/slack/c08r1smtzgd/ts-1781107326-880549` shows full agent setup then nothing (no LLM request, no error), and observability shows zero completed `SlackAgentDurableObject` invocations account-wide after the deploy.

Two independent bugs compose into "agent permanently silent":

## 1. Streams kernel: dead outbound connections never re-dialed (M1 from `tasks/streams-review-fixes.md`)

Batch delivery into a subscriber stub was fire-and-forget with the result disposed immediately and rejections discarded. When a subscriber DO dies (deploy eviction, abort), the broken connection stays in `#connections`, the pump keeps advancing its cursor into the dead stub, and `#reconcile` skips keys that already have a connection — so the subscriber is never re-dialed and delivery stalls silently for the rest of the stream incarnation.

- `retainProcessEventBatch` now takes `onDeliveryError` and observes the delivery result (rejections **and** synchronous stub throws), disposing only after settle (disposing a pending Cap'n Web promise cancels the call).
- The Stream DO drops the connection on a failed delivery and re-dials (outbound); the subscriber re-handshakes from its durable checkpoint, and the existing generation gate drops stragglers.
- Dropped the `Object.hasOwn` guard that prevented `onRpcBroken` from ever being wired for Cap'n Web proxy stubs (they expose no own descriptors). The wiring is defensive because property access on a native Workers RPC stub can fabricate a pipelined method that only fails at call time.

## 2. Agent processor: scheduled LLM turn dies with the DO and wedges the stream

`agent/llm-request-scheduled` is durable, but the debounce timer lived only in memory. After eviction+rehydration nothing re-armed it — and `#resetScheduledLlmRequestTimer` no-ops when there is no in-memory `scheduledEvent`, so **every later default-policy input was silently swallowed too**. Streams evicted between `llm-request-scheduled` and `llm-request-requested` were permanently wedged.

- The scheduled `currentRequest` state now records `scheduledOffset` (optional, so old checkpoints still parse).
- `snapshot()` — called by the host on every subscription (re-)handshake, even with no new events — and `processEventBatch` re-arm the lost timer; a history-read fallback covers checkpoints written before `scheduledOffset` existed, so already-wedged prod streams recover on their next re-handshake after this deploys.
- The handoff append retries instead of dropping the turn. All recovery is idempotent: `llm-request-requested` is keyed off the scheduled event's offset.

## Regression tests (all verified to fail on the unfixed code)

- `packages/streams/src/workers/durable-objects/stream-redial.workers.test.ts` — end-to-end in workerd: Stream DO → Callable dial → `StreamProcessorRunner` (echo), then `ctx.abort()` the runner to simulate a deploy and assert the next append still delivers and the subscriber's durable checkpoint advances. (The Stream-side cursor advances even into a dead stub — that *is* the bug — so the assertions are on actual delivery and the subscriber checkpoint.)
- `packages/streams/src/workers/rpc-lifecycle.test.ts` — proxy-shaped `onRpcBroken` wiring, pipelined-fake survival, delivery rejection reporting, settle-then-dispose.
- `apps/os/.../agent/implementation.test.ts` — four recovery tests: re-arm via `snapshot()` rehydration, re-arm via batch replay, history fallback for old checkpoints, handoff append retry.

`pnpm typecheck && pnpm lint && pnpm format && pnpm test` green, plus the streams workers suite.

---

> [!NOTE]
> **High Risk**
> Changes core stream delivery and agent LLM scheduling paths that run on every deploy/eviction; mistakes could duplicate LLM requests or drop events, though idempotency keys and generation gates mitigate double-handoff.
> 
> **Overview**
> Fixes a production outage where deploy eviction left Slack agents permanently silent by addressing **two independent failure modes** that composed together.
> 
> **Streams (M1):** Outbound batch delivery into a dead subscriber DO no longer leaves a zombie connection in `#connections`. `retainProcessEventBatch` now accepts `onDeliveryError`, observes rejections/sync throws, and disposes RPC results only after settle; `onRpcBroken` wiring no longer uses the broken `Object.hasOwn` guard. The Stream DO closes the connection and calls `#reconcile()` on delivery failure so subscribers are re-dialed and re-handshake from durable checkpoints.
> 
> **Agent processor:** Scheduled LLM turns persist `scheduledOffset` on `currentRequest` so a rehydrated DO can rebuild the in-memory debounce timer. Recovery runs on `snapshot()` (subscription re-handshake) and after `processEventBatch` replay; old checkpoints without `scheduledOffset` fall back to stream history. Failed `llm-request-requested` appends re-arm the timer instead of wedging the stream.
> 
> Regression coverage: `stream-redial.workers.test.ts`, `rpc-lifecycle.test.ts`, and agent `implementation.test.ts` scheduled-recovery suite.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 714b21a3d9cc29ba49216e83038e581631e4403b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Environment Config Lease

### OS
Status: released
Commit: `714b21a`
[Workflow run](https://github.com/iterate/iterate/actions/runs/27297848885)
Updated: 2026-06-10T19:11:51.700Z

### DIFF
--- a/apps/os/src/domains/agents/stream-processors/agent/contract.ts
+++ b/apps/os/src/domains/agents/stream-processors/agent/contract.ts
@@ -49,7 +49,16 @@ export const AgentProcessorContract = defineProcessorContract({
       .default({ model: DEFAULT_WORKERS_AI_AGENT_MODEL, runOpts: {}, debounceMs: 1000 }),
     currentRequest: z
       .discriminatedUnion("phase", [
-        z.object({ phase: z.literal("scheduled"), requestId: z.string() }),
+        z.object({
+          phase: z.literal("scheduled"),
+          requestId: z.string(),
+          // Offset of the llm-request-scheduled event. Lets a rehydrated
+          // processor instance re-arm the debounce timer (the timer dies with
+          // the Durable Object; this is the durable anchor it is rebuilt
+          // from). Optional so checkpoints written before this field existed
+          // still parse.
+          scheduledOffset: z.number().int().positive().optional(),
+        }),
         z.object({ phase: z.literal("requested"), llmRequestId: z.number().int().positive() }),
       ])
       .nullable()
@@ -270,7 +279,11 @@ export function reduceAgentEvent(args: { state: AgentState; event: AgentConsumed
     case "events.iterate.com/agent/llm-request-scheduled":
       return {
         ...state,
-        currentRequest: { phase: "scheduled" as const, requestId: event.payload.requestId },
+        currentRequest: {
+          phase: "scheduled" as const,
+          requestId: event.payload.requestId,
+          scheduledOffset: event.offset,
+        },
         pendingTriggerCount: 0,
       };
     case "events.iterate.com/agent/llm-request-requested":
--- a/apps/os/src/domains/agents/stream-processors/agent/implementation.test.ts
+++ b/apps/os/src/domains/agents/stream-processors/agent/implementation.test.ts
@@ -282,6 +282,191 @@ describe("AgentProcessor", () => {
     expect(payload.content).toContain("offset 44");
   });
 
+  // Regression tests for the 2026-06-10 prod incident: a deploy evicted the
+  // agent's Durable Object between `llm-request-scheduled` committing and the
+  // in-memory debounce timer firing. The rehydrated instance never re-armed
+  // the timer, so the turn never fired — and default-policy follow-up inputs
+  // only tried to reset the nonexistent timer, wedging the stream forever.
+  describe("scheduled request recovery after rehydration", () => {
+    it("re-arms the debounce timer from a rehydrated checkpoint on snapshot()", async () => {
+      vi.useFakeTimers();
+      const { stream, appended } = memoryStream();
+      const scheduled = agentEvent({
+        type: "events.iterate.com/agent/llm-request-scheduled",
+        payload: { requestId: "req_lost", debounceMs: 1000, model: "test-model" },
+        offset: 7,
+      });
+      const triggeringInput = agentEvent({
+        type: "events.iterate.com/agent/input-added",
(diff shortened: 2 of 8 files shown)


jonastemplestein:
folding into 1460

jonastemplestein:
Absorbed into #1460 (commit 76e501af4), adapted to that branch's structure:

- **M1 kernel fix ported wholesale**: `rpc-lifecycle.ts` (onDeliveryError + dropped Object.hasOwn guard) and its unit tests verbatim; the Stream-DO-side drop-and-redial now lives in `CoreStreamProcessor.openConnection` (connection management moved there on that branch) and additionally appends a `subscriber-disconnected` presence fact with reason `delivery-failed`. The end-to-end abort/re-dial workerd regression tests ported verbatim and pass — they also caught a missing hunk during the port, so they bite.
- **Agent scheduler**: #1460 already had the durable scheduled-offset + recovery (triggered by the `subscriber-connected` presence fact every re-handshake appends, rather than snapshot()/processEventBatch hooks). Adopted from here: `scheduledOffset` being optional with the history-read fallback (so pre-fix prod checkpoints parse and recover), and the handoff append retry. Your recovery scenarios are covered by three new connected-driven tests.
- `tasks/streams-review-fixes.md` M1 checkboxes carried over.

PR #1479, opened 2026-06-10
streams review: Stages 4–7 (browser robustness, perf, dead code, docs)

jonastemplestein:
## Streams review — Stages 4–7 (browser robustness, perf, dead code, docs)

The final batch of the `packages/streams` adversarial review (`tasks/streams-review-fixes.md`), bundled into one PR per request. Stages 0/1/3 already shipped (#1455, #1458, #1459); **Stage 2 (lazy init) was abandoned** — #1460's subscriber-presence model appends a `subscriber-connected` fact on every `subscribe()`, which fundamentally conflicts with "don't initialize storage until the first `append()`". `created`/`woken` stay eager.

All findings were re-verified against current `main` after #1460/#1457 moved the codebase; items those PRs already fixed or made obsolete are called out below.

### Stage 4 — browser-runtime correctness
- **C1-browser** — the server delivery pump is fire-and-forget for inbound subscribers and never reports a failed `ingest`, so a browser mirror that fails to apply a batch silently desyncs forever. The browser now self-heals: `ingestWithSelfHeal` resubscribes from the persisted checkpoint with bounded exponential backoff on ingest failure.
- **B1** — connection-epoch guard so a stale connection's late status callback can't clobber the live connection.
- **B2** — `appendBatch`/`runtimeState` await connection readiness (`whenStreamReady`) instead of throwing "disposed" during a transient reconnect; only throw when actually disposed.
- **B3** — the `ROLLBACK` in `stream-db.worker.ts` `batch()` is now in its own try/catch so it can't mask the original error (which `withBusyRetry`/`isBusyError` need to see).
- **B4** — reconcile guards on stream incarnation (server `createdAt` + a `mirror_meta` table): rebuild the mirror on incarnation change rather than trusting the offset comparison after a `reset()`.
- **B5** — `AbortSignal` on the Web Lock request (released on disposal) + the request rejection is surfaced instead of `void`-swallowed.
- **B6** — query lifecycle: arm GC at query creation (not only on unsubscribe), skip listenerless queries in `#onChange`, and equality-check before notifying to avoid spurious `useSyncExternalStore` churn.

### Stage 5 — performance
- **P1** — `browser-event-feed` O(n²) write amplification: coalesce to one op per `local_index` (was a cumulative op per event, each re-serializing the whole accumulated array).
- **P2** — bound group rows with `MAX_GROUP_EVENTS = 200` so one dominant event type can't grow a single blob unboundedly.
- **P3** — drop the per-event full-state `stateSchema.parse` in core `reduce`; state is validated only at the KV/recovery trust boundary now.
- **P4** — already fixed by #1460 (subscribe override forwards `eventTypes`; filtering is server-side). No change.

### Stage 6 — dead code / elegance (~247 lines deleted)
- **E1** dead exports in `shared/stream-processors.ts`; **E2** `waitForOpen`; **E3** unreachable circuit-breaker `paused` guard; **E5** dead `waitForEvent`/`messageInbox.error` machinery in `subscription.ts`; **E6** redundant circuit-breaker `consumes` entries.
- **E4 obsolete** (#1460 removed `processor-registered`; the shared `circuit-breaker-types` is a live, separately-imported hierarchy, not dead code). **E7 declined** — post-#1460 the subscribe override genuinely needs to retain the client callback / wire `onRpcBroken`, so folding it into the generic `makeRpcTargetClass` isn't worth it.

### Stage 7 — docs
- **D1** `design.md`: status banner, fixed 1-based offsets + core-event examples, the callable subscriber spec, SQLite/512KB-chunking storage, live-tail `replayAfterOffset` default, OPFS mirror; superseded banners on the never-shipped `implementProcessor`/`connectStream` sections.
- **D2** `README.md`: real browser export (`withStreamConnectionFromBrowser`), sync `Disposable`, `?path=` route map; new "Append & subscription semantics" section (D5).
- **D3** ADR 0001 superseded with the shipped per-`(namespace, path, slug)` singleton model.
- **D4** remaining comment drift (`beforeAppend`→`validateAppend`, `afterAppendBatch`→`processEventBatch`).

### Testing
`pnpm typecheck`, `pnpm lint`, node (79) + workers (15) suites, and example-app typecheck all green. Example-app Playwright e2e run locally against a clean `vite dev` (real Stream DO via Miniflare): **26 passed**. `origin/main` merged in (no `packages/streams` overlap; the new project-config stream-processor consumers don't touch any deleted exports).

---

> [!NOTE]
> **Medium Risk**
> Touches browser reconnect, mirror discard, and ingest self-heal paths that affect live viewers and local SQLite mirrors; server stream DO behavior is mostly unchanged aside from core reduce hot-path optimization.
> 
> **Overview**
> Closes out **Stages 4–7** of the `packages/streams` adversarial review: browser mirror reliability, feed/core hot-path perf, dead-code removal, and doc alignment. Stage 2 (lazy stream init) stays **abandoned** and is only noted in `tasks/streams-review-fixes.md`.
> 
> **Browser runtime** (`stream-browser-store.ts` and friends) is reworked so the OPFS mirror can recover instead of silently drifting. Inbound delivery is fire-and-forget, so failed `ingest` now triggers **`ingestWithSelfHeal`** (resubscribe from the last checkpoint with capped exponential backoff). Reconnects go through one **`scheduleReconnect`** path with a **connection epoch** so stale WebSocket callbacks cannot tear down the replacement connection. **`appendBatch` / `runtimeState`** use **`callWhenReady`** / **`whenStreamReady`** during transient reconnects instead of throwing “disposed”. Mirror trust uses server **`createdAt`** as incarnation identity via new **`mirror_meta`** helpers in **`stream-browser-db.ts`**. Web Lock **`release()`** aborts pending lock requests; the SQLite worker preserves the original batch error if `ROLLBACK` fails; reactive queries GC earlier, skip unobserved refreshes, and avoid notify when snapshots are unchanged.
> 
> **Performance:** `browser-event-feed` **`planFeedOps`** coalesces one SQL op per group row and caps groups at **`MAX_GROUP_EVENTS` (200)**. Core **`reduce`** no longer exit-parses the full `stateSchema` on every appended event.
> 
> **Cleanup:** ~247 lines removed (unused `stream-processors` exports, `waitForOpen`, dead subscription `waitForEvent` machinery, trimmed circuit-breaker `consumes` and an unreachable guard). **Docs:** `README` (real browser API, `?path=` routes, append/subscription semantics), `design.md` and ADR 0001 marked superseded where they diverge from shipped code.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f52f305a9f1ab43c9749f80e30bcd96705cbce59. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Environment Config Lease

### OS
Status: released
Commit: `f52f305`
Preview: https://os.iterate-preview-3.com
[Workflow run](https://github.com/iterate/iterate/actions/runs/27309066809)
Updated: 2026-06-10T22:05:11.534Z


jonastemplestein:
@/tmp/pr-comment.md

jonastemplestein:
@/tmp/pr-comment2.md

PR #1481, opened 2026-06-10
agent: recover triggering inputs skipped by the side-effect anchor

jonastemplestein:
## What was broken

Slack agents in prd receive messages but never reply — no LLM request is ever made. Observed live on 2026-06-10 in `iterate` project stream `/agents/slack/c08r1smtzgd/ts-1781124999-011519`:

- the slack-agent processor rendered the webhook into a triggering `agent/input-added` at **offset 9** (20:56:46.2)
- the agent processor's `subscription-configured` event landed at **offset 15** (20:56:46.7) — the AGENT DO wake hook appends it after D1 reads and workspace setup, so slack-agent reliably wins this race on a cold thread
- the host anchors side effects at the subscription-configured offset (`stream-processor-host.ts`), so the input at offset 9 was reduced as historical replay and its scheduling side effect was skipped: no `llm-request-scheduled`, no `llm-request-requested`, no `openai-ws` activity, no reply — and nothing ever retriggers it
- visible fingerprint in the stream: capability-noted renders exist only for offsets above the anchor (18–23), none for 8–9

The anchor mechanism is correct for re-attach (don't re-fire historical LLM requests), but it shipped in #1402 without anything making the *first* message of a new thread durable. Regression from #1402, same symptom as #1372 but a different mechanism. **Every first message of every new prod Slack thread is dropped.**

## The fix

Make the trigger a durable obligation in reduced state instead of a fire-and-forget side effect:

- **`AgentState.pendingTriggerOffset`** — set by a triggering `input-added`, cleared by `llm-request-scheduled` / `llm-request-requested` / `llm-request-queued`. If it survives in reduced state, the scheduling side effect never ran.
- **`subscriber-connected` reconciliation recovers it** (the presence fact always lands above the anchor, so this handler always runs live): schedule a request when idle, append the queued fact when a request is in flight (never interrupts in-flight work). Appends are keyed off the trigger event exactly like the live path (`agent/llm-request-scheduled@<offset>`), so raced duplicates dedup in the stream.
- **Gated on `pendingTriggerOffset <= sideEffectsAfterOffset`** so recovery fires only for anchor-skipped triggers and never races the live `input-added` handler. Crash/restart cases above the anchor remain owned by the existing scheduled-phase reconciliation.
- `StreamProcessor.processEvent` args now expose `sideEffectsAfterOffset` (the batch-level hook already had it); the core processor's inline path passes 0 (inline appends are always live).

The scheduled phase needs no queued fact on recovery: its handoff rebuilds the request body from full committed history, which already includes the skipped trigger.

## Verification

- Unit tests replay the prod stream shape: trigger below anchor + subscriber-connected above → exactly one `llm-request-scheduled@9`; non-triggering inputs don't recover; in-flight requests get a queued fact; live triggers aren't double-scheduled.
- New token-gated e2e (`schedules and completes an LLM request for a plain routed Slack message`) drives a real Slack root message + routed webhook through webhook → input → scheduled → requested → completed(success) against a live deployment.
- E2E run against the preview deployment with the real Slack bot token: results to follow in a comment.

---

> [!NOTE]
> **High Risk**
> Changes core agent LLM scheduling and subscriber-connected reconciliation on a production outage path; incorrect gating could double-schedule or miss triggers on every new Slack thread.
> 
> **Overview**
> Fixes **first-message silence** on new Slack thread streams when a triggering `input-added` lands **before** the agent subscription is configured: the host’s side-effect anchor replays that input into state but skips scheduling, so no LLM turn ever starts.
> 
> **Agent processor** now records **`pendingTriggerOffset`** in reduced state for triggering inputs and clears it when a durable schedule/request/queue fact exists. On **`subscriber-connected`**, when that offset is at or below the anchor, it **recovers** the missed obligation—`llm-request-scheduled` when idle (same idempotency key as the live path) or **`llm-request-queued`** when a request is already in flight—without double-scheduling live triggers above the anchor. **`#appendLlmRequestScheduled`** arms the debounce timer with the **committed** `requestId` after idempotent dedup so raced recovery paths don’t wedge the handoff.
> 
> **Streams**: `processEvent` receives **`sideEffectsAfterOffset`** so reconcilers can detect anchor-skipped side effects; the core inline path passes **`0`** (always live).
> 
> **Verification**: new unit coverage for anchor-skip recovery, deduped schedule, queue-when-busy, and no recovery for non-triggering inputs; token-gated e2e asserts routed Slack webhook → scheduled → requested → completed(success).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eb3a7ac0bb17f468c1d5490f0b6951bfe612374e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Environment Config Lease

### OS
Status: released
Commit: `eb3a7ac`
[Workflow run](https://github.com/iterate/iterate/actions/runs/27308869802)
Updated: 2026-06-10T22:04:58.830Z

### DIFF
--- a/apps/os/e2e/vitest/agents.e2e.test.ts
+++ b/apps/os/e2e/vitest/agents.e2e.test.ts
@@ -898,6 +898,97 @@ itIfSlackBotToken(
   180_000,
 );
 
+itIfSlackBotToken(
+  "schedules and completes an LLM request for a plain routed Slack message",
+  async () => {
+    // Regression test for the 2026-06-10 prod outage: on a freshly
+    // bootstrapped Slack thread stream, the slack-agent processor rendered
+    // the webhook into a triggering agent input before the agent processor's
+    // subscription was configured. The host anchored side effects at the
+    // subscription-configured offset, so the trigger was replayed as
+    // historical and no LLM request was ever scheduled — the agent never
+    // replied. The subscriber-connected reconciliation must recover the
+    // skipped trigger, so the LLM turn completes no matter which side wins
+    // the bootstrap race.
+    await using fixture = await createTestProjectFixture({ slugPrefix: "slack-agent-llm" });
+    const { client, project } = fixture;
+    const suffix = uniqueSuffix();
+    const slackChannelId = await requireSlackChannelId();
+    const rootText = `OS slack-agent llm trigger proof ${suffix}`;
+    const rootMessage = await postSlackMessage({
+      channel: slackChannelId,
+      text: rootText,
+      token: requireSlackToken(),
+    });
+    const routedAgentPath = slackAgentPath({
+      channel: slackChannelId,
+      threadTs: rootMessage.ts,
+    });
+
+    await client.project.streams.append({
+      projectSlugOrId: project.id,
+      streamPath: "/integrations/slack",
+      event: slackProcessorSubscriptionEvent({ projectId: project.id, suffix }),
+    });
+
+    await client.project.streams.append({
+      projectSlugOrId: project.id,
+      streamPath: "/integrations/slack",
+      event: {
+        type: "events.iterate.com/slack/webhook-received",
+        idempotencyKey: `slack-agent-e2e-llm-webhook:${suffix}`,
+        payload: {
+          slackTeamId: "T_E2E",
+          body: {
+            type: "event_callback",
+            team_id: "T_E2E",
+            event_id: `EvLlm${suffix}`,
+            event: {
+              type: "message",
+              channel: slackChannelId,
+              channel_type: "channel",
+              user: "U_E2E",
+              ts: rootMessage.ts,
+              event_ts: rootMessage.ts,
+              text: `please acknowledge: ${rootText}`,
+            },
+          },
+        },
+      },
+    });
+
+    const events = await readUntil({
+      agentPath: routedAgentPath,
+      client,
+      projectId: project.id,
+      afterOffset: "start",
+      predicate: (event) =>
+        event.type === "events.iterate.com/agent/llm-request-completed" &&
+        (event.payload as { result?: { status?: unknown } }).result?.status === "success",
+      timeoutMs: 150_000,
+    });
+
+    // The durable handoff chain the regression silently dropped: trigger →
+    // scheduled (keyed off the triggering input, so live path and recovery
+    // converge) → requested → completed.
+    expect(events).toContainEqual(
+      expect.objectContaining({
+        type: "events.iterate.com/agent/llm-request-scheduled",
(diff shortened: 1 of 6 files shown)


jonastemplestein:
**E2E verification record** (run against the leased `preview-4` deployment of eb3a7ac0b, with the real Slack bot token from Doppler `preview_4`):

- ✅ `schedules and completes an LLM request for a plain routed Slack message` — the new regression e2e — passed in 16.7s: real root message posted to `#slack-agent-e2e-test`, routed webhook, `agent/input-added` → `llm-request-scheduled` → `llm-request-requested` → `llm-request-completed(success)`, no `stream/error-occurred`.
- ✅ `routes Slack webhooks into slack-agent streams and executes bang command replies`
- ✅ `lets a real agent conversation post to Slack through codemode`
- Full `agents.e2e.test.ts` run: 8/10 passed. The 2 failures (`recovers and still replies when the agent host DO is killed mid-turn` timeout; `project config worker customizes fresh agents` git-push `ok:false`) coincided exactly with the post-merge **Preview / cleanup destroying the preview worker mid-suite** (preview-4 now 522s; isolated retries fail at `projects.create` with `MALFORMED_ORPC_ERROR_RESPONSE`). Not code regressions — the kill-recovery branch is condition-for-condition identical and covered by unit tests.

Prd smoke test on the real `iterate` project to follow after the main deploy.

jonastemplestein:
**Prd smoke test — passed** ✅

Verified on the real `iterate` project (prj_d871bac9722d45aba4e3dbb50057900d) after the main deploy, in the exact thread from the original incident (`#test-blank`, root ts `1781124999.011519`):

- Injected a human-shaped `slack/webhook-received` (thread reply mentioning the bot) into `/integrations/slack`; the real prd pipeline routed it to the existing thread stream.
- The chain that was silently missing during the outage now fires end to end: `agent/input-added@42` → **`agent/llm-request-scheduled@42`** → `llm-request-requested@47` → `openai-ws/llm-request-started` → `agent/llm-request-completed (success)` → `itx/execution-completed { ok: true, durationMs: 7077 }`.
- The agent posted a real reply in the real Slack thread with the project's bot token: *"Acknowledged — smoke test reply received in this thread."* (`1781129770.658629`), confirmed via `conversations.replies` using the Doppler prd Slack token.

Note: a bot-authored root message (posted with the Doppler CI bot token) correctly does **not** trigger the agent — `slack-agent` skips bot messages at input render. The thread streams still bootstrap with both subscription sets in either race order (verified in the fresh stream `ts-1781129521-178669`: agent subscriptions at offsets 13–15, input at 26).

PR #1580, opened 2026-06-30
Expose Workers AI through minimal ITX

jonastemplestein:
## Summary

- expose the Workers AI binding as project/agent ITX capability `ai`
- include `ai` in project descriptions as a built-in capability
- document the minimal web-chat LLM agent shape and AI provider boundary

## Validation

- `pnpm exec oxlint . --threads 1 --deny-warnings`
- `pnpm typecheck`
- `pnpm knip`
- `pnpm --dir apps/minimal-itx-v4 e2e itx.e2e.test.ts -t "Project describe exposes Workers AI"`

### DIFF
--- a/apps/minimal-itx-v4/src/rpc-targets.ts
+++ b/apps/minimal-itx-v4/src/rpc-targets.ts
@@ -34,6 +34,7 @@ import type {
   Agent,
   AgentCollection,
   AgentItx,
+  Ai,
   CfExecutionContext,
   ItxAuth,
   ItxRoot,
@@ -57,6 +58,11 @@ import type {
   DynamicWorkerCollection,
   DynamicWorkerRef,
 } from "./types.ts";
+
+type AiBinding = {
+  models?: () => unknown;
+  run(model: string, body: unknown): Promise<unknown>;
+};
 
 export class StreamRpcTarget extends RpcTarget implements Stream {
   constructor(readonly props: { auth: ItxAuth; projectId: string | null; path: string }) {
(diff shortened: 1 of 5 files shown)


apps/minimal-itx-v4/src/rpc-targets.ts
```diff
@@ -347,6 +353,23 @@ class SecretRpcTarget extends RpcTarget implements Secret {
   }
 }
 
+class AiRpcTarget extends RpcTarget implements Ai {
+  constructor(readonly binding: AiBinding) {
+    super();
+  }
+
+  models() {
+    if (typeof this.binding.models !== "function") {
```
jonastemplestein, line 362: dumb

apps/minimal-itx-v4/src/rpc-targets.ts
```diff
@@ -347,6 +353,23 @@ class SecretRpcTarget extends RpcTarget implements Secret {
   }
 }
 
+class AiRpcTarget extends RpcTarget implements Ai {
```
jonastemplestein, line 356: this is way over-engineered. just use env from import { env } from "cloudflare:workers" - it's fine - doesn't need injecting here. in fact, you can structure this more similar to apps/os where we just literally proxy through to env.AI

jonastemplestein: though we do want the constructor to take some arguments that are used to configure the AI gateway that should be used

apps/minimal-itx-v4/src/rpc-targets.ts
```diff
@@ -58,6 +59,11 @@ import type {
   DynamicWorkerRef,
 } from "./types.ts";
 
+type AiBinding = {
```
jonastemplestein, line 62: don't want

jonastemplestein:
Closing this because #1581 subsumes it. I addressed the review feedback from this PR on #1581: AiRpcTarget now proxies directly to env.AI, removes the injected AiBinding type, and keeps constructor gateway configuration for env.AI.run options.

PR #1602, opened 2026-07-02
Streams domain: narrative Stream DO, isolated connections, contract/runtime split

jonastemplestein:
Thermo-nuclear code-quality pass over `apps/os/src/domains/streams` (in the spirit of [cursor's review skill](https://github.com/cursor/plugins/blob/main/cursor-team-kit/skills/thermo-nuclear-code-quality-review/SKILL.md)): same storage/event/wire formats, dramatically clearer structure, plus one deliberate API unification (below). Every event payload, KV key, subscription key, and error message is byte-identical to before; the only public-surface change is that `subscribeConfigured` is folded into `subscribe`.

## What was wrong

- `stream-durable-object.ts` (1,678 lines) interleaved four concerns: chunked SQLite storage, the stream's own fold, the append commit path, and live connection management (pump closures, RPC stub retention, idle timers, reconciliation) — you could not read the append path without wading through delivery plumbing.
- The stream's own "core" processor didn't rhyme with the 13 other processors in the codebase. Everyone else is `*-processor-contract.ts` + a class with `reduce`/`processEvent`; the core was a contract file plus private methods with unrelated names scattered through the DO.
- Subscription/connection logic had no home: a 160-line closure factory (`#startSubscription`), retention helpers in a separate `rpc-lifecycle.ts`, idle timers and reconcile loops interleaved with storage code.
- Two subscription verbs (`subscribe` / `subscribeConfigured`) meant two code paths into the Stream DO that differed only in admission rules.
- `stream-processor.ts` (1,406 lines) was a type-machinery library and a runtime base class in one file, with some dead API surface and a stale reference to a deleted lint rule.
- Every DO callsite passed the processor slug twice: `host.add(RepoProcessorContract.slug, (deps) => new RepoProcessor(deps))`.
- The deepest structural idea in the domain — that waking a stream subscriber and taking its `processEventBatch` callback **is** an itx live-capability provide — was documented nowhere.
- Duplicated schemas: subscription-configured and rule-configured payloads were each defined twice (event catalog + reduced-state records), free to drift.
- No domain README; `CONTEXT.md` documented an API (`withStreamProcessor`) that hasn't existed for months.

## Architecture after

```
domains/streams/
  README.md                    the narrative: model, connections, wake handshake, file map
  stream-durable-object.ts     THE STREAM — append commit point, core processor
                               (validateAppend/reduce/processEvent), checkpoint, policy
  core-processor-contract.ts   core contract: state schema + events.iterate.com/stream/* catalog
  stream-storage.ts            NEW — chunked SQLite event log (2 MB cell limit → JS chunking)
  stream-connections.ts        NEW — connection table, delivery pump, RPC stub retention
                               (absorbs rpc-lifecycle.ts), idle teardown
  processor-contracts.ts       NEW (split) — defineProcessorContract + event-type resolution machinery
  stream-processor.ts          the StreamProcessor base class (runtime only)
  stream-processor-host.ts     hosts processors in a DO; subscriber half of the wake handshake
  schemas.ts / utils.ts        unchanged surface
```

### 1. The Stream DO is now a narrative

`stream-durable-object.ts` reads top-to-bottom in the order things happen:

1. **Append: the commit point** — validate → assign offsets → reduce → persist, one await-free turn; then post-commit fan-out.
2. **The core processor** — explicitly structured to rhyme with every hosted processor:

   | | hosted processor | core processor |
   | --- | --- | --- |
   | contract/schemas | `*-processor-contract.ts` | `core-processor-contract.ts` |
   | pure fold | `reduce` | `#reduce` |
   | side effects | `processEvent` | `#processEvent` |
   | pre-commit gate | — impossible | `#validateAppend` |

   The core runs inline in the append turn instead of behind a subscription, which grants exactly the two powers no hosted processor can have: it is synchronous with the commit, and `#validateAppend` can **reject an event before it becomes a durable fact** (pause door, configured-subscriber target validation, cross-post rule scoping). Per discussion, there is deliberately **no separate core implementation file**: the contract stays the schema layer and the DO owns the fold — that's what makes the whole thing readable in one sitting.

3. **Checkpoint** — KV state read/write/catch-up/recovery, one section.
4. **Subscriptions** — policy only (who may subscribe, which facts to append); mechanics live in `stream-connections.ts`.
5. **Operator verbs** — `runIdleTeardownNow` / `reset` / `kill`.

### 2. One subscription entry point (public-surface change)

`subscribeConfigured` is gone from the Stream DO, `StreamRpcTarget`, and the host. **`Stream.subscribe` is now the only way a delivery connection is opened**; the wake-handshake response is the same verb with an intent flag:

```ts
// the host answering a wake — previously stream.subscribeConfigured({...})
const handle = await options.stream.subscribe({
  subscriptionKey,        // the durable key the stream woke us with
  configured: true,       // open the configured subscription, not an ephemeral one
  replayAfterOffset: snapshot.offset,
  ...
});
```

All previous semantics are preserved exactly, they're just admission rules on one path now:

- `configured: true` requires trusted-internal auth (gated in `StreamRpcTarget`, the auth boundary — same place the old `subscribeConfigured` gate lived) and an existing `subscription-configured` fact for the key (same "not configured" error, so a stale wake still fails loudly instead of silently opening an ephemeral connection).
- Ephemeral subscribes still cannot squat on a configured key (same "reserved for a configured subscriber" error).
- The host consequently needs only the plain `Stream` capability — the `ConfiguredStream` extension type is deleted.

### 3. Connections are isolated behind a policy/mechanism seam

`StreamConnections` owns the connection table, the per-connection pump (catch-up replay → live batches), stub retention/disposal, and the idle-teardown timer. The Stream DO stays the policy owner and reaches in through five explicit hooks:

```ts
readonly #connections = new StreamConnections({
  idleTeardownMs: idleTeardownMs(this.env),
  hooks: {
    readEvents: (args) => this.getEvents(args),
    coreState: () => this.#coreProcessorState,
    appendConnectedFact: (payload) => this.#appendPresenceFact({ type: ".../subscriber-connected", payload }),
    appendDisconnectedFact: (payload) => this.#appendPresenceFact({ type: ".../subscriber-disconnected", payload }),
    onConfiguredConnectionLost: () => this.#reconcileConnections(),
  },
});
```

The load-bearing `open()` ordering (register → connected fact → wire broken-transport signals → wake pump, so the connected fact rides the tail of the first replay batch) is preserved verbatim and now documented in one place. `rpc-lifecycle.ts` is gone; the dup/dispose retention helpers live next to the only thing that uses them.

### 4. The wake handshake is documented as a live-capability provide

The rhyme with `domains/itx/live-capability.ts` is now stated at all three corners — the stream's `#wakeConfiguredSubscriber`, the host's file header, and the domain README:

1. Stream **wakes** the subscriber with serializable coordinates only (`wakeStreamSubscriber({ stream, subscriptionKey })`).
2. Subscriber answers with the one public `subscribe` verb (`configured: true` + its durable key), handing the stream a live `processEventBatch` **callback capability**.
3. Stream dups + retains the stub past the delivering RPC call, invokes it per committed batch, disposes on close — the same ownership rules as itx capability provision.

The host's remaining complexity is spent on exactly one problem — making that callback safe to re-issue (connection generations, checkpoint replay, poison-batch cutoff) — and its header now says so.

### 5. Host callsites

`add` derives the registry name from `processor.contract.slug` (which is what every callsite passed anyway), and the three "invalidate + reopen" paths (wake / ingest recovery / idle) share one `supersedeConnection`:

```ts
// before
readonly #repoProcessor = this.#host.add(
  RepoProcessorContract.slug,
  (deps) => new RepoProcessor({ ...deps, ... }),
);

// after
readonly #repoProcessor = this.#host.add((deps) => new RepoProcessor({ ...deps, ... }));
```

All five DOs (agent, project, repo, secret, itx) updated; 9 now-unneeded contract imports deleted. KV checkpoint keys are unchanged (`stream-processor:${slug}:snapshot` — the slug was already the registry name everywhere).

### 6. Contract machinery split from the runtime class

`processor-contracts.ts` is the "what is a contract" layer (`defineProcessorContract`, `buildEvent`, event-type→payload-schema resolution); `stream-processor.ts` is the "how batches run" layer (the `StreamProcessor` class). This mirrors the contract/implementation split every processor in the codebase already uses. The two `defineProcessorContract` overloads (with/without `processorDeps`) collapsed into one signature with a defaulted type parameter; `parseEvent`/`parseEventInput` share one runtime body.

### 7. Dead surface removed

- `StreamProcessor.buildEvent` (the protected class method) — zero callers. `contract.buildEvent(...)` is **kept** per review.
- `onClose` connection arg — never passed.
- Stale comment referencing the deleted `iterate/stream-processor-override-args` lint rule.

### 8. Schemas/docs

- `core-processor-contract.ts`: subscription-configured and rule-configured payloads defined **once** and shared between the event catalog and the reduced-state records (`latestConfiguredEvent(type, payload)` helper) — they previously existed twice each and could drift.
- Every newly introduced named type carries a docstring saying what it represents and why it exists.
- New `domains/streams/README.md` (the domain narrative + file map).
- `apps/os/src/README.md` streams section refreshed; `CONTEXT.md`'s fictional `withStreamProcessor` entry replaced with `createStreamProcessorHost`, and "Stream Processor waitUntil" renamed to the real `keepAliveWhile`.

## LOC budget

Domain server code (excluding tests):

| File | Before | After | Δ |
| --- | ---: | ---: | ---: |
| `stream-durable-object.ts` | 1,678 | 1,107 | **−571** |
| `stream-processor.ts` | 1,406 | 616 | −790 |
| `processor-contracts.ts` (split from ↑) | — | 631 | +631 |
| `stream-processor-host.ts` | 393 | 395 | +2 |
| `core-processor-contract.ts` | 356 | 360 | +4 |
| `rpc-lifecycle.ts` | 170 | — | −170 |
| `stream-connections.ts` (new) | — | 461 | +461 |
| `stream-storage.ts` (new) | — | 159 | +159 |
| `schemas.ts` / `utils.ts` | 128 | 128 | 0 |
| **Domain total** | **4,131** | **3,857** | **−274 (−6.6%)** |

Plus: −25 lines across the five DO callsites, +134 lines of README (docs, not code). The honest read: the raw win is modest because the original was dense with load-bearing invariants that were kept (and their comments largely preserved — several were *added* back per review). The real change is shape — the biggest file dropped 34%, every other file is ≤640 lines with one job, one whole public verb and ~100 lines of dead surface are gone.

## Decision & difficulty records

1. **Core processor lives in the DO, not a separate implementation file.** Considered a `core-processor.ts`; rejected per discussion — the contract file stays the schema layer, and keeping `#validateAppend`/`#reduce`/`#processEvent` in the DO is what makes the commit path readable end-to-end. The rhyme is carried by naming + the section banner + the README table.
2. **`subscribe` unification kept all three admission properties.** The design question was where the trusted-internal gate lives once the verbs merge: the DO can't see the caller's principal (DO methods have no auth), so the gate stays in `StreamRpcTarget` (the auth boundary, same as before) while the DO enforces the state-level rules (configured key must exist; ephemeral must not squat on one). An alternative — deriving "configured" purely from whether the key is registered, with no intent flag — was rejected because a stale wake (config removed mid-handshake) would then silently open an *ephemeral* connection that idle-teardown never severs, instead of failing loudly with "not configured" as it does today.
3. **`stream-durable-object.ts` is 1,107 lines, just over the 1,000-line review threshold.** Justification: ~35% is narrative documentation (which this refactor was asked to strengthen, including docstrings on every named type), and the remaining extraction candidates are core policy (validation, reduce cases, wake-target resolution) that was explicitly directed to stay. Extracting policy to duck a line count would have redistributed complexity, not removed it.
4. **File count went 8 → 9 (+README) despite "fewer files" being a goal.** `rpc-lifecycle.ts` was absorbed, but storage and connections were extracted and the contract machinery split out. Each split removes a real interleaving; a lower file count would have meant >1,000-line files. LOC went down; file *sizes* went way down.
5. **`defineProcessorContract` single signature.** The old with-deps/without-deps overload pair collapsed into one signature with `ProcessorDeps extends readonly unknown[] = readonly []`. TS cannot relate the generic helper-method shapes to the runtime `Object.assign` result (TS2394), so the implementation signature is untyped with an explicit comment — the standard typed-overload/untyped-impl pattern. All 13 existing contracts + the e2e-local contract infer identically (verified by typecheck across every reducer switch).
6. **Pump/retention code moved verbatim, not rewritten.** The delivery pump, `retainProcessEventBatch` (including the Cap'n Web `onRpcBroken` pipelined-fake defenses and the pay-per-observe `onDeliveryError` trade-off), and the host's generation gates are copied with their comments intact. These encode incident-derived behavior; this PR deliberately changes their *location and isolation*, not their logic.
7. **One subtle ordering kept and now named:** the exactly-one-event exclusion (`subscriber-disconnected`) from append-time configured-subscriber re-wake, which would otherwise undo every teardown. It kept its full explanatory comment in `append`.
8. **Behavior-identical wake-path change:** `wakeStreamSubscriber` previously did unsubscribe-then-bump-generation, recovery did bump-then-unsubscribe; both are now the shared `supersedeConnection` (bump-then-unsubscribe, with the idle path's try/catch). Equivalent because there is no await between the statements; the unification also makes a dead-handle `unsubscribe()` throw non-fatal on the wake path (strictly more robust).
9. **Restored per review:** `contract.buildEvent` (typed append-input builder on every contract) and `previousState` on core reduced events (keeps the core's `#processEvent` args shaped like hosted processors' `ReducedEvent`), both initially removed as "zero callers".
10. **`iterate(colocate-single-use-types)` fallout:** the lint rule pushed `ResolvedEventInput` next to its consumers — a nice example of the repo's lint rules steering the cleanup.

## Verification

- Rebased on latest `main` (includes #1601's saga-append batching and #1596's MCP client migration); clean at `origin/main` as of this push.
- `pnpm typecheck` / `pnpm lint` / `pnpm format` — clean.
- Unit: 216 passed (includes the agent/slack processor suites that drive `StreamProcessor` through a MemoryStream harness — debounce, idempotency, recovery, commit-before-side-effect ordering).
- Stream e2e vs local dev (`doppler run --config dev -- pnpm e2e`), re-run after the rebase **and** after the subscribe unification: `stream-lifecycle` + `streams` + `stream-security` — 17 passed, 1 skipped. Covers idle teardown, append-after-teardown re-wake from checkpoint, configured-target project validation (rejection **and** nothing committed), cross-post rules + recursion guard, replay/live/state-only subscribe, malformed-descriptor rejection, Cap'n Web session-drop cleanup.
- itx e2e (subscribe + processor filters): 6 passed — Cap'n Web callback retention through the stateless Worker proxy, same-key replacement, nested subscriber callbacks, processor install + child-event replay.
- Full `itx.e2e`: 35/41 passed locally; the 2 unique failures (dynamic-worker composition tests, `capnweb serialize "internal error"`) were **baselined against a clean `origin/main` worktree and fail there identically** — pre-existing local-dev-only failure mode, not a regression (details in PR comment).

---

> [!NOTE]
> **Medium Risk**
> Touches the synchronous append commit path and live subscription/RPC delivery for a core platform primitive, but the PR claims byte-identical persistence and was validated with broad stream/itx e2e coverage.
> 
> **Overview**
> Large **structure-only** pass on `apps/os/src/domains/streams`: the monolithic `stream-durable-object.ts` is reorganized into a top-to-bottom narrative (append → core processor → checkpoint → subscription policy) while **SQLite/KV formats, event payloads, and wire semantics stay the same**.
> 
> **New modules:** `stream-storage.ts` (chunked event log), `stream-connections.ts` (delivery pump, RPC stub retention, idle teardown — absorbs deleted `rpc-lifecycle.ts`), and `processor-contracts.ts` (`defineProcessorContract` + event-type machinery split out of `stream-processor.ts`). The stream “core” fold is renamed to rhyme with hosted processors (`#validateAppend` / `#reduce` / `#processEvent`).
> 
> **Public API change:** `subscribeConfigured` is removed; configured handshakes use **`subscribe({ subscriptionKey, configured: true, ... })`** (trusted-internal gate still on `StreamRpcTarget`). **`createStreamProcessorHost().add`** no longer takes a slug argument — registration key comes from **`processor.contract.slug`**. All domain DOs and contract imports are updated accordingly.
> 
> **Docs:** new `domains/streams/README.md`, refreshed `src/README.md` and `CONTEXT.md` (`createStreamProcessorHost`, `keepAliveWhile`). Core contract schemas dedupe subscription/rule payloads via shared helpers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7639ab7d9c506fed183734ae1acf8a55e9a57cb6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Environment Config Lease

<details>

| Auth | released | `7639ab7` | [https://auth.iterate-preview-1.com](https://auth.iterate-preview-1.com) | 27.4s | 447ms | 25.1s | [Workflow run](https://github.com/iterate/iterate/actions/runs/486108830995309) | 2026-07-03T07:40:27.358Z | Preview app released. |
| OS | released | `7639ab7` | [https://os.iterate-preview-1.com](https://os.iterate-preview-1.com) | 47.9s | 56.7s | 46.2s | [Workflow run](https://github.com/iterate/iterate/actions/runs/486108830995309) | 2026-07-03T07:40:48.485Z | Preview app released. |
| Semaphore | released | `6a87053` | [https://semaphore.iterate-preview-1.com](https://semaphore.iterate-preview-1.com) | 37.5s |  | 19.3s | [Workflow run](https://github.com/iterate/iterate/actions/runs/464490960076004) | 2026-07-03T07:40:21.539Z | Preview app released. |
| Streams Example App | released | `7639ab7` | [https://streams-example-app-preview-1.iterate-dev-preview.workers.dev](https://streams-example-app-preview-1.iterate-dev-preview.workers.dev) | 15.2s | 15.7s | 14.6s | [Workflow run](https://github.com/iterate/iterate/actions/runs/486108830995309) | 2026-07-03T07:40:16.853Z | Preview app released. |

</details>


jonastemplestein:
**Full itx.e2e baseline** — 2 of 41 tests fail locally on this branch ("Project repos, workers, runScript, and dynamic worker refs compose" and "repo worker source projection is cleared when the main worker file is deleted"), both with `capnweb serialize.ts:658 — Error: internal error; reference = …`. I ran the same two tests against a clean `origin/main` worktree with its own dev server: **both fail identically on main**. Pre-existing local-dev-only failure mode (the known capnweb-vs-vite-dev issue), not a regression from this PR. The other 35 itx e2e tests plus all 17 stream e2e tests pass on this branch.

**Side-finding while baselining, unrelated to this PR:** a *fresh* `pnpm install` of current main cannot start `pnpm dev` — alchemy@0.83.3's inner miniflare dep now resolves to 4.20260617.0, so alchemy emits compatibility date `2026-06-17` into the generated dev worker configs, but `@cloudflare/vite-plugin@1.33.2`'s pinned workerd only supports up to `2026-05-01` → `MiniflareCoreError: This Worker requires compatibility date "2026-06-17"`. Existing worktrees only work because they carry a stale alchemy→miniflare@4.20260424.0 link that a reinstall destroys (mine did, mid-verification). Workaround: repoint the six `node_modules/.pnpm/alchemy@0.83.3*/node_modules/miniflare` symlinks to `../../miniflare@4.20260424.0/node_modules/miniflare`. Proper fix is probably pinning alchemy's compat date or its miniflare resolution — worth a separate issue/PR.

jonastemplestein:
Update on the "known local-dev flake" from the earlier comment: it wasn't capnweb at all. Root cause found and fixed in #1614 — local dev runs all eight itx workers in one workerd whose worker-loader cache is shared across parent workers, so the api worker was being handed dynamic-worker isolates created by the stream/project workers (with their foreign `ctx.exports` stubs baked into the env), and every invocation failed with a redacted internal error. With #1614 both previously-failing dynamic-worker tests pass locally.

Two further local-only failures remain (`TypeError: The RPC receiver does not implement the method "snapshot"` in admin-project + the egress-secrets test) — baselined against clean origin/main: they fail there identically, so also pre-existing and unrelated to this PR. Tracked in the #1614 description.

PR #1654, opened 2026-07-04
Per-agent sandboxes: itx.sandbox capability + project-confined egress (MITM)

jonastemplestein:
## What

Give every `/agents` agent its own Cloudflare Sandbox — a real Linux container that is durable, restorable, sleeps when idle, and whose egress is confined to the project. It shows up in an agent's scope as `itx.sandbox`.

```js
async (itx) => {
  const r = await itx.sandbox.exec("bun test");   // YOUR container, boots on first use
  return { exitCode: r.exitCode, out: r.stdout };
}
```

## How

**`itx.sandbox` is a provided capability, not a built-in.** Like `itx.agent`/`itx.chat` are derived from the scope path, but mounted the way capabilities are mounted: the agent birth certificate provides a durable itx-expression on the agent's own capability host —

```ts
{ path: ["sandbox"], type: "itx-expression",
  expression: ["sandboxes", ["get", <agent path>]] }
```

so every `itx.sandbox.<method>(...)` re-evaluates `itx.sandboxes.get(<the agent's own path>)` at call time and dispatches through the capability host like any other mount. The mount replays with the stream; no built-in surface, no durable state on the itx object.

**Any non-root path can be a sandbox.** Sandboxes live in their own Durable Object namespace, so a sandbox path never collides with the stream/agent/secret at the same path — it *names* them. `normalizeCloudflareSandboxPath` → `normalizeSandboxPath`; the `/sandboxes/cloudflare/` prefix requirement is dropped. An agent's sandbox is the sandbox at the agent's own `/agents/...` path; standalone sandboxes still live under `/sandboxes/cloudflare/...` by convention.

**Durable / restorable / idle-sleep are the SDK's defaults, unchanged.** Getting the stub is cheap (no container); the first command boots it; the SDK's durable `sleepAfter` idle alarm stops it (`3m`, from main). Durable Object storage + identity survive sleep; the container filesystem is re-provisioned on every start by the existing project-repo clone — that's what makes it restorable rather than merely long-lived.

**All sandbox egress goes through the project egress fetcher, HTTPS included (MITM).** A sandbox has no direct internet path. Every outbound request — HTTP and, via `interceptHttps`, HTTPS — is intercepted by `@cloudflare/containers` and forwarded to the owning project's Durable Object, the same decision point dynamic workers' `globalOutbound` uses. So secret substitution, allow/deny, and audit apply to sandbox traffic exactly as to the rest of the project. HTTPS interception is a TLS man-in-the-middle using the Cloudflare container CA, which the stock `cloudflare/sandbox` image installs at start when `SANDBOX_INTERCEPT_HTTPS` is set (the SDK sets it from the flag) — no Dockerfile change needed.

Wiring is three points: `worker.ts` re-exports `ContainerProxy` (from `@cloudflare/sandbox`, so its build shares one outbound registry with the DO's `Container` base); the DO sets `interceptHttps = true` and registers a catch-all `outbound` handler that recovers the project from the container id and calls `projectStub(env.PROJECT, projectId).fetch(request)`.

> ⚠️ One subtle bug worth calling out for reviewers: the `outbound` handler **must** be registered in a `static {}` block, not `static outbound = …`. Under our `useDefineForClassFields` (ES2024) target a static field *defines an own data property* that shadows the base `Container`'s `outbound` *accessor*, so the registration setter never runs — egress then silently falls through to a direct fetch (TLS still MITM'd, but project policy bypassed). The assignment in a static block invokes the inherited setter. There's a comment on the block.

## Verification (real Cloudflare, not local OrbStack)

Deployed to preview-7 (real Firecracker containers) and verified end-to-end:

- **Cold boot ~1.1s**, warm exec ~112ms, kernel `6.18.36-cloudflare-firecracker x86_64`.
- **New e2e `sandbox-egress.e2e.test.ts`** (runs in the node preview lane; skips local dev) proves all three egress properties in one request made *from inside a container*:
  - **MITM** — the cert the container is handed for the echo host is issued by `CN = Cloudflare TLS proxy-everything Intercept CA`, not the origin's CA.
  - **Routed through the project egress fetcher** + **secret substitution** — the echo sees the real secret *material* where the container sent a `getSecret({ path })` placeholder; `secret.describe()` never exposes it and the usage audit increments. The container never holds the material.

`pnpm typecheck && pnpm lint && pnpm test` green; 256 unit tests pass.

## Notes for reviewers

- Auto-create at agent birth is deliberately non-fatal (`.catch`): container-less local dev throws in the sandbox constructor and must not wedge agent birth. Minting is identity-only (no container) so birth stays cheap at any agent count.
- Adds `@cloudflare/containers` as a direct dep (for the `OutboundHandler` type; pnpm dedupes it to the SDK's copy).
- The agent boot-context and system prompt now describe `itx.sandbox`; the `sandbox-exec` example and `docs/sandboxes.md` are updated.

---

> [!NOTE]
> **High Risk**
> Changes sandbox network egress (MITM + project policy routing) and agent birth wiring; a mis-registered outbound handler would silently bypass policy, and broader sandbox path rules affect identity minting for every agent.
> 
> **Overview**
> Agents get a **durable `itx.sandbox` capability** at birth: the birth certificate mounts an itx-expression `sandboxes.get(<agent path>)` on the capability host, eagerly mints sandbox DO identity on agent creation (non-fatal on failure), and prompts/types/docs now describe dotted `itx.sandbox.*` calls alongside standalone `itx.sandboxes.get(path)`.
> 
> **Sandbox paths are no longer limited to `/sandboxes/...`** — any codec-safe non-root path is valid (including `/agents/...`), so an agent’s container shares the agent’s path in a separate DO namespace.
> 
> **All sandbox container egress is forced through the project egress path**: `CloudflareSandboxDurableObject` enables HTTPS MITM (`interceptHttps`), registers outbound forwarding via a `static {}` block (avoids ES2024 field-shadowing bypass), resolves `projectId` from the container id, and `worker.ts` re-exports `ContainerProxy`. Adds deployed e2e `sandbox-egress.e2e.test.ts`, `@cloudflare/containers`, and optional local containers via `OS_SANDBOX_CONTAINER_LOCAL_DEV`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 05ae3ae90756cfe0036c3475ac3fb2434a60240c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Environment Config Lease

<details>

| Auth | released | `05ae3ae` | [https://auth.iterate-preview-3.com](https://auth.iterate-preview-3.com) | 19.3s | 505ms | 4.1s | [Workflow run](https://github.com/iterate/iterate/actions/runs/128543827363062) | 2026-07-04T21:08:13.791Z | Preview app released. |
| OS | released | `05ae3ae` | [https://os.iterate-preview-3.com](https://os.iterate-preview-3.com) | 49.7s | 75.4s | 6.0s | [Workflow run](https://github.com/iterate/iterate/actions/runs/128543827363062) | 2026-07-04T21:08:15.677Z | Preview app released. |
| Streams Example App | released | `fa4e5ad` | [https://streams-example-app-preview-3.iterate-dev-preview.workers.dev](https://streams-example-app-preview-3.iterate-dev-preview.workers.dev) | 14.0s | 14.8s | 919ms | [Workflow run](https://github.com/iterate/iterate/actions/runs/73576023855040) | 2026-07-04T21:08:10.580Z | Preview app released. |

</details>


jonastemplestein:
Should be secret.__describe

And should be tested in dev

And entire sandbox sdk surface area should automatically be available

And why can "/" paths not have a sandbox?

PR #1706, opened 2026-07-07
[codex] Refresh sessions after auth JWKS drift

jonastemplestein:
## Summary

This fixes the production OS deploy logout loop where a previously signed-in browser lands on the OS sign-in page with:

> Sign-in could not be verified
>
> Auth returned a session, but OS could not verify it. Try signing in again; if this keeps happening, contact Iterate support.

There are two fixes here:

- **Deploy-time root cause:** relying-party deploys now always fetch the live auth worker JWKS when baking `APP_CONFIG_ITERATE_AUTH__JWKS`. A pinned Doppler JWKS is ignored by the deploy helper; if the live issuer JWKS cannot be fetched after the retry window, the deploy fails closed.
- **Runtime recovery:** if a real JWT-shaped session cookie fails verification, `@iterate-com/auth/server` refreshes once with the cookie's refresh token, verifies the new token pair, and rotates the cookie instead of treating the user as logged out.

I also removed the Doppler JWKS snapshot path rather than leaving it as a dormant fallback:

- `apps/os/scripts/sync-auth-clients.ts` no longer writes `APP_CONFIG_ITERATE_AUTH__JWKS`.
- OS local dev derives a forge-only JWKS binding from `AUTH_FORGE_PRIVATE_JWK` while generating Wrangler config.
- `APP_CONFIG_ITERATE_AUTH__JWKS` was deleted from `os/prd` and `os/preview_1` through `os/preview_9` in Doppler and verified absent without printing secret material.

## Confirmed Production Evidence

I checked production key metadata without printing key material or private secrets.

Before the Doppler cleanup, `os/prd` had a pinned `APP_CONFIG_ITERATE_AUTH__JWKS` with:

- `kid`: `lYquIFJhuLEjMp4u8QP77GHHVwZRksfm`
- public JWK fingerprint prefix: `05dad5679a349e14`

Live `https://auth.iterate.com/api/auth/jwks` served:

- `kid`: `8omkm0VcESZimwi02IQm2GXZwXNwQwGS`
- public JWK fingerprint prefix: `b95164626213bef0`

So production OS deploys were not just theoretically exposed to JWKS drift: the Doppler JWKS snapshot was stale relative to live production auth. Before this PR, the deploy helper could bake that stale snapshot into OS instead of using live auth.

That means a production OS deploy could ship a worker whose local JWT verifier trusted the wrong auth signing key. Existing browser cookies could still be valid in auth, but the newly deployed OS worker could no longer verify their access/id token pair.

## Why This Can Happen On An OS Production Deploy

OS does not introspect browser sessions against auth on every request. The normal hot path is local JWT verification:

1. The browser carries an `iterate_session` cookie.
2. That cookie contains an auth-worker-issued access token, id token, and usually a refresh token.
3. OS verifies the access/id JWTs locally with a JWKS in its runtime config.
4. If verification succeeds, OS builds a request-local principal from the claims.

For performance and for forge-token support, OS bakes an auth JWKS at deploy time. `apps/os/scripts/deploy.ts` calls `bakeStaticAuthJwks(...)`, which fetches `${authBaseUrl}/api/auth/jwks` and then merges the forge public key.

The bug was that a Doppler-pinned `APP_CONFIG_ITERATE_AUTH__JWKS` could win over, or stand in for, the live fetch. That was originally intended as a break-glass path for deploying while auth is down, but `apps/os/scripts/sync-auth-clients.ts` also wrote a JWKS snapshot into Doppler. Once that snapshot went stale, deploys could faithfully rebake stale trust.

## Why There Is No Doppler Fallback Now

Failing an OS deploy is better than deploying a worker that cannot verify the sessions auth is actually issuing.

The live auth worker is the issuer and therefore the source of truth for deployed relying-party trust. If OS cannot fetch the issuer JWKS after retrying for about four minutes, we do not know which key set production auth is actually using. Baking a cached Doppler value at that point silently converts an auth availability problem into a production login-verification problem.

The old fallback only helped one case: shipping OS while the auth JWKS endpoint is unavailable. It hurt the more important case: deploying correct trust after auth rotated keys but Doppler kept an old snapshot. That is the class that triggered this investigation.

The cases are now handled this way:

- **Auth is still deploying in preview CI:** keep polling. This is expected because preview deploys run auth and relying parties in parallel.
- **Auth is down or its JWKS endpoint is broken after the retry window:** fail the OS/Semaphore/streams-example deploy and fix or roll back auth first.
- **Emergency OS hotfix during an auth outage:** do not silently use a stale trust snapshot. If we ever need that, it should be an explicit, reviewed break-glass path with fresh operator-provided key material, not an automatic Doppler fallback.
- **Local/dev minting:** do not store a JWKS snapshot in Doppler. Local OS dev derives the forge public key from `AUTH_FORGE_PRIVATE_JWK`; deployed workers merge the forge public key while baking from live auth JWKS.

## Why Users Saw The Sign-In Verification Error

Before this PR, a request with an `iterate_session` cookie that failed local JWT verification returned `session: null`. OS middleware saw “there is a session cookie, but auth verification failed” and converted that into `session_verification_failed`, which produced the user-facing sign-in error.

The user was not intentionally logged out. OS simply could not verify the cookie's access/id token pair after deploy-time key drift.

The important detail is that the cookie can still contain a usable refresh token. If auth still accepts that refresh token, OS can ask auth for a fresh access/id token pair signed by the current auth key. That is the runtime repair this PR adds.

### Deploy Path

`scripts/lib/bake-auth-jwks.ts` now:

- fetches the live auth JWKS as the deploy-time source of truth
- ignores a pinned Doppler JWKS while baking deploy trust
- fails the deploy if the live fetch keeps failing after retry
- still merges the forge public key into the baked JWKS
- keeps the production forge-key safety gate unchanged

`scripts/lib/bake-auth-jwks.test.ts` covers:

- pinned Doppler JWKS is ignored when the live auth JWKS is available
- pinned Doppler JWKS does not mask a live fetch failure
- forge public key still merges into the live JWKS

### Doppler And Local Dev Cleanup

`apps/os/scripts/sync-auth-clients.ts` now syncs OAuth clients and related OS URLs/secrets only; it no longer fetches auth JWKS or writes `APP_CONFIG_ITERATE_AUTH__JWKS` to Doppler.

`apps/os/scripts/generate-wrangler-config.ts` no longer lists `APP_CONFIG_ITERATE_AUTH__JWKS` as a Doppler-loaded local optional secret. During local dev config generation it derives `{ keys: [forgePublicJwk] }` from `AUTH_FORGE_PRIVATE_JWK` and binds that as `APP_CONFIG_ITERATE_AUTH__JWKS` only for local workerd. The generated binding strips `d`, so the private key is not passed into the worker.

Docs were updated in `apps/os/docs/architecture-and-operations.md`, `docs/dev-environments.md`, and `apps/auth/README.md`. The local-dev auth-minting task is marked resolved.

### Runtime Auth Path

`apps/auth/src/lib/server.ts` now:

- factors token-pair verification into `verifyTokenSet(...)`
- tries one refresh after verification failure when the cookie has JWT-shaped tokens and a refresh token
- verifies the refreshed access/id token pair before authenticating the request
- emits `Set-Cookie` only after successful refresh+verify
- applies the same repair path to `/api/iterate-auth/session`
- keeps malformed/non-JWT cookies on the fail-fast path so junk cookie contents do not hit the token endpoint

`apps/os/src/auth/iterate-auth-login.test.ts` covers both affected runtime entry points:

- middleware authentication refreshes once when old cookie tokens fail the current JWKS
- `/api/iterate-auth/session` does the same and returns an authenticated response with a rotated cookie

## Why Both Layers Are Useful

The deploy fix prevents future OS deploys from rebaking stale Doppler trust when production auth is reachable.

The runtime fix handles the remaining legitimate drift cases:

- a user cookie was issued before a real auth key rotation
- auth's current JWKS no longer serves the key that signed an older but refreshable token pair
- a worker version is already deployed with a stale baked JWKS
- a browser hits a worker deployed before this fix, then receives a refreshed token pair that matches current auth

Refreshing is the right repair because the refresh token is already the session-family authority for rotating access/id tokens. OS still authenticates only after verifying the refreshed JWTs against issuer, audience, and client id.

## Tests

- `pnpm --dir scripts test`
- `pnpm --dir apps/os test src/auth/iterate-auth-login.test.ts`
- `pnpm --dir apps/auth typecheck`
- `pnpm --dir apps/os typecheck`
- `pnpm exec oxlint apps/auth/src/lib/server.ts apps/os/src/auth/iterate-auth-login.test.ts scripts/lib/bake-auth-jwks.ts scripts/lib/bake-auth-jwks.test.ts apps/os/scripts/generate-wrangler-config.ts apps/os/scripts/sync-auth-clients.ts --deny-warnings`
- `git diff --check`
- fake local forge-key Wrangler generation smoke: verified generated `APP_CONFIG_ITERATE_AUTH__JWKS` contains the public forge key, omits private `d`, and is not loaded as a local secret

## User Impact

After this lands, a normal OS deploy should bake the live production auth JWKS rather than a stale Doppler snapshot. If auth is unreachable during deploy, the deploy fails instead of shipping stale trust. If a user still arrives with a refreshable cookie whose JWT pair fails verification, their first request can rotate the cookie and continue authenticated instead of sending them to the sign-in verification error page.

## Environment Config Lease

<details>

| Auth | released | `5d221af` | [https://auth.iterate-preview-2.com](https://auth.iterate-preview-2.com) | 28.8s | 617ms |  | 3.0s | [Workflow run](https://github.com/iterate/iterate/actions/runs/238332168894396) | 2026-07-07T20:18:50.956Z | Preview app released. |
| OS | released | `5d221af` | [https://os.iterate-preview-2.com](https://os.iterate-preview-2.com) | 170.5s | 481.4s | 1 retried: runs "sandbox-exec" through the project REPL (specs x1) | 4.9s | [Workflow run](https://github.com/iterate/iterate/actions/runs/238332168894396) | 2026-07-07T20:18:52.759Z | Preview app released. |
| Semaphore | released | `5d221af` | [https://semaphore.iterate-preview-2.com](https://semaphore.iterate-preview-2.com) | 15.9s | 7.6s |  | 711ms | [Workflow run](https://github.com/iterate/iterate/actions/runs/238332168894396) | 2026-07-07T20:18:48.617Z | Preview app released. |
| Streams Example App | released | `5d221af` | [https://streams.iterate-preview-2.com](https://streams.iterate-preview-2.com) | 14.7s | 19.3s |  | 842ms | [Workflow run](https://github.com/iterate/iterate/actions/runs/238332168894396) | 2026-07-07T20:18:48.752Z | Preview app released. |

</details>

---

> [!NOTE]
> **High Risk**
> Changes production session verification and deploy-time JWKS trust; a mistake could block deploys or break login until rolled forward, though tests and fail-closed deploy reduce silent mis-trust.
> 
> **Overview**
> Fixes production OS **“Sign-in could not be verified”** loops when deploy-time trust drifted from the live auth signing keys.
> 
> **Deploy trust:** `bakeStaticAuthJwks` always uses the **live** auth `/jwks` (pinned Doppler `APP_CONFIG_ITERATE_AUTH__JWKS` is ignored). If the fetch still fails after the retry window, the deploy **fails closed** instead of baking stale keys. `sync-auth-clients` no longer writes JWKS snapshots to Doppler.
> 
> **Local dev:** Wrangler config generation binds a **forge-only** JWKS derived from `AUTH_FORGE_PRIVATE_JWK` (not Doppler JWKS), so `pnpm auth:mint --browser-url` works without a separate auth worker.
> 
> **Runtime recovery:** `@iterate-com/auth/server` centralizes JWT pair verification in `verifyTokenSet` and, on verify failure for JWT-shaped cookies with a refresh token, **refreshes once**, re-verifies, and rotates the session cookie (middleware + `/session`). Docs and tests cover deploy baking, refresh-on-drift, and session/userinfo failure paths.
> 
> Also documents preview sandbox **`max_instances` 31** (quota fit for `standard-1`) in ops/flake-hunt notes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5d221afac4209a2dae729f6cb779a11394ae3727. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

| Group | Changes |
| --- | ---: |
| Product | +129 -39 🟩🟩🟩🟩🟥 |
| Tests | +139 -0 🟩🟩🟩🟩⬜ |
| CI & scripts | +19 -42 🟩🟥⬜⬜⬜ |
| Docs | +36 -20 🟩🟥⬜⬜⬜ |
| **Total** | **+323 -101** 🟩🟩🟩🟩🟥 |

<sub>Source lines changed (blank lines and JS comments ignored) between b938431 and 5d221af, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>


jonastemplestein (commented):
Self-review follow-up:

Findings: no blocking findings found.

Residual risk / follow-up:

- Deploy behavior intentionally changed to fail closed if the live auth JWKS endpoint stays unavailable after the retry window. That is the right tradeoff versus silently baking stale trust, but it means an auth outage now blocks relying-party deploys instead of being papered over by Doppler.
- The runtime repair path adds a token refresh on JWT verification failure for JWT-shaped cookies with refresh tokens. Focused unit coverage exercises middleware and `/session`, but the final confidence step is the normal preview deploy/e2e lane plus a production deploy after merge.
- `APP_CONFIG_ITERATE_AUTH__JWKS` has been deleted from `os/prd` and `os/preview_1..9` in Doppler and verified absent. `auth:sync-clients` no longer recreates it. Future cleanup should keep that key out of Doppler; deployed workers still receive the freshly baked secret during deploy.
- Local dev now derives the forge public JWKS from `AUTH_FORGE_PRIVATE_JWK` at generated-Wrangler-config time. I verified with a fake forge key that the generated binding contains the public key, omits private `d`, and is not listed as a Doppler-loaded local secret.

Validation run:

- `node --import tsx --test scripts/lib/bake-auth-jwks.test.ts`
- `pnpm --dir apps/os test src/auth/iterate-auth-login.test.ts`
- `pnpm --dir apps/auth typecheck`
- `pnpm --dir apps/os typecheck`
- `pnpm exec oxlint apps/auth/src/lib/server.ts apps/os/src/auth/iterate-auth-login.test.ts scripts/lib/bake-auth-jwks.ts scripts/lib/bake-auth-jwks.test.ts apps/os/scripts/generate-wrangler-config.ts apps/os/scripts/sync-auth-clients.ts --deny-warnings`
- `git diff --check`

jonastemplestein (commented):
Self-review follow-up after Bugbot comments:

Addressed both actionable inline findings:

- `/session` now tracks only successful refresh rotations as `refreshed`, so a failed non-fatal proactive refresh does not block the verification-failure repair refresh.
- `/session` now catches `getUserInfo` failures and returns the controlled unauthenticated response with the session cookie cleared, matching the previous behavior.

Added regression coverage for both cases in `apps/os/src/auth/iterate-auth-login.test.ts`; the focused auth test file now passes with 14 tests.

Validation rerun after the fix:

- `pnpm --dir apps/os test src/auth/iterate-auth-login.test.ts`
- `pnpm --dir apps/auth typecheck`
- `pnpm --dir apps/os typecheck`
- `pnpm exec oxlint apps/auth/src/lib/server.ts apps/os/src/auth/iterate-auth-login.test.ts scripts/lib/bake-auth-jwks.ts scripts/lib/bake-auth-jwks.test.ts apps/os/scripts/generate-wrangler-config.ts apps/os/scripts/sync-auth-clients.ts --deny-warnings`
- `node --import tsx --test scripts/lib/bake-auth-jwks.test.ts`
- `git diff --check`

PR #1716, opened 2026-07-07
fix-stream skill + trial run: HEIC attachment silently kills the agent turn

mmkal:
## What

Two deliverables, per the task file (moved to `tasks/complete/2026-07-07-fix-stream-skill.md`):

### 1. A `fix-stream` skill

`.agents/skills/fix-stream/SKILL.md` (caveman-compressed): given a broken stream URL like `https://os.iterate.com/projects/misha2/agents/streams/agents/1648` — dump the journal via itx, diagnose the user-level complaint, seed the real events into an in-memory test stream with checkpoint priming (DO-restart semantics: reduced state covers history, side effects fire only for the newly appended "bad part"), fake the LLM transport to behave exactly like the real provider did, assert broadly that the turn doesn't go wrong, push red, fix, push green, then minimise the fixture. The skill was written first, trial-run below, then updated with lessons learned (hop-by-hop delivery driving, a red-for-the-right-reason dump trick, the minimal-fixture recipe).

### 2. Trial run: HEIC attachment silently kills the agent turn

In prod stream misha2 `/agents/1648`, the user sent "Who's this" with an iPhone photo (`image/heic`). `toResponsesInput` forwards *any* `image/*` attachment to OpenAI as a native `input_image` part; OpenAI supports only jpeg/png/gif/webp, 400s the whole request, and the agent goes permanently silent:

```
[3008] input-added        "Who's this" + IMG_3159.heic
[3012] llm-response-chunk 400 "The image data you provided does not represent a valid image..."
[3014] llm-request-completed { result: { status: "failure" } }
        (silence forever)
```

The fix whitelists OpenAI-ingestible image formats; anything else rides as the existing `renderFileHintLine` fallback ("never fail the turn") — the same path every non-image file already takes — so the agent can convert/inspect the bytes and still answer.

The repro test (`stream-repros/misha2-1648-heic-image-silence.test.ts`) replays the real journal and reproduced prod byte-for-byte, down to `llmRequestId: 3010` and the exact 400 error chunk. Its assertion is deliberately at the user level: after the input, `agent/output-added` must eventually appear.

## Commit sequence (red/green demonstrated on CI)

1. `c574882` task spec · `3065c22` skill
2. `93dcd82` **red**: repro test + full-feed fixture (338 events; chunks + `rawResponse` stripped at dump time) + harness extraction to `test-helpers.ts` (with a fix: `MemoryStream.append` now assigns last-offset+1 so gap-y seeded histories don't collide)
3. `f3391eb` **green**: the `input_image` whitelist fix
4. `8a0243f` **red again on purpose**: fixture minimised 338 events → 13 (8K), fix temporarily reverted so CI proves the minimal fixture still repros
5. `b3c8366` **green**: fix unreverted
6. `210b239` skill updated with trial-run lessons; task → complete
7. `7f2de1b` review feedback applied · `978428b` pre-commit hook now formats the regenerated skills registry (root cause of the autofix.ci failure) · `14540fa` fixture marked linguist-generated

## Follow-up (out of scope)

The agent goes silent whenever an LLM request fails permanently — any cause, not just HEIC. Notifying the user without re-triggering a doomed request (the poisoned input would loop) is a design question worth its own task.

## Environment Config Lease

<details>

| Auth | released | `7c8d3e2` | [https://auth.iterate-preview-4.com](https://auth.iterate-preview-4.com) | 18.9s | 582ms |  | 2.9s | [Workflow run](https://github.com/iterate/iterate/actions/runs/79864716774990) | 2026-07-07T22:02:06.295Z | Preview app released. |
| OS | released | `7c8d3e2` | [https://os.iterate-preview-4.com](https://os.iterate-preview-4.com) | 168.4s | 394.7s | 4 retried: catalogue example "sandbox-exec" runs identically across runtimes (vitest x1) · provider toggle: cloudflare-ai answers after llm-provider-selected (vitest x1) · runs "scheduler-agent-checkin" through the project REPL (specs x1) · runs "sandbox-exec" through the project REPL (specs x1) | 4.7s | [Workflow run](https://github.com/iterate/iterate/actions/runs/79864716774990) | 2026-07-07T22:02:08.108Z | Preview app released. |

</details>

| Group | Changes |
| --- | ---: |
| Product | +145 -1 🟩🟩🟩⬜⬜ |
| Tests | +100 -135 🟩🟩🟥🟥🟥 |
| Config | +164 -0 🟩🟩🟩⬜⬜ |
| Docs | +115 -0 🟩🟩⬜⬜⬜ |
| Other | +102 -0 🟩🟩⬜⬜⬜ |
| **Total** | **+626 -136** 🟩🟩🟩🟩🟥 |

<sub>Source lines changed (blank lines and JS comments ignored) between 3126c8b and 7c8d3e2, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>

---

> [!NOTE]
> **Medium Risk**
> Changes how user image attachments are sent to OpenAI (agent chat path); scope is small and covered by a prod-derived repro, but wrong MIME routing could still affect vision behavior for edge formats.
> 
> **Overview**
> Adds a **`fix-stream`** agent skill that turns broken prod stream URLs into red/green in-memory repro tests (dump journal → seed fixture → fake provider → assert user-visible outcome → minimise fixture).
> 
> **Product fix:** OpenAI WS `toResponsesInput` now only sends **jpeg/png/gif/webp** as native `input_image` parts; other image types (e.g. **HEIC**) use the existing **`renderFileHintLine`** path so the whole Responses request is not 400’d and the agent can still reply.
> 
> **Tests:** Shared harness moved to `test-helpers.ts` (`MemoryStream` offset assignment fixed for gap-y seeded journals); new **`stream-repros/misha2-1648-heic-image-silence`** test with minimised prod events. Minor **lint-staged** formatting for generated skills registry and **linguist-generated** markers for repro JSON.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7c8d3e2225646f28118f81334fa9e2c4e4b5a147. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.agents/skills/fix-stream/SKILL.md
```diff
@@ -0,0 +1,105 @@
+---
+name: fix-stream
+description: Turn a broken agent-stream URL (os.iterate.com/projects/<slug>/agents/streams/agents/<id>) into a red repro test, then a green fix, then a minimal fixture. Use when user says "fix <stream url>", pastes a stream URL with a complaint, or reports an agent chat that went wrong/silent.
+publish: false
+---
+
+# fix-stream
+
+Given stream URL where agent chat went wrong. Dump events. Judge complaint. Seed real events into unit test. Red test -> PR -> fix -> green -> minimise fixture. Written caveman-style: terse on purpose, all substance kept.
```
mmkal:
```suggestion
Given stream URL where agent chat went wrong. Dump events. Judge complaint. Seed real events into unit test. Red test -> PR -> fix -> green -> minimise fixture.
```

ffs

.agents/skills/fix-stream/SKILL.md
```diff
@@ -0,0 +1,105 @@
...
+```bash
+doppler run --project os --config prd -- pnpm --dir apps/os cli itx run \
+  -e 'return (await itx.projects.list({ scope: "deployment" })).filter(p => p.slug === "<slug>")'
+```
+
+Dump full journal (getEvents caps at 500 -> page):
+
+```bash
+doppler run --project os --config prd -- pnpm --dir apps/os cli itx run \
+  --context <prj_id> --file dump.ts   # script body below, save in scratchpad
```
mmkal:
```suggestion
  --context <prj_id> --file dump.ignoreme.ts   # script body below, save in scratchpad
```

.agents/skills/fix-stream/SKILL.md
```diff
@@ -0,0 +1,105 @@
...
+  after = page[page.length - 1].offset;
+}
+return all;
+```
+
+Strip pnpm banner from stdout before JSON.parse (find first `[\n`).
+
+## 2. Diagnose
+
+Print conversation: `user-message-received` payload.content vs `web-message-sent` payload.message, with offsets. Find where user visibly lost: silence after input, wrong answer, error leak. Zoom into offsets around bad part — full payloads. Usual smoking gun: `agent/llm-request-completed` with `result.status: "failure"` and no `agent/output-added` after. Complaint = user-level symptom, not mechanism. Write it down before reading product code.
```
mmkal:
```suggestion
Print conversation: `user-message-received` payload.content vs `web-message-sent` payload.message, with offsets. Find where user visibly lost: silence after input, wrong answer, error leak. Zoom into offsets around bad part — full payloads. Complaint = user-level symptom, not mechanism. If there was an API error but the system recovered fine, that's probably not what the complaint is about. Write down the complaint before reading product code.
```

.agents/skills/fix-stream/SKILL.md
mmkal: Also I don't think we should say "unit test" here. Unclear to me what level of test this might be , but not unit i wouldn't think

apps/os/src/domains/agents/stream-repros/misha2-1648.events.json
mmkal, line 1: oof. this has got to be set as linguist-generated=true in .gitattributes or whatever the thing we do is

.agents/skills/fix-stream/SKILL.md
mmkal: 🤖 Applied in 7f2de1b93 — dropped the self-commentary, and replaced the "unit test" framing throughout with "repro test" / "in-memory test: real processors + real reducers, fake transport, no deployment, no real LLM", which is what it actually is.

.agents/skills/fix-stream/SKILL.md
mmkal: 🤖 Applied in 7f2de1b93 — `dump.ignoreme.ts`.

.agents/skills/fix-stream/SKILL.md
mmkal: 🤖 Applied verbatim in 7f2de1b93 (including "recovered fine ≠ the complaint").

apps/os/src/domains/agents/stream-repros/misha2-1648.events.json
mmkal: 🤖 Added `**/stream-repros/*.events.json linguist-generated=true` to .gitattributes (14540fabd). Also fixed the file's missing trailing newline, which my first append tripped over.

jonastemplestein:
This is a good fix, but in addition a more general fix would be good. It should not be possible for an error of any kind to brick a stream. Just like pi agent, any errors should be turned into input events that can be included in subsequent requests

mmkal:
🤖 Agreed — that's the general fix this PR deliberately left out. Done as a stacked PR: https://github.com/iterate/iterate/pull/1729. Failed LLM requests now become model-visible input events (same pattern as thrown scripts), with a consecutive-failure cap (2 auto-retries, counter resets on success) so a persistent failure can't retry-loop while the error still sits in context for the next user turn.

mmkal:
Yep agreed this is just a skill I want to have (even in this not-very-good form) so we can say `fix-stream https://os.iterate.com/projects/misha2/agents/streams/agents/1648` and it'll do roughly the right thing. I think we'll want to change what we consider "the right thing" significantly from this though

PR #1735, opened 2026-07-07
itx.chat.sendMessage takes a plain string

mmkal:
## What

`itx.chat.sendMessage` now takes the message directly as a string, and anything beyond the message rides in an optional second `options` argument:

```ts
// before
await itx.chat.sendMessage({ message: "Checking your email now..." });
await itx.chat.sendMessage({ message: "Here you go!", files: [{ filename: "cat.png", contentType: "image/png", data: image }] });

// after
await itx.chat.sendMessage("Checking your email now...");
await itx.chat.sendMessage("Here you go!", { files: [{ filename: "cat.png", contentType: "image/png", data: image }] });
```

All prompts (agent contract, MCP-scope guidance), docs, examples, the embedded itx type surface, the project-repo-template `sdk.ts`, and tests move to the string form. `AgentChatSendOptions` (exported from `types.ts`) is the second argument's type — future options land there, never back in an object first arg.

## Back-compat

None — by request. The runtime originally kept accepting the legacy `{ message, files? }` object form (live agents' conversation histories are full of it), but review asked for a clean break: `sendMessage(message: string, options?: AgentChatSendOptions)` is the only signature, in types, prompts, and at runtime. A legacy-form call now fails when the script runs; in-flight agents are expected to self-correct off the error plus the string-form examples that now saturate their prompts.

The e2e project-tool test asserts the string form and the `("msg", { files })` second-arg attachment plumbing, including stored attachment records on the `web-message-sent` payloads.

## Why

Beyond less ceremony, this sets up a follow-up PR: when an agent streams `itx.chat.sendMessage("Checking your ema…` token-by-token, the browser can pre-parse the unclosed string literal and show a live message preview before the script even runs. A string literal in argument position is trivially extractable mid-stream; an object literal isn't.

## Notes

- The task spec mentioned regenerating `itx-api.generated.ts` — no such file exists; the two real generated artifacts (`types-source.generated.ts`, `project-repo-template.generated.ts`) were regenerated and their freshness tests pass.
- `pnpm typecheck && pnpm lint && pnpm format && pnpm test` all green locally; the e2e lane (needs a live env) is left to CI.

See `tasks/complete/2026-07-08-itx-chat-send-message-string.md` for the full spec.

## Environment Config Lease

<details>

| Auth | released | `b5752df` | [https://auth.iterate-preview-4.com](https://auth.iterate-preview-4.com) | 18.4s | 527ms |  | 2.9s | [Workflow run](https://github.com/iterate/iterate/actions/runs/310727857902050) | 2026-07-08T12:00:46.417Z | Preview app released. |
| OS | released | `b5752df` | [https://os.iterate-preview-4.com](https://os.iterate-preview-4.com) | 161.5s | 368.8s | 3 retried: seeded waitrose integration files > seeded files build, mount, and dispatch (vitest x1) · itx > OpenAPI built-in connects directly and mounts as a described capability (vitest x1) · runs "sandbox-exec" through the project REPL (specs x1) | 4.3s | [Workflow run](https://github.com/iterate/iterate/actions/runs/310727857902050) | 2026-07-08T12:00:47.891Z | Preview app released. |

</details>

| --- | ---: | ---: |
| Product | +50 -31 | +31 -23 🟩🟥⬜⬜⬜ |
| Tests | +30 -16 | +30 -16 🟩🟥⬜⬜⬜ |
| Docs | +136 -2 | +121 -2 🟩🟩🟩🟩🟩 |
| Generated | +14 -10 | +13 -10 🟩⬜⬜⬜⬜ |
| **Total** | **+230 -59** | **+195 -51** 🟩🟩🟩🟩🟥 |

<sub>Lines counts every changed line; Significant ignores blank lines and JS comments. Between e0b7242 and b5752df, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>

---

> [!NOTE]
> **Medium Risk**
> Touches the agent web-chat RPC surface and attachment handling used on every agent reply; behavior is mostly additive with explicit back-compat, but regressions would affect all in-flight agents.
> 
> **Overview**
> **`itx.chat.sendMessage`** is documented and typed as a plain string plus an optional second **`AgentChatSendOptions`** argument (e.g. `await itx.chat.sendMessage("Here you go!", { files: [...] })`), replacing the old single `{ message, files? }` object.
> 
> **`AgentChatRpcTarget.sendMessage`** normalizes string vs legacy object input, merges **`options.files`** (including mixed legacy-object + options calls without dropping attachments), and updates **`__describe`** text accordingly.
> 
> Agent system prompts, MCP/onboarding guidance, README/smoke docs, React examples, generated **`types.ts`** / project **`sdk.ts`** snapshots, and unit/e2e tests are aligned on the string form; e2e now asserts string, legacy object, second-arg files, and mixed forms all emit correct **`web-message-sent`** payloads.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5257e24d9d368059ee61ea5baa118509cb41c3bd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


apps/os/src/rpc-targets.ts
```diff
@@ -1660,9 +1661,14 @@ class AgentChatRpcTarget extends RpcTarget implements AgentChat {
...
 
-  async sendMessage(input: Parameters<AgentChat["sendMessage"]>[0]) {
-    const message = input.message.trim();
-    if (message === "") throw new Error("itx.chat.sendMessage requires a non-empty message.");
+  async sendMessage(...[message, options]: Parameters<AgentChat["sendMessage"]>) {
+    // Live agents have conversation history full of the legacy
+    // `sendMessage({ message, files? })` object form, and models imitate
+    // history, so it stays accepted here. The documented form is a plain
+    // string with an optional options second argument.
+    const input = typeof message === "string" ? { message, files: options?.files } : message;
```
mmkal: 🤖 Edge case: a mixed-form call `sendMessage({ message: "x" }, { files: [...] })` silently drops `options.files` (the object branch ignores `options` entirely). Models that half-imitate old history while following the new docs are exactly the population this back-compat exists for, and a silently lost attachment is a confusing failure mode. Consider merging instead, e.g. `: { ...message, files: message.files || options?.files }` — or throwing on the mixed form if you'd rather not define merge semantics. Non-blocking.

apps/os/src/types.ts
```diff
@@ -258,18 +258,28 @@ export interface ProjectRpcTarget {
...
    */
-  sendMessage(input: {
-    message: string;
-    files?: Array<{ contentType: string; data: FileData; filename: string }>;
-  }): Promise<StreamEvent>;
+  sendMessage(
+    // The `{ message, files? }` object form is legacy: live agents have
+    // conversation history full of it and models imitate history, so it stays
+    // accepted. New code always passes a plain string (+ options).
+    message: string | { message: string; files?: AgentChatSendOptions["files"] },
```
mmkal: 🤖 Worth being deliberate about: `types.ts` is embedded verbatim into every agent system prompt via `ITX_TYPES_SOURCE` (agent-processor-contract.ts imports it), so the legacy `{ message }` union — and this comment — are visible to the model. That means the legacy form isn't really "undocumented" from the model's perspective, and since the follow-up streaming-preview PR depends on models emitting a string literal in argument position, the type surface is the strongest lever you have on which form they emit. The steering comment ("New code always passes a plain string") mitigates this, and showing the union does have the upside of telling the model its own historical object-form calls were valid. Fine to ship as-is, but if the follow-up PR finds models still emitting the object form, narrowing the *prompt-facing* generated surface to `sendMessage(message: string, options?)` while keeping the runtime union would be the fix to reach for.

apps/os/e2e/vitest/itx.e2e.test.ts
```diff
@@ -1859,14 +1859,22 @@ describe("itx", () => {
...
       type: AGENT_OUTPUT_ADDED_TYPE,
       payload: {
         content: fencedAgentScript(`
           async (itx) => {
             const message = await itx.projectTool.format({ text: "project-capability" });
-            await itx.chat.sendMessage({ message });
+            await itx.chat.sendMessage(message);
+            // Live agents' history is full of the legacy object form — it must
+            // keep working alongside the plain-string form.
+            await itx.chat.sendMessage({ message: "legacy object form still works" });
```
mmkal: 🤖 Nice that both call forms are exercised through the real rpc-target here. One gap: the new `("msg", { files })` second-arg plumbing (`files: options?.files` in `AgentChatRpcTarget.sendMessage`) has no direct test — this e2e covers both *message* forms without files, and the files coverage in `agent-processors.test.ts` operates at the event level, below the rpc-target. A wrong property name in the normalization would ship green. Cheap to lock in (the follow-up streaming PR will lean on this signature), but fine as a follow-up. Non-blocking.

mmkal (commented):
🤖 Reviewed for correctness of the back-compat union + options-as-second-arg handling, straggler `sendMessage({ message` callsites, generated-file freshness, test quality, and prompt wording.

Overall this looks solid:

- **No stragglers.** A repo-wide grep for object-form calls finds only intentional survivors: the back-compat e2e assertion, the task file quoting the old form, the explanatory comment in `rpc-targets.ts`, and `stream-repros/misha2-1648.events.json` (a captured historical stream — correctly left alone).
- **Generated files are fresh.** I ran `types-source.generated.test.ts` and `project-repo-template.test.ts` in the worktree — both pass, and the generated surfaces match `types.ts` / `project-repo-template/sdk.ts`.
- **Prompt wording reads well** — the contract's Promise.all progress guidance, the files rule, and the MCP reply-exactly-once line all consistently use the string form, and the `__describe` text spells out "the message is a plain string".
- `agent.sendMessage(message: string)` (user→agent) untouched, as intended.

Three non-blocking findings inline: a silent-drop edge case when the legacy object form is mixed with the new options arg, a note that the legacy union rides into every agent system prompt via `ITX_TYPES_SOURCE` (so it isn't really "undocumented" from the model's point of view), and a small coverage gap on the new `("msg", { files })` plumbing.

apps/os/src/rpc-targets.ts
mmkal: 🤖 Agreed and fixed in 306ceb5e7: the legacy-object branch now spreads the object and merges files (`message.files || options?.files` — the object's own files win if both are somehow given), so a mixed call keeps its attachment instead of silently dropping it. Covered by the extended e2e test, which sends a mixed-form message and asserts the stored attachment record on the web-message-sent payload.

apps/os/src/types.ts
mmkal: 🤖 Good catch on the lever — ITX_TYPES_SOURCE embedding means "undocumented" only holds for human-facing docs. Shipping as-is for now: the steering comment plus every prompt example promoting the string form should do the job, and showing the union has the upside of telling the model its own historical object-form calls were valid. Recorded in tasks/itx-chat-send-message-string.md under "Possible follow-up": if the streaming-preview PR finds models still emitting the object form, narrow the prompt-facing generated surface to `sendMessage(message: string, options?)` while keeping the runtime union.

apps/os/e2e/vitest/itx.e2e.test.ts
mmkal: 🤖 Added in 306ceb5e7: this test now also sends `("string form with files", { files })` and a mixed-form call, asserting the stored attachment records (filename/contentType/size) on the web-message-sent payloads — verified green against a local dev server. Kept at the e2e level rather than unit because AgentChatRpcTarget has no unit seam today (it reaches for env/DO stubs and R2-backed file storage).

mmkal:
Remove the runtime backcompat

mmkal:
🤖 Done in b5752df72 — `sendMessage(message: string, options?: AgentChatSendOptions)` is now the only signature. Removed: the `{ message, files? }` union from `types.ts` and the template `sdk.ts`, the runtime normalization/mixed-form merge branch in `AgentChatRpcTarget`, and the legacy/mixed-form e2e sends (the string-form + files coverage stays). Generated files regenerated; typecheck/lint/format/test green. PR body updated — a legacy-form call now fails at script runtime, so in-flight agents self-correct off the error plus the string-form examples in their prompts.

PR #1736, opened 2026-07-07
Live chat-message preview from streaming sendMessage snippets

mmkal:
Stacked on #1735 (`itx-chat-send-message-string`) — depends on the plain-string `itx.chat.sendMessage("...")` form landing first. Review this PR's own changes via the diff against that base branch.

While a code-mode agent streams its itx script token by token, the feed now pre-parses the partial snippet: if the first real statement is a chat message, the string literal renders as a live agent chat bubble **while the code is still being generated** — the user reads the message as it streams instead of watching raw code scroll by.

The streaming code block is show-only-when-needed:

- While the message literal is still streaming, the bubble renders **alone** — the code block underneath would show the same words twice.
- After the literal closes, the block stays hidden through trailing punctuation (`");`, newlines, `}`, the closing fence) — a message-only snippet never flashes a code block in just to end the turn.
- Only once a word character streams in after the close — real further code — does the block appear, and then with the leading call redacted to `itx.chat.sendMessage(...)`, since the message already lives in the bubble. The settled "Ran code" view always shows the unmodified script.

```js
// streamed so far (unclosed literal) — the bubble renders ALONE:
async (itx) => {
  const [, inbox] = await Promise.all([
    itx.chat.sendMessage("Checking your email no
```

renders an in-flight assistant bubble — `Checking your email no▊` — and nothing else. Once the literal closes and `itx.integrations…` starts streaming, the code block appears below the bubble, reading `itx.chat.sendMessage(...)`. It's honest fakery: nothing has been sent yet, the blinking cursor marks it in-progress, and when the script actually runs, the real `web-message-sent` row takes over.

### Demo

Recorded against local dev at a phone-ish viewport (~450px) — a real agent turn, streaming for real. The message is exactly "tell me a 200 word story": after the model's thinking streams by, the story bubble builds up token by token (that text is the still-unclosed string literal inside `itx.chat.sendMessage("…`) and **no code block ever appears** — a message-only snippet never needs one — then the settled message row takes over when the script runs:

streaming sendMessage preview demo (mobile viewport)

### How

- `extractStreamingSendMessagePreview(partialCode): { message, literalClosed, redactedCode } | null` — a pure, hard-unit-tested parser (`apps/os/src/components/streaming-send-message-preview.ts`). It strips a leading ```` ```js ```` fence, `async (itx) => {` opener, and comments; matches `(await) itx.chat.sendMessage(` as the first statement — directly or as the first entry of `(const [, x] = await) Promise.all([` (the contract-prompt progress-message pattern) — then reads the possibly-unclosed literal, processing escapes so the preview reads clean. `redactedCode` goes non-null only once a `\w` character streams in after the close: the snippet with the message literal collapsed to `...`.
- Deliberately conservative — `null` for anything else: the legacy object form `sendMessage({ message: ... })`, template `${}` interpolation, non-first statements, concatenation after the literal. A second options argument (`sendMessage("hello", { whatever: 123 })`) keeps the preview — and its word characters surface the (redacted) code block. Half-streamed escapes (`…said \`) and a `$` that might become `${` are held back so every prefix previews a leading slice of the final message.
- Wired into `LiveStepStream` in `agent-feed.tsx` as pure render-time derivation (`useMemo`) from the live LLM step's streaming `responseText` — no reducer, state, or server changes. If the preview bails-to-null mid-stream (`"hi" + name` — the literal wasn't the whole message), the plain code block takes over immediately.
- 17 unit tests simulate token-by-token growth by asserting the preview over **every prefix** of full snippets, including that `redactedCode` stays null until after the close and never hides again once shown (`streaming-send-message-preview.test.ts`).

| --- | ---: | ---: |
| UI components | +319 -2 | +196 -2 🟩🟩🟩🟩⬜ |
| Tests | +316 -0 | +248 -0 🟩🟩🟩🟩🟩 |
| Docs | +244 -0 | +225 -0 🟩🟩🟩🟩🟩 |
| **Total** | **+879 -2** | **+669 -2** 🟩🟩🟩🟩🟩 |

<sub>Lines counts every changed line; Significant ignores blank lines and JS comments. Between 5257e24 and 4b2236a, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>

## Environment Config Lease

<details>

| Auth | released | `0152af0` | [https://auth.iterate-preview-7.com](https://auth.iterate-preview-7.com) | 18.6s | 199ms |  | 2.7s | [Workflow run](https://github.com/iterate/iterate/actions/runs/127968042057691) | 2026-07-08T09:15:54.071Z | Preview app released. |
| OS | released | `0152af0` | [https://os.iterate-preview-7.com](https://os.iterate-preview-7.com) | 147.1s | 124.8s | 6 retried: itx > OpenAPI built-in connects directly and mounts as a described capability (vitest x1) · inbound Slack webhook routes to a slack agent that attempts a Web API reply (vitest x1) · state-only stream subscribe pushes initial state immediately, then state after appends (vitest x1) · onboarding agent replies to a chat message in the feed (specs x1) · can enter the dashboard with a forged session (specs x1) · reactivity page appends a batch and renders every delivered marker (specs x1) | 4.2s | [Workflow run](https://github.com/iterate/iterate/actions/runs/127968042057691) | 2026-07-08T09:15:55.613Z | Preview app released. |

</details>


apps/os/src/components/agent-feed.tsx
```diff
@@ -979,7 +991,12 @@ function LiveStepStream({ step }: { step: AgentUiStep }) {
         </div>
       )}
       {step.responseText === "" ? null : looksLikeCode(step.responseText) ? (
-        <StreamingCodeBlock code={step.responseText} />
+        <>
+          {sendMessagePreview == null ? null : (
```
mmkal, line 1012: 🤖 **Gate mismatch: `looksLikeCode` hides previews the parser deliberately supports.** The preview only renders inside the `looksLikeCode(step.responseText)` branch, but `CODE_START_PATTERN` is `/^\s*(async|await|function|const|let|import)\b/` — so snippets starting with bare `itx.chat.sendMessage("...`, `void itx.chat.sendMessage("...` (the head pattern explicitly allows `void`), or bare `Promise.all([itx.chat.sendMessage(` (the opener pattern makes `await` optional) parse to a preview that never renders; the raw code falls into the plain-prose branch instead. Verified:

```
'itx.chat.sendMessage("Hi there!")'  preview: "Hi there!"  looksLikeCode: false
'void itx.chat.sendMessage("Hi")'    preview: "Hi"         looksLikeCode: false
'Promise.all([itx.chat.sendMessage(' preview: "Hi"         looksLikeCode: false
```

The very first unit test (`await itx...` — that one does match) plus many others exercise the bare `itx.` form heavily, so the tests certify behavior the UI can't reach. Low real-world impact since contract snippets start with a fence or `async (itx) => {`, but a non-null preview *is* proof the text is code — `looksLikeCode(step.responseText) || sendMessagePreview != null` would close the gap for free. Alternatively drop the `void`/optional-`await` tolerance from the parser so the two layers agree.

apps/os/src/components/streaming-send-message-preview.ts
```diff
@@ -0,0 +1,211 @@
...
+  return out;
+}
+
+/**
+ * After the literal closes, the call must still look like
+ * `sendMessage("...")` or `sendMessage("...", options)` — a `)` or `,` (or
+ * nothing yet, mid-stream). Anything else (`"hi" + name`) means the literal
+ * wasn't the whole message, so drop the preview.
+ */
+function literalCloseLooksWellFormed(text: string, index: number): boolean {
```
mmkal, line 199: 🤖 **The streaming-prefix invariant only holds for snippets that stay recognizable — bail-later inputs make the preview appear and then vanish.** Verified over prefixes of `itx.chat.sendMessage("hi" + name)`:

```
'...("hi'      -> "hi"
'...("hi" '    -> "hi"   (close quote + trailing whitespace at EOI => well-formed)
'...("hi" +'   -> null   (bubble disappears)
```

Same for `` `hi ${name}` ``: previews `"hi "` right up until `${` streams in, then reverts to null. I think this is genuinely unavoidable — any preview of an unclosed literal can be invalidated by a later token — and bail-to-null is the right honest choice. But two things undersell it:

1. The doc comment says template literals with `${}` "are not previewed" — they are, up to the `$`; the user briefly sees a bubble that then vanishes.
2. `expectPreviewOverEveryPrefix`'s no-revert assertion is scoped to well-formed snippets, and the bail tests only assert the *final* string is null — nothing documents the mid-stream appear-then-vanish behavior as intended. A short test (e.g. `distinctPreviews('itx.chat.sendMessage("hi" + name)')` ending in `null`) would pin it down and stop a future refactor from "fixing" it in either direction accidentally.

No code change needed in my view — just a doc tweak and a pinning test.

apps/os/src/components/agent-feed.tsx
```diff
@@ -990,6 +1007,31 @@ function LiveStepStream({ step }: { step: AgentUiStep }) {
...
+function StreamingSendMessagePreview({ text }: { text: string }) {
+  return (
+    <Message
+      from="assistant"
+      className="py-1"
+      data-testid="agent-feed-message-preview"
+      data-kind="assistant-preview"
+    >
+      <MessageContent>
+        <div className="whitespace-pre-wrap leading-6">
```
mmkal: 🤖 **Handoff observation (minor, non-blocking).** The preview renders as plain `whitespace-pre-wrap` text inside the indented live rail (`border-l-2 pl-4`, under the amber spinner), while the settled `web-message-sent` row it hands off to is a top-level feed row rendered through `MessageResponse` (streamdown markdown). So a markdown-y message previews as raw source (`**Done!** [link](...)`) and then re-renders formatted at a different indentation — plus the acknowledged unmount-gap from the task file in between. This is consistent with how the existing plain-prose streaming branch below behaves, and the blinking cursor + rail placement do make it read as in-progress, so I'd call the UX honest. If the jump bothers anyone later, `<MessageResponse>` (with the default `parseIncompleteMarkdown`, i.e. the streaming mode the settled row opts out of) is already imported in this file and would make the preview typographically identical to its successor. Fine to leave as-is.

mmkal (commented):
🤖 Reviewed the delta vs `itx-chat-send-message-string` (parser, React wiring, preview UX, tests). Overall this is solid: the parser's hold-before-incomplete-escape semantics are careful, the every-prefix test helper genuinely enforces the streaming invariant for well-formed snippets, hooks usage is clean (unconditional `useMemo`, no state/effects), and the per-tick re-parse is O(snippet length) which is trivial at these sizes. I verified the 15 unit tests pass and probed the parser with adversarial inputs locally. Three non-blocking findings, inline: (1) a real gate mismatch where `looksLikeCode` hides previews the parser deliberately supports (bare `itx.` / `void itx.` / bare `Promise.all([` starts), (2) the preview can appear and then vanish mid-stream for concatenation/interpolation snippets — inherent to previewing unclosed literals, but the doc comment and tests present the no-revert invariant as unconditional, (3) a minor handoff observation on plain-text preview vs the settled markdown row.

apps/os/src/components/agent-feed.tsx
mmkal: 🤖 Agreed and fixed in f5f6ff9d4: the gate is now `looksLikeCode(step.responseText) || sendMessagePreview != null` — a non-null preview is itself proof the text is code, so the bare/`void`/bare-`Promise.all([` forms the parser supports now actually render their preview (above a code block, not in the prose branch). Kept the parser tolerance rather than narrowing it.

apps/os/src/components/streaming-send-message-preview.ts
mmkal: 🤖 Both taken in f5f6ff9d4: the doc comment now says outright that a preview can appear and then VANISH when a later token invalidates it (with the `"hi" + name` and `` `hi ${name}` `` examples), and a new `distinctPreviews` pinning test asserts the exact appear→grow→null sequences for both cases so a future refactor can't silently "fix" it in either direction. No behavior change — bail-to-null stays.

apps/os/src/components/agent-feed.tsx
mmkal: 🤖 Took the suggestion in f5f6ff9d4: the preview now renders through `MessageResponse` in its default streaming mode (`parseIncompleteMarkdown`), so a markdown-y message previews with its final typography and the handoff to the settled row is just a position change. One compromise: streamdown owns the markdown DOM, so the blinking cursor can't ride the last character — it blinks on its own line below the bubble.

mmkal:
Do a recording pls

mmkal:
🤖 Done — recording added to the PR body under "Demo". It is a real agent turn against local dev: the "I'm going to read the committed `AGENTS.md` file…" bubble streams in above the amber code block while the literal is still unclosed, then the real `web-message-sent` row takes over when the script runs. (Recorded with the browser gif recorder; frames are ~1s apart, so the token-by-token growth shows as a few discrete steps rather than perfectly smooth.)

mmkal:
Let's start only showing the code block preview when we "need" to. That is, don't show it when the first message is incomplete. Reason: it's weird and confusing seeing the message twice.

Also there should probably be a buffer, like only show it when we get the next word character (\w) - that way it won't appear just to add a semicolon, newline and curly brace only to disappear again.

Then when we do show it we should replace it with ellipsis like itx.chat.sendMessage(...)

Note: that's just for the preview obviously when you look at the "Ran code" expandable thing it should show the unmodified js

mmkal:
🤖 Done, all three parts:

1. **Bubble-only while the message streams** — while the first `sendMessage` string literal is still open, the preview bubble renders alone; the streaming code block is suppressed so the same words never show twice.
2. **\\w buffer** — after the closing quote, the block stays hidden through trailing punctuation (`");`, newlines, `}`, the closing fence). It only appears once a word character streams in after the close, i.e. a real second statement is coming — a message-only snippet never flashes a code block in just to end the turn.
3. **Ellipsis** — when the block does appear, it renders the snippet with the message literal collapsed to `itx.chat.sendMessage(...)` (only the first literal; an options arg like `sendMessage("hello", { wha…` stays visible as `sendMessage(..., { wha`). Display-only for the live stream — the settled "Ran code" expandable shows the unmodified js.

The parser now returns `{ message, literalClosed, redactedCode } | null` and the feed shows `redactedCode` when it's non-null (17 unit tests incl. every-prefix assertions that the block only surfaces after the close and never hides again).

Verified end-to-end against local dev — bubble streams alone → `sendMessage(...)` block joins once the count loop starts streaming → settled rows take over. I recorded it, but the GIF upload leg failed repeatedly (browser extension flaked on both the download and the drag-drop-to-textarea path), so the Demo GIF in the PR body still shows the previous behavior — I've marked it as predating this change in the body.

mmkal:
🤖 New recording is in — the Demo section of the PR body now shows the post-change behavior, captured at a phone-width viewport (~450px) with the exact prompt "tell me a 200 word story". What it demonstrates, per the three rules above:

1. the story bubble streams **alone** while the `sendMessage` literal is open — no duplicated words in a code block below;
2. the code block stays hidden through the trailing `");` + `}` — this message-only turn never flashes one in;
3. and when a snippet *does* continue past the close, the live view shows it with the message collapsed to `itx.chat.sendMessage(...)` (the settled "Ran code" expandable keeps the unmodified js).

The old desktop-width GIF is replaced.

PR #1768, opened 2026-07-08
Repo IDE: markdown preview toggle

mmkal:
Markdown files in the repo mini-IDE get a vscode-style **Code | Preview** tab pair in the top left of the file header. Preview renders the **current working-tree buffer** — unsaved and staged edits included, not just HEAD — as sanitized, design-system-styled prose (headings, lists, GFM tables, highlighted code blocks).

## Demo

The markdown Code/Preview toggle, recorded from `specs/repo-ide-markdown-preview.spec.ts` via `VIDEO_MODE=1`:

demo

## How it works

- **No new dependency, no `dangerouslySetInnerHTML`.** Instead of `marked` + DOMPurify, the preview reuses `streamdown` via the existing `MessageResponse` component (`packages/ui/ai-elements/message.tsx`) — the same renderer the agent feed uses for chat. The effective guard is `rehype-sanitize`'s default (GitHub) schema in streamdown's default pipeline (its `rehype-harden` config is permissive), which holds as long as nothing passes `rehypePlugins` — recorded as a code-comment invariant on `MarkdownPreview`.
- **URL-owned view state**: a `preview` search param alongside the IDE's `file`/`diff`/`scm`/`staged`. Preview and diff are mutually exclusive in every direction — the preview/diff toggles clear each other, and opening a file (from the tree or the Source Control panel) resets to Code.
- **Styling fix with reach beyond the IDE**: streamdown's built-in markdown components carry Tailwind classes that were never in our `@source` scan, so chat prose only looked right where classes overlapped with app code. `packages/ui/globals.css` now scans streamdown's dist per its install docs — agent-feed headings/lists/tables benefit too (screenshot below).

## Verification

- Playwright spec `specs/repo-markdown-preview.spec.ts`: seeds a repo with hostile markdown (`<script>document.title = "pwned"</script>`, ``), asserts the prose renders, the payloads don't execute, an **uncommitted** editor change shows up in the preview, and opening the modified file from Source Control shows the diff (not a lingering preview). Passes locally against the dev server.
- Live walkthrough on local dev (`test` project, `/repos/demo`):

Code view (toggle top-left, like vscode):

Preview of the committed README:

Preview of an **unsaved** edit (note the `modified` badge, Discard/Stage still in the header, and the M annotation in the tree):

Agent-feed chat markdown after the `@source` change (assistant message with heading/list/table/code seeded server-side) — confirming the styling side-effect is a fix, not a regression:

Task file: `tasks/repo-ide-markdown-preview.md` (spinoff #1 of the repos mini-IDE task). Sibling spinoffs (HTML renderer, git history, JSON-schema lint, TS language server) remain future tasks. Merge-coordination note: #1767 (HTML preview) adds the same `FileChrome.leading` slot with a different toggle implementation — whoever merges second should unify on one control (see review thread).

| --- | ---: | ---: |
| Product | +2 -2 | +0 -0 ⬜⬜⬜⬜⬜ |
| UI components | +97 -47 | +59 -32 🟩🟩🟩🟥🟥 |
| Tests | +127 -2 | +83 -2 🟩🟩🟩🟩⬜ |
| Docs | +109 -0 | +97 -0 🟩🟩🟩🟩🟩 |
| **Total** | **+335 -51** | **+239 -34** 🟩🟩🟩🟩🟥 |

<sub>Lines counts every changed line; Significant ignores blank lines and JS comments. Between 9f3f9fb and a20ee9a, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>

## Environment Config Lease

<details>

| Auth | released | `12e4fda` | [https://auth.iterate-preview-7.com](https://auth.iterate-preview-7.com) | 811.9 KiB | 16.7s | 591ms |  | 3.0s | [Workflow run](https://github.com/iterate/iterate/actions/runs/37159244836120) | 2026-07-09T08:26:02.085Z | Preview app released. |
| OS | released | `12e4fda` | [https://os.iterate-preview-7.com](https://os.iterate-preview-7.com) | 3.14 MiB (±0 vs main) | 60.1s | 116.0s | 2 retried: DESIRED: a live-capability fetch handler serves WebSockets at an app host (vitest x1) · toggle a markdown file between Code and its rendered Preview (specs x1, still failed) | 14.8s | [Workflow run](https://github.com/iterate/iterate/actions/runs/37159244836120) | 2026-07-09T08:26:13.924Z | Preview app released. |
| Semaphore | released | `4af748c` | [https://semaphore.iterate-preview-7.com](https://semaphore.iterate-preview-7.com) | 440.7 KiB | 15.0s |  |  | 939ms | [Workflow run](https://github.com/iterate/iterate/actions/runs/318950059067233) | 2026-07-09T08:26:00.061Z | Preview app released. |
| Streams Example App | released | `4af748c` | [https://streams.iterate-preview-7.com](https://streams.iterate-preview-7.com) | 1.29 MiB | 14.0s |  |  | 761ms | [Workflow run](https://github.com/iterate/iterate/actions/runs/318950059067233) | 2026-07-09T08:25:59.885Z | Preview app released. |

</details>

---

> [!NOTE]
> **Medium Risk**
> Renders user-supplied repo markdown in the UI; safety depends on streamdown’s default rehype-sanitize pipeline and not overriding `rehypePlugins`.
> 
> **Overview**
> Adds a vscode-style **Code | Preview** control for **markdown** files in the repo mini-IDE (alongside existing HTML preview), driven by a new `isPreviewablePath` predicate and the URL `preview` param.
> 
> **Markdown preview** renders the live working-tree buffer (unsaved edits included) via existing `MessageResponse` / streamdown, with `parseIncompleteMarkdown={false}` for settled files. The toggle moves to design-system **Tabs**; **Diff** is hidden while preview is active. Opening a file from commit history now uses `selectFile` so lingering `preview=true` does not stick.
> 
> **Styling:** `globals.css` adds a Tailwind `@source` scan of streamdown’s dist so preview (and agent-feed chat) markdown classes generate.
> 
> **Tests:** HTML spec selectors switch to `role="tab"`; new Playwright coverage for markdown toggle, hostile markdown sanitization, unsaved-buffer preview, and SCM → diff (preview cleared).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a20ee9a609e6e820ed8c03d7b446c85d37868d85. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


apps/os/src/components/repo-ide/repo-ide.tsx
```diff
@@ -276,6 +277,12 @@ export function RepoIde({ projectId, repoPath }: { projectId: string; repoPath:
                 change={changes.get(selectedPath)}
                 diffOpen={diff}
                 onToggleDiff={(open) => patchSearch({ diff: open ? true : undefined })}
+                previewOpen={preview}
+                // The preview replaces the editor wholesale, so it and the
+                // diff view are mutually exclusive.
+                onTogglePreview={(open) =>
+                  patchSearch({ preview: open ? true : undefined, diff: undefined })
```
mmkal: 🤖 The mutual exclusivity is enforced here and in `selectFile`, but the Source Control panel's handlers just above (`onOpen` at ~L233 and `onOpenStaged` at L240) don't clear `preview`. UI-reachable repro:

1. Open `notes.md`, toggle **Preview** (`?preview=true`)
2. Switch the sidebar to Source Control
3. Click the modified `notes.md` entry — `onOpen` patches `diff: true` but leaves `preview: true`

Result: URL says `diff=true&preview=true`, `showPreview` wins, so the user who clicked a modified file expecting a diff gets the preview instead — with the Diff button hidden (it's gated on `!showPreview`) and the `(Working Tree)` suffix showing. It also contradicts the PR body's "switching files resets to Code", since `onOpen` can switch files too.

`onToggleDiff` (L279) has the same asymmetry — it doesn't clear `preview` — though that one isn't UI-reachable today because the Diff button is hidden while previewing.

Sibling #1767 handled all three spots (`onOpen`, `onOpenStaged`, `onToggleDiff` each clear `preview`) — worth copying that here, both for the bug and to make the eventual merge with that branch a clean union.

mmkal (commented):
🤖 Reviewed the diff, the streamdown dist in the lockfile, the sibling PR #1767, and the spec against docs/testing.md. Overall this is in good shape: no new dependency, no `dangerouslySetInnerHTML`, URL-owned view state consistent with the IDE's existing params, helper at the bottom of the spec, hostile-markdown fixture that also covers the unsaved-buffer case. I verified the sanitization claim against the actual `streamdown@1.6.11` bundle rather than trusting the README — details inline.

Findings, roughly in priority order:

1. **One real bug**: the Source Control panel's open handlers don't clear `preview`, which breaks the PR body's "preview and diff are mutually exclusive" / "switching files resets to Code" claims via a UI-reachable path (inline comment on `repo-ide.tsx`).
2. **Merge coordination with #1767**: both PRs add the same `leading` slot (good) but grew different `CodePreviewToggle`s and different suffix behavior (inline comment).
3. Sanitization claim holds, with one caveat worth having on the record about how little `rehype-harden` contributes at streamdown's default config (inline comment on `MarkdownPreview`).
4. The `@source` addition is per streamdown's docs and the `*.js` glob is actually more correct than the doc's `index.js` (the Tailwind classes live in the code-split `chunk-*.js` files) — but it silently changes agent-feed chat rendering, which deserves an eyeball (inline comment).

Spec notes (no change requested): the negative `page.title()` assertion is soundly ordered for the `<script>` vector — React-inserted scripts execute synchronously at commit, so if sanitization broke, the title would already be "pwned" by the time the heading `waitFor` resolves. The `onerror` vector can't actually fire through React rendering even unsanitized (string event-handler props aren't attached), so that half is belt-and-braces rather than load-bearing — fine.

Also: no CI checks have reported on this branch yet — worth confirming CI runs before merge.

apps/os/src/components/repo-ide/repo-editor-pane.tsx
```diff
@@ -381,6 +405,50 @@ function FileChrome({
   );
 }
 
+/** vscode's "Code | Preview" tab pair for markdown files. */
+function CodePreviewToggle({
```
mmkal, line 420: 🤖 Merge-coordination flag, not a change request for this PR: sibling #1767 (HTML preview) adds the *same* `leading` slot on FileChrome (good — that part will merge as a near-union) but grew a **different** `CodePreviewToggle`: hand-rolled `Button`s in a bordered `span`, props named `previewOpen`/`onTogglePreview`, plus a `(Preview)` header suffix that this PR doesn't add. This one uses the design-system `Tabs` with `preview`/`onChange`.

Whoever merges second should unify on ONE toggle rather than keeping two visually different "Code | Preview" controls in the same file. My vote is this Tabs version: it's a design-system component and gets real `tab` roles for free (this PR's spec locates it via `getByRole("tab", ...)`, which is also the more robust selector). While unifying, consider hoisting the "does this file get a preview?" check into `repo-file-kinds.ts` alongside #1767's `isHtmlPreviewPath` — this PR inlines `kind.language === "markdown"` in two places, and a single predicate (markdown | html → which preview) is where this is obviously heading.

apps/os/src/components/repo-ide/repo-editor-pane.tsx
```diff
@@ -381,6 +405,50 @@ function FileChrome({
...
+  );
+}
+
+/**
+ * Rendered markdown for the current working-tree buffer. Streamdown (already
+ * the agent feed's chat renderer, via MessageResponse) runs raw HTML through
+ * rehype-sanitize + rehype-harden, so user-supplied repo content is safe
+ * without any dangerouslySetInnerHTML.
+ */
+function MarkdownPreview({ markdown }: { markdown: string }) {
```
mmkal, line 455: 🤖 I verified the sanitization claim against the shipped `streamdown@1.6.11` bundle rather than trusting the docs: the default rehype pipeline is `rehype-raw → rehype-katex → rehype-sanitize (default GitHub schema) → rehype-harden`, and since this passes no `rehypePlugins`, the defaults apply. The claim holds.

One caveat worth having on the record: streamdown configures `rehype-harden` essentially wide open — `allowedImagePrefixes: ["*"], allowedLinkPrefixes: ["*"], allowedProtocols: ["*"], allowDataImages: true` — so the layer actually doing the work is `rehype-sanitize`'s default schema (strips `<script>`, event handlers, `javascript:` URLs). The doc comment saying "rehype-sanitize + rehype-harden" is accurate but slightly flattering to harden; the safety property is really "rehype-sanitize default schema, as long as nobody passes `rehypePlugins` here". Fine today — just don't let a future patch casually add `rehypePlugins`.

On the coupling question (chat renderer ↔ repo preview): reasonable trade for now, but note what comes along for the ride from `MessageResponse`: the `memo` comparator only checks `children` (harmless here since the other props are constant), the lazy `Streamdown` import means the first Preview open briefly shows the raw markdown via `PlainMessageResponse` before the real renderer loads, and code blocks get the chat styling including the `max-h-[60vh]` inner scroll — nested scrollbars inside a document preview, which VS Code's preview doesn't do. None of these block merging. If/when the two surfaces want to diverge (heading anchors or relative-link handling here; streaming affordances in chat), the right move is a neutral `Markdown` component in packages/ui that both wrap, rather than the repo IDE continuing to reach into `ai-elements`.

packages/ui/src/styles/globals.css
```diff
@@ -3,6 +3,10 @@
 
 @source "../**/*.{ts,tsx}";
 @source "../../../services/*/src/ui/**/*.{ts,tsx}";
+/* Streamdown's built-in markdown components (headings, lists, tables…) carry
+   Tailwind classes of their own; scan its dist so they exist in the output.
+   https://streamdown.ai — "update your globals.css to include the following" */
+@source "../../node_modules/streamdown/dist/*.js";
```
mmkal, line 9: 🤖 This matches streamdown's install docs, and `dist/*.js` is actually *more* correct than the doc's `dist/index.js` — streamdown 1.6.x is code-split and the Tailwind classes live in the `chunk-*.js` files, which `index.js` alone would miss. The pnpm symlink at `packages/ui/node_modules/streamdown` resolves fine per your live verification.

Two side effects worth being conscious of:

1. **The agent feed changes appearance with this PR.** Chat headings/lists/tables now get streamdown's intended classes instead of whatever happened to overlap with app code. The PR frames this as a fix (agreed), but the PR only has screenshots of the repo preview — an eyeball of an actual agent chat with headings/tables/task lists before merging would confirm nothing regressed there.
2. **CSS bundle growth**: scanning all dist chunks makes every class in streamdown's mermaid, katex, and fullscreen-code-block components a candidate, including for features chat may never render. Modest and acceptable, just not free.

apps/os/src/components/repo-ide/repo-ide.tsx
mmkal: 🤖 Real bug, confirmed and fixed in 7bce5944d: `onOpen`, `onOpenStaged`, and `onToggleDiff` all clear `preview` now, mirroring #1767's shape so the eventual merge is a clean union. Also extended `specs/repo-markdown-preview.spec.ts` with the exact repro path — preview open → Source Control → click the modified file → asserts the `(Working Tree)` diff shows, not the preview. (Selector note for posterity: the activity-strip button's accessible name is its dirty-count badge, so the spec uses `getByTitle("Source control")`.)

apps/os/src/components/repo-ide/repo-editor-pane.tsx
mmkal: 🤖 Agreed on all points, and this PR now clears `preview` in the same three spots as #1767 so the state-plumbing side merges as a union. On the toggle itself: also my vote is the Tabs version (design-system component, real `tab` roles, and this PR's spec already locates it via `getByRole("tab", ...)`). Deliberately NOT unifying from this branch — whoever merges second should keep exactly one toggle and hoist the "which preview does this path get?" predicate into `repo-file-kinds.ts` next to `isHtmlPreviewPath`. Flagged in the PR body as a merge-coordination note.

apps/os/src/components/repo-ide/repo-editor-pane.tsx
mmkal: 🤖 Thanks for verifying against the shipped bundle — that's a stronger basis than the docs. Took the caveat on board in 7bce5944d: the `MarkdownPreview` doc comment now states the invariant precisely (the effective guard is rehype-sanitize's default GitHub schema in streamdown's DEFAULT pipeline; harden is configured wide open; the default only holds while nothing passes `rehypePlugins` — don't add that prop without re-checking sanitization). Agreed on the coupling trade-offs: the lazy-load flash, chat code-block styling, and `max-h-[60vh]` nested scroll all ride along, none blocking; if/when the surfaces diverge, a neutral `Markdown` component in packages/ui that both wrap is the right seam. Not building it in this PR.

packages/ui/src/styles/globals.css
mmkal: 🤖 Eyeballed it as requested: seeded an assistant message with heading/list/GFM table/inline code/code block into a fresh agent stream (`events.iterate.com/agents/web-message-sent`) on local dev and screenshotted the feed — everything renders correctly with the new classes, no regression. Screenshot added to the PR body (last image). On bundle growth: agreed it's not free; the mermaid/katex/fullscreen classes become candidates, but they're a bounded, static set and the alternative (hand-maintaining a safelist) rots. Accepting the modest cost.

apps/os/src/components/repo-ide/repo-ide.tsx
```diff
@@ -597,11 +597,13 @@ function GitPanel({
 
 /**
  * IDE view state, URL-owned like every stream view's: `file` is the open
- * path, `diff` whether the HEAD↔staged diff is showing, `scm`/`gh`/`history`
- * which sidebar shows instead of the file tree (Source Control / GitHub /
- * commit history), `commit` the expanded commit's oid (which also pins the
- * readonly commit diff the open file renders as). The repo detail route
- * validates these (RepoDetailSearch), so loose reads here are safe.
+ * path, `diff` whether the HEAD↔staged diff is showing, `preview` whether a
```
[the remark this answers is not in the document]
mmkal: 🤖 Fixed in 12e4fdaf9 — the History panel's `onOpenFile` now routes through `selectFile`, which clears `diff`/`preview`/`staged` (history/commit stay set so the commit diff keeps showing). A file picked out of a commit no longer inherits a lingering `preview=true`. Same class as the two open-path fixes on #1767.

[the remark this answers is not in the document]
mmkal: 🤖 Good catch, and it is real — but it is a pre-existing bug on `main`, not specific to this PR: it became reachable once #1767 (the `preview` param) and #1769 (the history panel) both merged. The History panel's `onOpenFile` was the only file-open path patching just `{ file }`, while the tree (`selectFile`) and Source Control paths clear the view params.

Fixed on a dedicated branch off main in #1785 (`onOpenFile={selectFile}`), so it lands regardless of this PR's timeline; this branch will pick it up on its next merge from main. Resolving here.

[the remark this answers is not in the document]
mmkal: 🤖 Real concern, but already fixed within this PR before merge. An earlier review round in #1768 flagged the same bug on the Source Control open path; the same treatment was applied to the History `onOpenFile` too. The commit Bugbot reviewed (9f1c5a2) still had `onOpenFile={(path) => patchSearch({ file: path })}`, but the merged version (07e316111, currently on main) reads:

```tsx
// selectFile clears diff/preview/staged too, so a lingering
// preview=true doesn't spuriously re-open Preview for the file
// you pick out of a commit (history/commit stay set — the
// commit diff keeps showing until you leave History).
onOpenFile={(path) => selectFile(path)}
```

`selectFile` patches `{ file, diff: undefined, preview: undefined, staged: undefined }`, so opening a file from History clears `preview` just like tree selection and the SCM path. Nothing to change.

PR #1769, opened 2026-07-08
Repo IDE: git history view (repo.log + commitDetails + readonly commit diffs)

mmkal:
Adds git history to the repo mini-IDE — a "Git Graph"-style commit list with expandable commit details and readonly per-file diffs — plus the itx read lanes it needs.

## Net effect

**itx surface** (three additions to the `Repo` capability):

```ts
const repo = itx.repos.get("/repos/demo");

const { commits } = await repo.log({ limit: 20 });
// [{ oid, message, author: { name, email }, timestamp /* epoch ms */, parents }, …] newest first

const details = await repo.commitDetails({ commitOid: commits[0].oid });
// details.parentOid, details.files:
// [{ path, status: "added"|"modified"|"deleted", additions, deletions, binary }]

await repo.readFile({ path: "src/greet.ts", commitOid: commits[1].oid }); // pinned historic read
```

Per-commit file stats are deliberately **not** part of `log` — they cost a checkout of the commit and its parent per commit, so they're a lazy `commitDetails` call matching the UI's expand-a-row access pattern. Line counts come from a dependency-free Myers line diff (`git diff --numstat` parity, NUL-byte binary sniff, rewrite fallback past a 4000-line budget).

**IDE**: a third activity-strip entry, History. View state is URL-owned like the rest of the IDE (`history`, `commit` params — an expanded commit survives reload and is shareable).

Clicking a changed file opens a readonly parent-vs-commit diff (same lock-icon/no-controls treatment as the staged Index view):

Deletions and binaries:

## Testing & verification

- `line-diff.test.ts` unit tests (numstat parity, trailing newlines, minimal counts when lines move, binary sniff, size fallback).
- `repo-history.itx.e2e.test.ts` drives all three lanes over real commits (add/modify/delete/binary, root commit, pinned reads, validation errors) — passes against a local dev server.
- Live Playwright walkthrough on local dev (screenshots above): list → expand (+/- counts) → modified/deleted/binary diffs → root commit → back to Files.
- `pnpm typecheck && pnpm lint && pnpm format && pnpm test` green; itx-api / types-source / template `sdk.ts` snapshots regenerated (the template snapshot test compares them verbatim).

## Non-goals (explicit)

- **No graph edges** — repos are effectively single-branch, so history renders as a linear list (`parents` is in the API for a future graph).
- No pagination past `limit`, no revert/checkout/reset operations, no rename detection (renames read as delete + add). Details in `tasks/repo-ide-git-history.md`.

## Review round (handled, threads resolved)

- **History freshness**: committing in the IDE now also invalidates the `repo-log` query, so your own commit appears in History immediately (per-commit details stay cached — immutable, keyed by oid).
- **Pinned reads share the read-your-write retry**: `readFile({ commitOid })` of a just-pushed commit used to hard-fail during the Artifacts eventual-consistency window; the pinned `#checkout` path now retries while the clone is behind the recorded push, and still fails fast on genuinely unknown oids.
- **Honest diff-stat docs**: trailing-newline-only changes count 0/0 here where git says 1/1 — deliberate, now documented as a divergence instead of claiming numstat parity.
- **Bounded clone costs**: `log` clones with `depth: limit` (verified isomorphic-git stops its log walk at the `.git/shallow` boundary) and `commitDetails` resolves its oid as a ref instead of walking the whole history.

Also merged `origin/main` (twice): the IDE GitHub panel (#1763) landed alongside — the activity strip now carries Files / Source control / History / GitHub side by side, re-verified live in the browser after the merge.

| --- | ---: | ---: |
| Product | +439 -49 | +275 -36 🟩🟩🟩🟩⬜ |
| UI components | +404 -36 | +357 -29 🟩🟩🟩🟩🟩 |
| Tests | +214 -0 | +175 -0 🟩🟩⬜⬜⬜ |
| Docs | +149 -0 | +129 -0 🟩🟩⬜⬜⬜ |
| Generated | +111 -7 | +77 -5 🟩⬜⬜⬜⬜ |
| **Total** | **+1,317 -92** | **+1,013 -70** 🟩🟩🟩🟩🟩 |

<sub>Lines counts every changed line; Significant ignores blank lines and JS comments. Between 1c87225 and 4b04cb9, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>

## Environment Config Lease

<details>

| Auth | released | `4b04cb9` | [https://auth.iterate-preview-1.com](https://auth.iterate-preview-1.com) | 812.0 KiB | 18.3s | 506ms |  | 5.0s | [Workflow run](https://github.com/iterate/iterate/actions/runs/153499760961907) | 2026-07-09T07:01:48.378Z | Preview app released. |
| OS | released | `4b04cb9` | [https://os.iterate-preview-1.com](https://os.iterate-preview-1.com) | 3.14 MiB (+4.7 KiB vs main) | 66.6s | 130.3s | 1 retried: DESIRED: a live-capability fetch handler serves WebSockets at an app host (vitest x1) | 11.3s | [Workflow run](https://github.com/iterate/iterate/actions/runs/153499760961907) | 2026-07-09T07:01:54.718Z | Preview app released. |

</details>

---

> [!NOTE]
> **Medium Risk**
> Adds new git checkout/log paths on the repo Durable Object (bounded depth for `log`, heavier work for `commitDetails`); read-only and UI-focused, but commit-detail expansion can be costly on large trees.
> 
> **Overview**
> Adds **git history** to the repo mini-IDE and the **itx `Repo` read API** that backs it.
> 
> **API / SDK:** `repo.log` returns a newest-first commit list (metadata and parents, no per-commit file stats). `repo.commitDetails` lazily returns one commit’s changed files with numstat-style +/- counts and binary flags. `readFile` gains optional `commitOid` for pinned historic content. Types are reflected in the project template `sdk.ts`.
> 
> **Server logic:** New `line-diff.ts` computes file change stats (Myers line diff, NUL binary sniff, rewrite fallback past ~4000 lines) for `commitDetails`; trailing-newline-only edits intentionally count 0/0 vs git.
> 
> **IDE:** A **History** activity-strip view lists commits; expanding a row loads `commitDetails`; selecting a changed file opens a readonly parent-vs-commit diff (`CommitDiffPane`). URL search params `history` and `commit` own view state. After a local commit, `repo-log` query invalidation refreshes the list. `FileChrome` / `EmptyPane` are exported for the diff pane.
> 
> **Tests:** Unit tests for `line-diff`; e2e covers `log`, `commitDetails`, pinned `readFile`, and validation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4b04cb93146ebbb6347519f4901681969dc44d8b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


apps/os/src/components/repo-ide/commit-history-panel.tsx
```diff
@@ -0,0 +1,190 @@
...
+  /** The commit whose row is expanded (URL-owned view state). */
+  expandedOid: string | undefined;
+  /** The file open in the diff pane, to highlight its row. */
+  selectedPath: string | undefined;
+  onExpand: (oid: string | undefined) => void;
+  onOpenFile: (path: string) => void;
+}) {
+  const log = useItxQuery({
+    key: ["repo-log", projectId, repoPath],
+    query: (itx) => itx.repos.get(repoPath).log({ limit: 50 }),
```
mmkal, line 32: 🤖 The history list goes stale after an in-IDE commit: the commit mutation's `onSuccess` (repo-ide.tsx) invalidates only `["itx", "repo-files", projectId, repoPath]`, and this `repo-log` key is never invalidated. With the app QueryClient's `staleTime: 60 * 1000`, the flow "open History → switch to Files → commit → click History again within a minute" serves the cached log **without the new commit** and does not background-refetch (`refetchOnMount: true` only refetches stale data).

Cheap fix: also invalidate `["itx", "repo-log", projectId, repoPath]` in the commit `onSuccess`. The `repo-commit`/`repo-commit-file` keys are genuinely immutable and can stay as they are.

apps/os/src/domains/repos/repo-durable-object.ts
```diff
@@ -304,25 +313,28 @@ export class RepoDurableObject extends DurableObject<Env> {
...
   async readFile(input: {
     path: string;
     encoding?: "utf8" | "base64";
+    commitOid?: string;
   }): Promise<{ commitOid: string; content: string; path: string } | null> {
     const path = normalizeRepoFilePath(input.path);
+    if (input.commitOid !== undefined) assertCommitOid(input.commitOid);
     if (input.encoding === "base64") {
-      const { filesystem, head } = await this.#checkout();
+      const { filesystem, head } = await this.#checkout({ commitOid: input.commitOid });
```
mmkal, line 362: 🤖 Pinned reads bypass the read-your-write retry: in `#checkout`, only the branch-head path retries against `repo-pushed-head`; the `commitOid` path clones once and hard-fails if the clone doesn't contain the oid. That was fine while pinned checkouts served worker builds (the pinned oid was always an already-observed head from an earlier clone), but this PR points the UI at pinned reads *immediately after a push*: commit in the IDE → History → expand the new commit (`commitDetails` goes through the retrying branch path, so it succeeds) → click a changed file → `readFile({ commitOid: <just-pushed oid> })`. A stale clone here — the exact Artifacts eventual-consistency behavior `#recordPushedHead` documents as observed in preview e2e — won't contain that commit, and the checkout surfaces a raw error to the diff pane with no retry.

Suggestion: in the pinned path, treat "requested oid missing from / checkout failed in this clone" as retryable the same way "head behind the recorded push" is (you could even reuse the same loop: retry while the clone's branch head is behind `repo-pushed-head`, since a clone that has caught up necessarily contains the new commit).

apps/os/src/domains/repos/line-diff.test.ts
```diff
@@ -0,0 +1,91 @@
...
+  });
+
+  test("counts whole files for creations and deletions", () => {
+    expect(computeLineDiffStats("", "one\ntwo\n")).toEqual({ additions: 2, deletions: 0 });
+    expect(computeLineDiffStats("one\ntwo\n", "")).toEqual({ additions: 0, deletions: 2 });
+    expect(computeLineDiffStats("", "")).toEqual({ additions: 0, deletions: 0 });
+  });
+
+  test("treats a missing trailing newline like git does (same line count)", () => {
+    expect(computeLineDiffStats("a\nb", "a\nb\n")).toEqual({ additions: 0, deletions: 0 });
```
mmkal, line 26: 🤖 This is not what git does, so the test name (and the module docstring's "`git diff --numstat` parity" claim) overstate it. Real git for `"a\nb"` → `"a\nb\n"` reports `1 1` — the final line changed from no-newline to newline:

```
-b
\ No newline at end of file
+b
```

(verified locally with `git diff --numstat`). Two knock-ons of the 0/0 behavior here:

- a trailing-newline-only change produces a `modified` file entry that the history panel renders as `+0 −0`, which reads like a bug;
- the docs claim parity where there is a deliberate deviation.

I actually think 0/0 is a *defensible* choice for a UI stat, but pick one: either count the changed last line like git (treat the no-newline marker as making the final lines unequal), or keep the behavior and fix the comments to say the deviation from numstat is intentional.

FWIW I fuzz-tested `computeLineDiffStats` against a reference LCS DP over 3000 randomized inputs (mixed trailing newlines) and the counts are minimal in every case — the Myers implementation itself is correct. This trailing-newline convention is the only place it diverges from git.

apps/os/src/domains/repos/repo-durable-object.ts
```diff
@@ -333,6 +345,52 @@ export class RepoDurableObject extends DurableObject<Env> {
...
 
+  /**
+   * Commit history of a branch, newest first. One full single-branch clone +
+   * `git log` — deliberately WITHOUT per-commit file stats, which cost a
+   * checkout of every commit and its parent; `commitDetails` computes those
+   * lazily for one commit at a time (the UI's expand-a-row access pattern).
+   */
+  async log(input: { branch?: string; limit?: number } = {}): Promise<RepoLogResult> {
+    const limit = parseLogLimit(input.limit);
+    const { branch, git } = await this.#checkout({ branch: input.branch, fullHistory: true });
```
mmkal: 🤖 Question, not a change request: could `log` clone with `depth: limit` instead of `fullHistory: true`? The backing library is isomorphic-git, which records shallow boundaries in `.git/shallow` and stops `log` there, so a `depth: limit` clone should yield exactly the commits this returns while keeping the sidebar's cost O(limit) instead of O(history) as repos age — and the read-your-write head check works the same on a shallow clone. (The boundary commit's `parents` would still list oids the clone lacks, but this method only reports them as strings anyway.)

Related: `commitDetails` genuinely needs deep history for the parent checkout, so full clone is fair there — but its unbounded `git.log()` walk just to locate the oid also scales with history length; `git.log({ ref: input.commitOid, depth: 1 })` (falling back to the friendly not-on-branch error when it throws) would cap that.

Fine to defer given project repos are small today — flagging it because the History sidebar makes these the hottest read lanes in the IDE (open = 1 full clone, expand = another, file diff = 2 more pinned clones).

mmkal (commented):
🤖 Reviewed the whole diff (backend lanes, line-diff module, codegen artifacts, UI, tests) plus the surrounding DO/itx/IDE code. This is in good shape — the PR body and task file accurately describe what the diff does, and the big design calls (stats lazy on `commitDetails`, not `log`; pinned `readFile` making the diff pane pure composition; URL-owned `history`/`commit` params) are the right ones. Four findings, all inline, severity-ranked:

1. **Stale History after an in-IDE commit** — the commit mutation never invalidates the `repo-log` query, and `staleTime` is 60s, so a commit you just made can be missing from the History panel you open right after. One-line fix. [comment](https://github.com/iterate/iterate/pull/1769#discussion_r3546169569)
2. **Pinned `readFile({ commitOid })` skips the read-your-write retry** — the new commit→history→file-diff flow can hit the documented Artifacts eventual consistency and surface a raw checkout error. Pre-existing pathway, but this PR is what makes it reachable from the UI seconds after a push. [comment](https://github.com/iterate/iterate/pull/1769#discussion_r3546171190)
3. **"numstat parity" is inaccurate for trailing-newline-only changes** — real git reports `1 1` for `"a\nb" → "a\nb\n"`; this module reports `0 0` and the UI would render a modified file as `+0 −0`. Behavior is defensible, docs/test comment are not. [comment](https://github.com/iterate/iterate/pull/1769#discussion_r3546172557)
4. **Question: `depth: limit` shallow clone for `log`** instead of full history, and a `ref`-targeted log in `commitDetails` — cost stays O(limit) as repos age. Fine to defer. [comment](https://github.com/iterate/iterate/pull/1769#discussion_r3546173848)

On the areas I dug into specifically:

- **Myers implementation**: I fuzz-tested `computeLineDiffStats` against a reference LCS DP over 3000 randomized inputs (mixed trailing newlines, empty files) — minimal counts in every case; the d/k seed conditions and the `(N+M−D)/2` LCS arithmetic are correct, no out-of-bounds on the `v` array at the diagonal extremes. Binary sniff matches git's 8000-byte NUL heuristic. The 4000-line budget applies *after* prefix/suffix trim, so large files with small edits still get exact counts — nice. On "should this have been a dependency": the only diff machinery already in the workspace is `@codemirror/merge` (client-side, DOM-adjacent — wrong thing to import into a DO), and there's no jsdiff/fast-diff anywhere in the tree, so ~60 hand-rolled, fuzz-coverable lines is a defensible trade.
- **Generated artifacts**: `itx-api.generated.ts`, `types-source.generated.ts`, and template `sdk.ts` additions are identical modulo backtick escaping in the template-embedded copy — consistent with source.
- **itx surface**: matches the existing `describeNode`/types.ts contract style; input validation (full 40-hex oid, limit 1–200) is exactly right for a public surface.
- **UI conventions**: `history`/`commit` follow the `scm`/`diff`/`staged` pattern and are validated in the route search schema; no useState/useEffect; the readonly pane reuses `FileChrome`/`unifiedMergeView` from the staged view; keying `commitDetails` by oid only (immutable) is a nice touch. e2e covers all three lanes including root commit, binary, delete, pinned reads and validation errors, and reads well.

None of this blocks moving out of draft; I'd want 1 and 2 addressed before merge.

apps/os/src/components/repo-ide/commit-history-panel.tsx
mmkal: 🤖 Fixed in ec521b570: the commit mutation's onSuccess now also invalidates `["itx", "repo-log", projectId, repoPath]`, so the History sidebar picks up your own commit immediately. The per-commit `repo-commit`/`repo-commit-file` queries are deliberately left alone — they key by oid and a commit's diff against its parent is immutable.

apps/os/src/domains/repos/repo-durable-object.ts
mmkal: 🤖 Fixed in ec521b570: the pinned-oid path now shares the read-your-write retry. A failed checkout on a clone whose branch tip is still BEHIND the recorded `repo-pushed-head` is treated as the documented eventual-consistency window and retried with the same budget/backoff as the branch path; a caught-up clone that lacks the oid still fails fast (so a genuinely bogus oid doesn't burn 7.5s of retries). On a test: there's no sensible seam today — the clone dials the real Artifacts remote through `createGit` with no injectable transport, which is the same coverage story as the pre-existing branch-path retry (also untested for the same reason). Recorded a controllable-remote seam as a follow-up in the task file.

apps/os/src/domains/repos/line-diff.test.ts
mmkal: 🤖 You're right — thanks for actually running git on it. Kept the behavior (0/0 for trailing-newline-only changes) and fixed the claims in ec521b570: the module docstring now states the deliberate divergence explicitly, `computeLineDiffStats`'s doc points at it, and the test is renamed to "ignores trailing-newline-only changes (deliberate divergence from git)" with a comment giving git's 1/1 answer. On the UI concern: such a file still renders with the M status letter and an honest `+0 −0`, so it reads as "changed, nothing to count" rather than unchanged — and the readonly diff pane shows the actual newline chunk since CodeMirror's diff is character-based.

apps/os/src/domains/repos/repo-durable-object.ts
mmkal: 🤖 Good call — implemented both in ec521b570 rather than deferring, after verifying the sharp edge you flagged doesn't exist: the vendored isomorphic-git 1.37.6's `_log` reads `.git/shallow` via GitShallowManager and skips the parent walk for shallow commits (`if (!shallowCommits.has(commit.oid))`), so a `depth: limit` clone can't throw on the boundary and the boundary commit's `parents` stay plain strings. `log` now clones with `depth: limit` (O(limit) as history grows), and `commitDetails` resolves its validated 40-hex oid directly via `git.log({ ref, depth: 1 })` instead of the unbounded walk — the single-branch clone only carries branch-reachable objects, so a foreign oid still fails with the friendly not-on-branch error. `#checkout`'s `fullHistory` flag became `historyDepth: number | \"full\"` (commitDetails keeps "full" since it must checkout the parent). Re-verified with the repo-history e2e against local dev.

apps/os/src/domains/repos/line-diff.ts
```diff
@@ -0,0 +1,140 @@
...
+}
+
+/**
+ * `git diff --numstat`-shaped line counts between two texts: the minimal
+ * number of added and removed lines (Myers edit distance over lines, with
+ * common prefix/suffix trimmed first so ordinary edits stay tiny). Not exact
+ * numstat parity — see the module docstring for the deliberate
+ * trailing-newline divergence.
+ */
+export function computeLineDiffStats(
```
mmkal, line 33: Could we/should we use a library for this?

mmkal (commented):
Needs proper review

apps/os/src/domains/repos/line-diff.ts
mmkal: 🤖 Good question — I dug into what's actually available before answering, and my recommendation is to **keep the hand-rolled version**, but here's the honest tradeoff so it's your call.

**What's already in the tree (the "free reuse" candidates):**

- **`diff` (jsdiff)** is in the pnpm store, but only as a *transitive build-time* dep of `@tanstack/router-utils` — it is **not** a runtime dependency of `apps/os`. Using it here would mean adding a genuine new runtime dep to the Durable Object bundle, not reusing something already shipping.
- **`@codemirror/merge`** *is* a real runtime dep and we already use its `presentableDiff` (in `change-gutter.ts`) and `unifiedMergeView` for the client-side diff rendering. But it's **client-only**: it's coupled to `@codemirror/state`/`view`, and `presentableDiff` returns character-range `Change[]`, not line counts. It can't run in the DO where `line-diff.ts` lives.

**Why the hand-roll is justified for this specific spot:**

1. This code runs **server-side in the repo Durable Object** and needs only the `--numstat` +/- **counts**, not an edit script. That's the cheap half of a diff — a Myers O(ND) distance pass, no chunk reconstruction/allocation.
2. It's ~90 lines of actual logic, and it's **fuzz-tested** against the invariant that matters, so the "did we get Myers right" maintenance risk is already pinned.
3. Two behaviors we'd have to keep even if we adopted jsdiff: the `MAX_DIFF_LINES` budget that degrades to "rewrite" instead of stalling the DO on a pathological input (jsdiff has no budget and is also O(ND)), and the deliberate trailing-newline divergence from git. So a swap wouldn't delete much — it'd wrap jsdiff and re-add the guardrails.

**If you'd still prefer a library:** the only clean path is adding `diff` as a direct `apps/os` dep and doing `diffLines(old, new)` then summing `.count` over the `added`/`removed` parts. Behaviorally equivalent for our counts, more allocation (builds the full script we then throw away), and we keep the cap + newline handling on top. Happy to switch if you want the reduced surface — it's ~30 min including re-pointing the fuzz test — but I lean toward the current version given it's counts-only, DO-resident, and already fuzzed.

(One consistency note either way: the client renders diffs via `@codemirror/merge` while the server counts via this — different engines. In practice +/- totals vs. rendered chunks agree for normal edits; the only visible gap is the documented trailing-newline case.)

PR #1770, opened 2026-07-08
Repo IDE: JSON/YAML json-schema validation (red squigglies)

mmkal:
Spinoff 4 from the repos mini IDE (#1759). JSON and YAML files in the repo IDE now get **json-schema validation**: red squigglies, hover docs, and schema-driven autocomplete.

## Demo

A `package.json` with an invalid `name` gets the schema squiggle once schemastore loads (`specs/repo-ide-json-schema.spec.ts`, `VIDEO_MODE=1`):

demo

## What you get

Which schema applies, in priority order:

1. a top-level `"$schema"` prop (JSON) — `{"$schema": "https://www.schemastore.org/tsconfig.json", ...}`
2. a `# yaml-language-server: $schema=<url>` modeline comment (YAML)
3. a well-known filename map: `package.json`, `tsconfig*.json`, `jsconfig*.json`, `pnpm-workspace.yaml`, `.github/workflows/*.yml`

The active schema's name shows as a muted note in the file header (hover it for the full URL). Resolution runs against the live buffer, so adding a `$schema` line takes effect immediately, before any commit.

## Decisions (details in `tasks/repo-ide-json-schema.md`)

- **Egress**: schemas are fetched client-side from schemastore.org (serves CORS `*`), cached in tanstack query with `staleTime: Infinity`. Vendoring was the wrong call — the tsconfig schema alone is 435KB.
- **Placement**: all in `apps/os/src/components/repo-ide/repo-json-schema.ts`, flowing through the existing `codeMirrorExtensions` prop (same pattern as the itx repl's TS extensions). `codemirror-json-schema` drags in markdown-it/shiki, which stays out of `@iterate-com/ui` and lands in the already-lazy repo-IDE chunk.
- **Upstream bug worked around**: @lezer/json 1.0.3 (Dec 2024) added the `:` token to the parse tree, which makes codemirror-json-schema ≤0.8.1 put every squiggly on the *colon* instead of the value. `parseJsonDocumentColonFixed` wraps their parser and shifts those pointers one sibling over. Worth an upstream issue.
- **packages/ui side-fix**: `SourceCodeBlock` now preserves selection+focus when the CodeMirror view is rebuilt due to an extensions change (schema arriving, diff toggling) — previously the cursor reset to the top.

## Testing

- 14 egress-free unit specs (`repo-json-schema.test.ts`): URL resolution policy, and real diagnostics produced from an inline fixture schema via a headless EditorState — exact from/to positions asserted for JSON and YAML.
- Live playwright walkthrough against local dev (screenshots above), including a live-edit pass (squiggly count drops when you fix a value; no cursor loss) and a blocked-egress pass (`**schemastore.org**` aborted).

## Review round (all threads resolved)

- **Schema fetches debounced**: the resolved URL passes through a 500ms settle query (tanstack `keepPreviousData`) before it can reach the network — typing out a `$schema` URL fires zero fetches to garbage prefix hosts, exactly one once stable (verified live with a request listener). Plus a `content.includes('"$schema"')` guard before the per-keystroke `JSON.parse`.
- **SourceCodeBlock focus-loss bug found & fixed**: the selection-preservation in the first commit never actually ran — the effect *cleanup* destroys the view and nulls the ref before the next effect body could capture it, so any extensions change mid-typing dumped focus to `<body>` and silently ate keystrokes. The cleanup now stashes state/focus for the next effect body to restore. Live-verified: typing a full `$schema` line through two schema-association changes keeps every keystroke, the cursor, and focus — ending with the new schema active:

- **Regex fallback anchored** to shallow indent / minified `{"$schema":` form, so nested `$schema`s (vendored schema collections) aren't grabbed mid-edit (+2 unit specs).
- **JSONC** (strict-parse squiggles on comment-bearing tsconfig.json): deliberately not fixed here — staged as #1774 (`repo-ide-jsonc`, based on this branch) using codemirror-json-schema's `/json5` mode.
- Two more upstream pointer potholes (array-element diagnostics on `,`/`]`; object-valued property errors vanishing) documented in the task file for the eventual upstream filing.

## Known limitation

json-schema-library (the validator inside codemirror-json-schema) doesn't descend into the github-workflow schema's `oneOf`-heavy `jobs` subtree, so e.g. `runs-on: 123` is silent while top-level violations squiggle correctly. Flat schemas (package.json, tsconfig, pnpm-workspace) validate deeply and precisely. Upstream validator quality issue — noted as a follow-up in the task file.

| --- | ---: | ---: |
| Product | +8 -0 | +5 -0 🟩⬜⬜⬜⬜ |
| UI components | +336 -3 | +214 -3 🟩🟩🟩🟩⬜ |
| Tests | +191 -0 | +145 -0 🟩🟩🟩⬜⬜ |
| Config | +5 -0 | +5 -0 🟩⬜⬜⬜⬜ |
| Docs | +63 -0 | +50 -0 🟩⬜⬜⬜⬜ |
| Generated | +328 -2 | +255 -2 🟩🟩🟩🟩🟩 |
| **Total** | **+931 -5** | **+674 -5** 🟩🟩🟩🟩🟩 |

<sub>Lines counts every changed line; Significant ignores blank lines and JS comments. Between 757f1d1 and f1f8b1e, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>

## Environment Config Lease

<details>

| Auth | released | `f1f8b1e` | [https://auth.iterate-preview-7.com](https://auth.iterate-preview-7.com) | 811.9 KiB | 17.6s | 599ms |  | 3.0s | [Workflow run](https://github.com/iterate/iterate/actions/runs/553554558542631) | 2026-07-09T09:19:30.434Z | Preview app released. |
| OS | released | `f1f8b1e` | [https://os.iterate-preview-7.com](https://os.iterate-preview-7.com) | 3.14 MiB (+0.0 KiB vs main) | 61.9s | 145.3s | 6 retried: catalogue example "browse-examples" runs identically across runtimes (vitest x1) · agent runs an itx script that appends a proof event, then replies (vitest x1) · DESIRED: a live-capability fetch handler serves WebSockets at an app host (vitest x1) · stream getEvents defaults to a bounded page and supports event type filters (vitest x1) · flags a package.json schema violation with a red squiggle (specs x1) · preview a staged snapshot from the readonly Index view (specs x1) | 10.5s | [Workflow run](https://github.com/iterate/iterate/actions/runs/553554558542631) | 2026-07-09T09:19:37.894Z | Preview app released. |
| Semaphore | released | `6d13622` | [https://semaphore.iterate-preview-7.com](https://semaphore.iterate-preview-7.com) | 440.7 KiB | 22.9s | 6.1s |  | 996ms | [Workflow run](https://github.com/iterate/iterate/actions/runs/400231827175701) | 2026-07-09T09:19:28.440Z | Preview app released. |
| Streams Example App | released | `6d13622` | [https://streams.iterate-preview-7.com](https://streams.iterate-preview-7.com) | 1.29 MiB | 15.0s | 16.3s |  | 1.1s | [Workflow run](https://github.com/iterate/iterate/actions/runs/400231827175701) | 2026-07-09T09:19:28.574Z | Preview app released. |

</details>

---

> [!NOTE]
> **Medium Risk**
> Client-side fetches to user-controlled `$schema` URLs are intentional but increase exposure to third-party schema hosts; changes are mostly IDE/editor UX with substantial new dependency surface in apps/os.
> 
> **Overview**
> Adds **json-schema validation** to the repo IDE for JSON and YAML: lint squiggles, hover docs, and schema-driven autocomplete via `codemirror-json-schema`, wired through new `repo-json-schema.ts` and the existing `codeMirrorExtensions` hook on `SourceCodeBlock`.
> 
> Schema URLs resolve from the live buffer in order: top-level `$schema`, YAML `# yaml-language-server: $schema=…` modelines, then a small well-known filename map (`package.json`, `tsconfig*.json`, GitHub workflows, etc.). Schemas are **fetched in the browser** (schemastore, debounced 500ms), cached with TanStack Query, and failures show a muted “schema unavailable” header note instead of breaking the editor. `RepoEditorPane` applies extensions to the working editor and the readonly staged Index view (each validates its own snapshot).
> 
> **`SourceCodeBlock`** now preserves **selection and focus** when CodeMirror rebuilds on extension changes (e.g. schema load mid-edit). Includes a workaround for upstream diagnostic positions landing on `:` after @lezer/json 1.0.3 (`parseJsonDocumentColonFixed`), 14 unit tests, a Playwright spec, and vitest inlining for `codemirror-json-schema`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f1f8b1e8a1c5cc0ce5b105008edd050bb72d23a7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mmkal (commented):
🤖 Reviewed the diff, the installed packages, and the upstream sources; verified the key claims empirically rather than trusting the PR body. Overall this is solid: the lockfile does resolve @lezer/json 1.0.3, schemastore URLs serve `access-control-allow-origin: *` with 200s (re-checked all three well-known URLs just now), the 12 specs exist and exercise the real diagnostics pipeline with exact positions, hover is unaffected by the colon bug (it uses cursor-based `jsonPointerForPosition`, not the corrupted pointers map), and the task file/PR body describe the diff accurately. New code follows house style (no useState/useEffect, helpers at the bottom of the test file, `toMatchObject`).

Findings, severity-ranked (details inline):

1. **Medium — fetch-per-keystroke while editing a `$schema` URL** (inline on `useQuery`). Each intermediate prefix of the URL is a valid `https://` URL, so typing one out fires a query (with a retry) per keystroke to garbage hosts, each cached as an error forever.
2. **Medium — `jsonParseLinter` squiggles JSONC `tsconfig.json`** (inline). tsconfig with comments is extremely common and it's a flagship entry in the well-known map. Acknowledged in the task follow-ups, but I think it's worth deciding before merge, not after.
3. **Low — colon workaround: verified correct and forward-safe, but two adjacent upstream potholes remain** (inline, with a repro I ran against the installed packages).
4. **Low — regex `$schema` fallback can grab nested / string-literal matches while the doc is invalid** (inline).

Non-blocking observations:

- **SourceCodeBlock change (packages/ui)**: I looked for consumers depending on the old reset-to-top/blur behavior — only `itx-repl.tsx` and the repo IDE pass `codeMirrorExtensions`, and for the repl this is a strict improvement (extensions arriving from the TS worker previously nuked your cursor). The doc-equality guard makes the restored selection always valid, and `previousHadFocus` means focus is only re-grabbed when the editor itself had it at rebuild time, so no focus stealing from other inputs. LGTM.
- **Caching semantics**: `staleTime: Infinity` + `gcTime: 1h` means nothing invalidates a fetched schema within a session except eviction after 60min unused. For schemastore (updated weekly-ish) that's fine; sessions don't outlive schema churn in any way that matters.
- **Egress surface**: the hook is only mounted from `repo-editor-pane.tsx`, so no non-IDE consumer gains schemastore egress. One deliberate-looking consequence worth stating out loud: merely *viewing* a committed file whose `$schema` points at an arbitrary host makes the viewer's browser GET it (credential-less). That matches vscode/yaml-language-server behavior, so I think it's fine, but it's a choice, not an accident — maybe worth one sentence in the module doc.
- **Nit**: in the staged (Index) view the displayed doc is `staged.content` but schema resolution reads the *working* buffer (`working?.type === "write" ? working.content : textBaseline`), so if working and staged disagree about `$schema`, the Index view validates against the working buffer's schema. Contrived, transient, not worth restructuring for — just noting I saw it.

apps/os/src/components/repo-ide/repo-json-schema.ts
```diff
@@ -0,0 +1,208 @@
...
+
+/**
+ * Upstream bug workaround: @lezer/json 1.0.3 (Dec 2024) added the ":" token
+ * to the parse tree, but codemirror-json-schema (<= 0.8.1) still assumes
+ * `PropertyName.nextSibling` is the property's value node — so every
+ * diagnostic's squiggly lands on the colon instead of the offending value.
+ * This wraps the stock parser and shifts such pointers one sibling over.
+ * (YAML mode is unaffected: its pointer walk already skips the ":" node.)
+ */
+export function parseJsonDocumentColonFixed(state: EditorState) {
```
mmkal, line 140: 🤖 I read this against both tree shapes and the installed reality, and the core workaround checks out:

- The lockfile resolves `@lezer/json@1.0.3` (whose changelog entry is "Emit tokens for colons **and commas**"), so the fix targets what actually ships.
- **Forward-compat is graceful**: if upstream codemirror-json-schema stops treating `PropertyName.nextSibling` as the value (or a future lezer stops emitting the token), no pointer slices to a bare `:` and the loop no-ops. The `":"`–with-quotes reasoning for real string values is sound — under the 1.0.3 bug those pointers land on the structural colon anyway, and post-fix they'd slice as three chars.
- The colon-fix test (asserting positions land on `123`) doubles as a regression canary: it passes both with the workaround active and after an upstream fix.

Two adjacent potholes it deliberately(?) doesn't cover — I confirmed both by running the linter with this parser against the installed packages:

1. **Array-of-objects**: `[{"x": 1}, {"y": 2}]` with `items: {type: "string"}` puts the two type errors on the `,` and the `]`. That's the `Object`-token branch of upstream `getJsonPointers` taking `object.nextSibling` — the `,` case is new in 1.0.3, the `]` case existed even on 1.0.2. The right target there would be `keyFrom/keyTo` (the object's own range), not a sibling shift, so this loop's heuristic can't be trivially extended.
2. **Object-valued property errors vanish**: `{"a": {"b": 1}}` with `a: {type: "string"}` produces **zero** diagnostics — upstream's `Object` branch overwrites the pointer entry without `valueFrom/valueTo` and the linter skips it. Not a regression from this PR (same on 1.0.2), and it may be a second contributor to the "github-workflow jobs subtree is silent" limitation alongside json-schema-library's oneOf handling.

Neither blocks this PR, but both belong in the upstream issue the task file promises — worth capturing there so the eventual upstream fix covers all three.

apps/os/src/components/repo-ide/repo-json-schema.ts
```diff
@@ -0,0 +1,208 @@
...
+function jsonDollarSchemaUrl(content: string): string | null {
+  try {
+    const parsed: unknown = JSON.parse(content);
+    const schema =
+      parsed && typeof parsed === "object" && "$schema" in parsed ? parsed.$schema : undefined;
+    return typeof schema === "string" ? httpsUrlOrNull(schema) : null;
+  } catch {
+    // Mid-edit the doc is usually invalid JSON; a cheap regex keeps the
+    // schema association from flickering away while the user types.
+    const match = content.match(/"\$schema"\s*:\s*"([^"]+)"/);
```
mmkal: 🤖 The fallback regex is document-wide, so while the doc is invalid it can match a `"$schema"` that isn't a top-level prop — a nested object's `$schema` (common in files that *contain* schemas, e.g. a vendored schema collection or an openapi doc) or even one inside a string literal. The `nested $schema props do not count when the doc parses` test is honest about the boundary, but the mis-association window is exactly when the user is mid-edit and staring at squigglies from the wrong schema.

It self-heals the moment the doc parses again, so severity is low. If you want to cheaply bias it toward top-level without a real parser: anchor to line start with shallow indent, e.g. `/^\s{0,2}"\$schema"\s*:\s*"([^"]+)"/m` — top-level props in practice sit at indent 0–2, nested ones deeper. Still a heuristic, but it shrinks the false-positive surface a lot for one character class.

apps/os/src/components/repo-ide/repo-json-schema.ts
```diff
@@ -0,0 +1,208 @@
...
+  language: JsonSchemaLanguage | null;
+  content: string;
+}): RepoFileJsonSchema {
+  const { path, language, content } = input;
+  const url = useMemo(
+    () => (language === null ? null : repoFileSchemaUrl({ path, language, content })),
+    [path, language, content],
+  );
+
+  const schemaQuery = useQuery({
```
mmkal, line 218: 🤖 Two things about this query, one of which I'd actually fix:

1. **Typing out a `$schema` URL fires a fetch per keystroke.** `url` is recomputed from the buffer on every change, and almost every prefix of a real URL (`https://w`, `https://ww`, …) passes the `https://` check — so each keystroke mints a new queryKey, fires a fetch (plus `retry: 1`), and caches an error entry forever (`staleTime: Infinity`). It's all failure-tolerant so nothing breaks, but it's DNS/egress spam to garbage hosts and a growing error cache for the session. A small debounce on `url` before it reaches `useQuery` (or only enabling the query once the url has been stable for ~500ms) would kill this cleanly.
2. Related micro-nit: `repoFileSchemaUrl` runs `JSON.parse` on the whole buffer per keystroke for every JSON file, schema-associated or not. A `content.includes('"$schema"')` guard before the parse would skip the work for the overwhelmingly common case (and the filename map needs no content at all).

Also stating the egress model out loud since it's user-controlled: a committed file's `$schema` makes any *viewer's* browser GET an arbitrary https URL (credential-less, CORS-gated). That's what vscode/yaml-language-server do too, so I think it's the right call — but a one-line note in the module doc that this is intentional would save the next reader the security double-take.

apps/os/src/components/repo-ide/repo-json-schema.ts
```diff
@@ -0,0 +1,208 @@
...
+ * those include the `json()`/`yaml()` language extensions which
+ * SourceCodeBlock already provides.
+ */
+function jsonSchemaCodeMirrorExtensions(input: {
+  language: JsonSchemaLanguage;
+  schema: JSONSchema7;
+}): SourceCodeBlockExtension {
+  if (input.language === "json") {
+    return [
+      linter(jsonParseLinter()),
```
mmkal, line 115: 🤖 `linter(jsonParseLinter())` is strict JSON, but `tsconfig.json`/`jsconfig.json` — flagship entries in the well-known map — are JSONC in real life: comments and trailing commas are everywhere. Those files will now carry a permanent syntax-error squiggle at the first comment. (The *schema* linter is fine on them — I checked that `best-effort-json-parser`, which `parseJSONDocumentState` falls back to, parses `//` comments cleanly, so schema validation still produces correct data.)

The task file lists jsonc/json5 handling as a follow-up, and the full fix (a jsonc language lane via their `/json5` mode) is genuinely out of scope. But the narrow regression is avoidable today: skip `jsonParseLinter` (keep the schema linter) when the schema came from the tsconfig/jsconfig map entries, or for `.jsonc` files (which `repo-file-kinds.ts` already maps to `json`). Worth deciding deliberately before merge rather than shipping red marks on the most common config file in every repo.

mmkal: 🤖 The full jsonc lane now exists: #1774 (stacked on this PR) gives `*.jsonc`, `tsconfig*.json`/`jsconfig*.json`, and `.vscode/*.json` a comment/trailing-comma-tolerant language via `codemirror-json5` + `codemirror-json-schema/json5` — schema validation, hover, and autocomplete keep working, plain `.json` stays strict. If this PR merges first with red marks on commented tsconfigs, #1774 removes them right behind it; the narrow skip-`jsonParseLinter` workaround suggested here would just be churn against that.

apps/os/src/components/repo-ide/repo-json-schema.ts
mmkal: 🤖 Fixed in 0db6a8b74, all three parts:

1. **Fetch-per-keystroke**: the resolved URL now passes through a 500ms "settle" query before it can reach the network — built on tanstack (`keepPreviousData` holds the last settled url while the new one is quiet-period-pending) since this repo avoids `useState`/`useEffect`. Verified live with a request listener: retyping an entire `$schema` line fires **zero** requests to prefix hosts and exactly one request once the URL is stable. On error-cache poisoning: `staleTime` only governs *data* freshness, and an errored query has none, so reopening the file refetches — errored keys were never actually pinned forever, and now only settled URLs can even error.
2. **`JSON.parse` per keystroke**: added the `content.includes('"$schema"')` guard.
3. **Egress model**: documented at the fetch site (hook doc) — the viewer's browser GETs the user-committed URL, https-only, credential-less, CORS-gated, response only ever parsed as a schema; intentional vscode/yaml-language-server parity.

No unit spec for the debounce: it's pure query/react lifecycle wiring with no seam that doesn't require a DOM/renderHook lane (apps/os vitest is node-env), so it's covered by the live playwright pass instead.

Bonus find while verifying this live: the *reason* my earlier "preserve selection across rebuilds" change didn't make URL-typing safe is that it never ran — the effect **cleanup** destroys the view and nulls the ref before the next effect body captured it, so any mid-typing extension change dumped focus to `<body>` and silently ate subsequent keystrokes. The cleanup now stashes state/focus for the next effect body to restore; typing through two schema-association changes keeps every keystroke, cursor, and focus.

apps/os/src/components/repo-ide/repo-json-schema.ts
mmkal: 🤖 Agreed on the regression, and the fix is already staged rather than squeezed in here: #1774 (branch `repo-ide-jsonc`, stacked on this PR) adds the comment/trailing-comma-tolerant lane via codemirror-json-schema's `/json5` mode, which covers `.jsonc` files and the tsconfig/jsconfig family. Keeping it out of this PR keeps the diff reviewable; the "decide deliberately before merge" concern is satisfied by merge ordering being Misha's call with the fix sitting on top. Cross-referenced in the task file (`tasks/complete/2026-07-08-repo-ide-json-schema.md`, follow-ups section).

apps/os/src/components/repo-ide/repo-json-schema.ts
mmkal: 🤖 Both potholes captured in the task file's upstream-filing notes (`tasks/complete/2026-07-08-repo-ide-json-schema.md` → follow-ups), alongside the colon regression, with your verification details: (1) array-of-objects diagnostics landing on `,`/`]` — `,` new in @lezer/json 1.0.3, `]` present on 1.0.2, right fix is the object's own `keyFrom/keyTo` rather than a sibling shift; (2) object-valued property errors vanishing via the `Object`-branch pointer overwrite (also pre-1.0.3, and plausibly a second contributor to the github-workflow `jobs` silence). The eventual upstream issue/PR should cover all three; no code change here since neither is a regression from this PR and the heuristic can't be trivially extended to (1).

apps/os/src/components/repo-ide/repo-json-schema.ts
mmkal: 🤖 Agreed it's cheap and safe — implemented in 0db6a8b74. The fallback is now `/^(?:[ \t]{0,3}|\{[ \t]*)"\$schema"[ \t]*:[ \t]*"([^"]+)"/m`: shallow-indent line starts for pretty-printed docs, plus a `{`-prefixed alternative so the minified `{"$schema": …` form still matches. (Two exclusive alternatives on purpose — a naive `[ \t]{0,3}\{?[ \t]*` backtracks its way into matching 6-space-deep nested props, which the new unit spec caught immediately.) Two specs added: deeply-indented nested `$schema` is ignored mid-edit; minified top-level form still matches.

apps/os/src/components/repo-ide/repo-editor-pane.tsx
```diff
@@ -82,6 +83,21 @@ export function RepoEditorPane({
...
+  // itself is fetched from schemastore/wherever by the browser and failure
+  // just means no squigglies.
+  const jsonSchema = useRepoFileJsonSchema({
+    path,
+    language:
+      kind.kind === "text" && (kind.language === "json" || kind.language === "yaml")
+        ? kind.language
+        : null,
+    content: working?.type === "write" ? working.content : (textBaseline ?? ""),
+  });
```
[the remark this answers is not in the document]
mmkal: 🤖 Fixed in 9478160ea. The Index view now resolves its schema from the STAGED snapshot via a second `useRepoFileJsonSchema` (keyed on staged content), so a working-tree-only `$schema` change can't misvalidate the staged buffer; the Index header names that schema (`stagedSchemaNote`).

apps/os/src/components/repo-ide/repo-json-schema.ts
```diff
@@ -0,0 +1,251 @@
...
+  });
+
+  const schema = schemaQuery.data;
+  const extensions = useMemo(
+    () =>
+      language !== null && schema !== undefined
+        ? jsonSchemaCodeMirrorExtensions({ language, schema })
+        : NO_EXTENSIONS,
+    [language, schema],
+  );
```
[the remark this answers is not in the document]
mmkal: 🤖 Fixed in 9478160ea — added `placeholderData: keepPreviousData` to `schemaQuery`, so when the settled URL changes the prior schema's data (and its squigglies) stay applied until the new fetch resolves, matching the settle step. No more mid-switch blink.

apps/os/src/components/repo-ide/repo-json-schema.ts
```diff
@@ -0,0 +1,255 @@
...
+
+  if (url === null) return { status: "none", extensions: NO_EXTENSIONS };
+  // Mid-settle (url !== settledUrl) the previous schema keeps applying — the
+  // header note names what's actually validating the buffer.
+  if (schema !== undefined && settledUrl !== null) {
+    return { status: "active", url: settledUrl, extensions };
+  }
+  if (schemaQuery.status === "error" && settledUrl !== null) {
+    return { status: "unavailable", url: settledUrl, extensions };
```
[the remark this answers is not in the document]
mmkal: 🤖 Good catch — that was a consequence of the keepPreviousData change. Fixed in f1f8b1e8a: the `schemaQuery.status === "error"` check now runs BEFORE the active branch and returns `unavailable` with no extensions, so a failed fetch for a new URL drops the squigglies instead of silently linting against the stale placeholder schema. keepPreviousData still smooths a *successful* refetch (pending, not error).

apps/os/src/components/repo-ide/repo-editor-pane.tsx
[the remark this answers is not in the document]
mmkal: 🤖 Already addressed before this PR merged. The readonly Index view now resolves its schema from a separate hook — `stagedJsonSchema = useRepoFileJsonSchema({ content: staged.content })` — and renders `stagedDiffExtensionsWithSchema` with `stagedSchemaNote`, so the staged snapshot validates against the schema *it* declares, not the working buffer’s (which may have a diverged `$schema`). The reviewed commit 6d13622 was squashed out; the final fixed code is in merge 5b8a3c4. Verified on main.

apps/os/src/components/repo-ide/repo-json-schema.ts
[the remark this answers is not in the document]
mmkal: 🤖 Already addressed. `schemaQuery` carries `placeholderData: keepPreviousData`, so when the settled URL changes the previous schema’s data (and thus its lint extensions) stays applied through the in-flight fetch instead of reverting to `NO_EXTENSIONS` — squigglies no longer blink off mid-fetch. Reviewed commit was squashed out; final code is in merge 5b8a3c4. Verified on main.

apps/os/src/components/repo-ide/repo-json-schema.ts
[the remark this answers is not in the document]
mmkal: 🤖 Already addressed, and deliberately reconciled with the keepPreviousData behavior from the sibling thread. The hook checks `schemaQuery.status === "error"` *before* the active branch and returns `status: "unavailable"` / `NO_EXTENSIONS`, so a failed fetch for a new URL clears the stale schema instead of masking the failure as active. Net semantics: keepPreviousData holds the prior schema only while pending; on error it drops to no-squigglies. Reviewed commit 9478160 was squashed out; final code is in merge 5b8a3c4. Verified on main.

PR #1771, opened 2026-07-08
Repo IDE: TypeScript language server (diagnostics, hover, autocomplete)

mmkal:
The repo mini-IDE now has a real TypeScript language service: red squigglies, hover type info, and autocomplete (with doc strings) for `.ts`/`.tsx`/`.js`/`.jsx` buffers — multi-file aware, so relative imports between repo files resolve and an edit in one buffer immediately affects diagnostics in the others. No monaco: it's the itx REPL's proven `@valtown/codemirror-ts` + `@typescript/vfs`-in-a-web-worker setup, extended from one fixed file to the whole repo working tree.

## What you get

Open any TypeScript file in `/repos/*`:

- **Diagnostics** with vscode-style tooltips, honoring type-level options from the repo's own `tsconfig.json` (the `'wrong' is declared but its value is never read` line below comes from the demo repo's `noUnusedLocals: true`):

- **Hover type info**, including symbols imported from other repo files:

- **Autocomplete** with members of an interface defined in a different file:

- **Cross-file awareness**: here `greet.ts`'s signature was edited (uncommitted, working tree only — note the `M`) and the *untouched* `main.ts` buffer immediately flags the now-wrong call, with the updated signature in the tooltip:

## How it works

- `repo-typescript.worker.ts` — one web worker per repo hosting a `@typescript/vfs` virtual TS environment (default lib files from the TypeScript CDN), with bundler-flavored compiler options merged with a whitelist of type-level options parsed from the repo's `tsconfig.json`.
- `repo-typescript.ts` — a host-side session seeds the worker from HEAD (capped readFile fan-out) overlaid with the in-browser working tree, then keeps the vfs in sync by **reconciliation**: it subscribes to the IDE's `WorkingTreeStore`, diffs desired-vs-pushed contents, and pushes updates/deletes. Keystrokes, file creates, deletes, discards, and renames are all the same code path. Commits keep the warm worker and just resync HEAD. A module-level registry keeps at most one live worker — navigating between repos terminates the old one.
- `RepoEditorPane` attaches the extension bundle (linter, hover, autocomplete-with-docs, facet) to editable buffers in both plain and inline-diff modes; the readonly Index view stays plain.

Live verification shook out three integration landmines worth knowing about (all fixed here): TS lib entries with dots (`dom.iterable`) parse as filenames and blow up env creation; React 19's dev-mode perf logging stringifies props and `String(comlinkProxy)` **throws**, aborting the whole commit (the facet now carries a plain delegating facade, never the proxy); and the IDE's lazy codemirror imports triggered a mid-session vite dep re-optimization that split `@codemirror/state` into two instances (now pre-bundled via `optimizeDeps.include`, which also hardens the REPL's identical pattern).

## What typm adds on top (stacked follow-up)

External npm imports are deliberately untyped (`any`) here, via a `declare module "*"` ambient wildcard. The seam for the `typm` spinoff: the worker's `setFiles`/`deleteFiles` write arbitrary paths into the same vfs, and resolution is real (Bundler mode) with the wildcard as lowest-priority fallback — so typm just drops `.d.ts` trees under `/node_modules/...` and they win automatically. That also naturally retires two shims: the TS2347 ("type args on untyped calls") diagnostic filter and the permissive global JSX namespace. Full notes in `tasks/repo-ide-typescript-lsp.md`.

| --- | ---: | ---: |
| Product | +14 -1 | +9 -1 🟩⬜⬜⬜⬜ |
| UI components | +632 -4 | +413 -3 🟩🟩🟩🟩🟩 |
| Tests | +81 -0 | +66 -0 🟩⬜⬜⬜⬜ |
| Docs | +85 -0 | +66 -0 🟩⬜⬜⬜⬜ |
| **Total** | **+812 -5** | **+554 -4** 🟩🟩🟩🟩🟩 |

<sub>Lines counts every changed line; Significant ignores blank lines and JS comments. Between 6a8c556 and 7b17098, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>

## Environment Config Lease

> [!CAUTION]
> This PR is a draft, so it doesn't claim a preview slot. To get previews: add the `preview` label, mark the PR ready for review, or dispatch the Cloudflare Previews workflow for a one-off run.

<details>

No preview apps recorded.

</details>


apps/os/src/components/repo-ide/repo-typescript.ts
```diff
@@ -0,0 +1,296 @@
...
+      }
+      const reads = await Promise.all(
+        seedPaths.map(async (path) => [path, (await repo.readFile({ path }))?.content] as const),
+      );
+      this.#headContents = new Map(
+        reads.flatMap(([path, content]) => (typeof content === "string" ? [[path, content]] : [])),
+      );
+      this.#store = workingTreeStore({ ...this.input, commitOid });
+      this.#unsubscribe?.();
+      this.#unsubscribe = this.#store.subscribe(() => this.#reconcile());
```
mmkal: 🤖 There's a terminate-vs-ensureSynced race here that leaks a subscription which later **throws inside the store's listener loop**:

1. Open repo A → `ensureSynced` starts, awaiting the `listFiles`/`readFile` fan-out (seconds on first seed).
2. Navigate to repo B → the registry evicts A: `terminate()` runs `#unsubscribe?.()` (still `null`), releases the proxy, terminates the worker, and drops A from the map.
3. A's still-in-flight `ensureSynced` continues past its awaits and executes this line — subscribing `#reconcile` to the module-level (permanent) store A, on a session nothing will ever `terminate()` again.

Now navigate back to repo A (a fresh session A2 gets created) and type: store A's `#commit` iterates its listeners in insertion order, the leaked listener runs first, `#reconcile` computes updates and calls `this.#remote.setFiles(...)` — and comlink throws **synchronously** on a released proxy (`"Proxy has been released and is not useable"`, comlink's `throwIfProxyReleased`). That exception aborts the listener loop, so listeners registered after it — A2's reconcile and the UI's `useSyncExternalStore` subscriptions — never fire for that change. The visible symptom would be the IDE silently no longer reflecting edits, which is a rough one to debug.

The window is only the seed fan-out, but rapid back-and-forth between repos while a big repo seeds makes it reachable. Cheapest fix I can see: a `#terminated` flag set in `terminate()`, checked in `ensureSynced` after each await (bail out before subscribing) and at the top of `#reconcile` (drop the push instead of throwing). `WorkingTreeStore#commit` looping listeners inside a try/catch per listener would also defang the blast radius, but the session-side guard is the actual fix.

apps/os/src/components/repo-ide/repo-typescript.ts
```diff
@@ -0,0 +1,296 @@
...
+      const entry = effectiveEntry(change);
+      if (entry === undefined) continue;
+      if (entry.type === "write") desired.set(path, entry.content);
+      else if (entry.type === "delete") desired.delete(path);
+      // write-base64 entries are binary uploads — nothing for the program.
+    }
+    return desired;
+  }
+
+  #reconcile(): void {
```
mmkal, line 114: 🤖 Keystrokes that land during the seed window are silently dropped from the vfs, because this subscription goes live before `initializeRepo` resolves and `#reconcile` unconditionally records what it pushed:

1. `ensureSynced` subscribes (line 110), computes `desired`, then awaits `#remote.initializeRepo(...)` — whose worker side awaits `createDefaultMapFromCDN` (a network fetch, seconds cold).
2. User types meanwhile (the editor is fully usable — extensions attach later, but the store is shared with the plain editor). The store fires, `#reconcile` runs, `setFiles` reaches the worker — where `createWorker`'s `updateFile` does `if (!env) return;` and **discards it**.
3. Host-side `#pushed` now claims that content is in the vfs, so once the env exists, the diff never re-sends it. That buffer's vfs content stays at its seed-time value — wrong squiggly positions, stale cross-file signatures — until the *next* edit of that same file happens to replace it.

Self-heals per file on the next keystroke, but the first diagnostics a user sees after typing-during-seed are wrong. Fix options: subscribe only after `initializeRepo` resolves and run one `#reconcile()` immediately (mirrors what the `#initialized` else-branch already does), or early-return from `#reconcile` while `!this.#initialized` *without* updating `#pushed`. Worker-side queueing of `setFiles` until env-ready would also work but is more moving parts.

apps/os/src/components/repo-ide/repo-typescript.ts
```diff
@@ -0,0 +1,296 @@
...
+
+  /** Seed on first call; on later calls re-sync HEAD contents when a commit
+   * moved the oid (the worker and its warm language service survive). */
+  ensureSynced(itx: ItxReactHandle, commitOid: string): Promise<void> {
+    const run = this.#chain.then(async () => {
+      if (this.#syncedCommitOid === commitOid) return;
+      const repo = itx.repos.get(this.input.repoPath);
+      const { paths } = await repo.listFiles();
+      const seedPaths = paths.filter(isTypeScriptSeedPath).sort();
+      if (seedPaths.length > MAX_SEED_FILES) {
```
mmkal: 🤖 The cap has an unfortunate interaction with the sort: `seedPaths.sort()` is lexicographic, and `tsconfig.json` starts with `t`, so it sorts **after** every path under `app/`, `lib/`, `src/`, etc. In a >500-file repo, the root tsconfig is among the first things truncated — and then `desired.get("tsconfig.json")` quietly hands the worker `null` and the repo's own compiler options (the headline stretch feature of this PR) silently revert to the defaults, with only the generic count warning in the console.

Cheap fix: pin `tsconfig.json` (and arguably the open buffer's path) into the seed before applying the cap. Something like filtering it out, truncating, then unshifting it back.

apps/os/src/components/repo-ide/repo-editor-pane.tsx
```diff
@@ -72,16 +73,31 @@ export function RepoEditorPane({
   // snapshot when one exists, else HEAD.
   const textBaseline = staged?.type === "write" ? staged.content : (headContent ?? undefined);
 
+  // TypeScript language service (diagnostics, hover, autocomplete) for
+  // ts/tsx/js/jsx working-tree buffers — empty for everything else, and for
+  // the readonly Index view (an inspection surface, not a live buffer).
+  const typeScriptExtensions = useRepoTypeScriptExtensions({
```
mmkal, line 90: 🤖 Heads-up on a UX cost that's easy to miss from the code: `SourceCodeBlock`'s inner `CodeMirror` component **destroys and recreates the whole `EditorView`** whenever the extensions array identity changes (its effect deps are `[editable, extensions]`, and it does `viewRef.current?.destroy(); new EditorView(...)`). Content survives (it re-seeds from the store via `valueRef`), but cursor position, scroll, focus, and **undo history** do not. With this hook in the chain that now happens:

1. **Once per file open**, when the TS query resolves a beat after mount — `data` flips `undefined → object`, `typeScriptExtensions` gets a new identity, `editorExtensions` recreates, view rebuilt. Warm sessions resolve fast so it usually lands before the user types, but on the cold path (dev: "tens of seconds" per the task file) it'll yank focus/undo mid-typing.
2. **Twice per commit**: the commit-oid in the queryKey changes → new cache entry → `data` is `undefined` (squigglies vanish) → resolves → second rebuild. The pre-existing deps (`textBaseline`/`headContent`) already rebuilt the view on commit, so the *marginal* cost is one extra rebuild plus the diagnostics gap — but it compounds the existing undo-loss-on-commit behavior rather than just riding along.

Not blocking — but `placeholderData: keepPreviousData` on the query would bridge the commit flip (the facade and module objects are reference-stable across fetches, so the previous data is still perfectly valid while `ensureSynced` re-runs), and it composes with the short-circuit in `ensureSynced`. The once-per-open rebuild is harder to avoid without reworking `SourceCodeBlock` to reconfigure via a `Compartment` instead of recreating the view — which might be worth a follow-up task in its own right, since every extension-changing feature will keep paying this.

mmkal (commented):
🤖 Reviewed the whole diff plus the surrounding repl code (`itx-repl.tsx`, `itx-repl-typescript.worker.ts`, `createWorker` in `@valtown/codemirror-ts`, comlink's released-proxy behavior, and `SourceCodeBlock`'s CodeMirror lifecycle). The architecture is sound and the PR body/task file describe the diff accurately — the reconciliation really does cover keystrokes/creates/deletes/renames/discards in one path (I traced rename-while-dirty, discard-all, and commit-that-deletes-a-file; the last has only a transient stale window that self-heals on HEAD resync), `terminate()` does unsubscribe, JSONC tsconfigs are handled correctly via `ts.parseConfigFileTextToJson` (comments/trailing commas fine, malformed → defaults), and the facade covers every React-visible flow of the worker in the IDE. No blocking issues, but a few real ones worth fixing, severity-ranked:

1. **Medium — terminate-mid-seed race leaks a store subscription that later throws inside the store's listener loop** (inline comment on `ensureSynced`'s subscribe line).
2. **Medium-low — keystrokes during the seed window are silently dropped from the vfs**: `#reconcile` runs (and updates `#pushed`) before `initializeRepo` resolves, while the worker's `updateFile` no-ops on a missing env (inline comment on `#reconcile`).
3. **Medium-low — extension identity churn recreates the whole EditorView**: `SourceCodeBlock`'s CodeMirror effect destroys/recreates the view when the extensions array changes, so the TS query resolving (once per file open) and the commit-oid key flip (twice per commit) each cost cursor/scroll/focus/undo (inline comment in `repo-editor-pane.tsx`).
4. **Low — the 500-file seed cap silently drops `tsconfig.json`** in big repos because it sorts lexicographically after `src/**` (inline comment).
5. **Low — the TS2347 filter is slightly broader than its stated intent** and there's a narrower seam already available (inline comment in the worker).

Non-inline notes:

- **The REPL still carries the raw comlink proxy in React state**: `useReplTypeScriptExtensions` in `itx-repl.tsx` does `setExtensions([tsFacetWorker.of({ path, worker: remoteWorker }), …])` — the exact shape this PR's task file warns about ("never let the comlink proxy become reachable from React-visible values"). If the IDE crashed on this, the REPL is a latent instance of the same bug. Worth either porting `workerFacade` over (it's generic enough) or noting in the task file why the REPL is exempt.
- **Lifecycle**: leaving `/repos` entirely (to the dashboard, say) keeps the worker + TS program + store subscription alive until some *other* repo's session is acquired — there's no route-unmount teardown at all. That's consistent with the `workingTreeStore` pattern and keeps the service warm for back-navigation, so I think it's fine, but a full TS program is a much heavier thing to keep warm than a Map; worth a sentence in the module docstring saying it's deliberate. Two editors on *different* repos can't coexist in one SPA (single route), and two browser tabs are separate realms, so the single-session registry can't strand or double-terminate in normal use — the only hole is the mid-seed race in the inline comment.
- **Perf shape** (for the record, all reasonable): a keystroke costs one desired-map rebuild (copy of a ≤500-entry map; unchanged entries compare by reference) + one full-file postMessage + an incremental LS update; lints are debounced by `@codemirror/lint`; the seed reads every TS-relevant file upfront over the multiplexed capnweb socket (`Promise.all` over ≤500 reads — unbounded concurrency but one pipe, fine). `getAutocompletionWithDocs` calling `getCompletionEntryDetails` per option on every 0-delay completion request is the priciest thing here, but it's inherited from the REPL unchanged.
- **tsconfig `extends` is silently ignored** — a repo whose real options live in an extended base gets the defaults with no signal. Fine for v1, but deserves a line in the task file's follow-ups next to "live tsconfig edits".
- **vite `optimizeDeps.include` additions**: additive, and the reasoning (pre-bundle the lazily-imported codemirror family so a mid-session re-optimization can't split `@codemirror/state`) checks out; it genuinely also protects the REPL's identical lazy-import pattern. No concerns.
- **Conventions/tests**: no `useState`/`useEffect` in the new code (tanstack query + `useMemo`, per the house rules). No tests — consistent with the rest of `repo-ide/`, but `#desired`/`#reconcile` is the one genuinely test-shaped seam here (pure map-diff logic given a fake store + recorded remote calls); if any of the inline fixes land, a small vitest spec for the seed/reconcile ordering would pay for itself.

apps/os/src/components/repo-ide/repo-typescript.worker.ts
```diff
@@ -0,0 +1,227 @@
...
+
+/**
+ * TS2347 "Untyped function calls may not accept type arguments": with every
+ * bare npm import typed `any` (see the prelude), type arguments on external
+ * calls — `kv.get<number>(…)` against the repo sdk's runtime types — are
+ * guaranteed noise, not signal. Real acquired types (typm) make the calls
+ * typed and the code stops firing on its own; drop the filter then if it
+ * bothers anyone.
+ */
+const IGNORED_DIAGNOSTIC_CODES = new Set([2347]);
```
mmkal: 🤖 Two thoughts on the TS2347 filter's scope:

1. It's slightly broader than the stated intent. The rationale is "bare npm imports are `any`, so type args on external calls are guaranteed noise" — but TS2347 also fires on type args passed to *untyped local* functions, e.g. a plain `.js` file in the repo (`allowJs: true` is in the defaults) called from a `.ts` file with `someJsHelper<number>(x)`. That's a real, actionable diagnostic typm will never fix, and it's suppressed with no trace.
2. There's already a narrower seam for this: `tsLinterWorker({ diagnosticCodesToIgnore: [2347] })` on the host side (or injecting the code in `workerFacade.getLints`, which already carries a `diagnosticCodesToIgnore` parameter through). Same effective behavior for everything that currently surfaces diagnostics, but the language service itself stays honest — so a future Problems-panel/whole-repo-diagnostics consumer (explicitly listed as a follow-up in the task file) doesn't silently inherit the filter.

Since the whole filter is documented as temporary-until-typm either way, this is a judgment call, not a demand — but the host-side version is also one line less magical than monkey-patching `getSemanticDiagnostics`. (The permissive global `JSX` namespace, by contrast, looks correctly scoped to me: it only matters while react types can't resolve, real module-scoped JSX types win once typm lands, and a repo declaring its own global JSX merges rather than breaks.)

apps/os/src/components/repo-ide/repo-typescript.ts
mmkal: 🤖 Agreed — nasty one, and your listener-starvation analysis checks out (the store's `#commit` loop has no per-listener isolation, so one released-proxy throw would starve every later subscriber). Fixed in 1f9b31343 with the `#terminated` flag exactly as suggested: set first thing in `terminate()`, checked at the start of the chained `ensureSynced` body, again after the seed fan-out awaits (before any store contact), after `initializeRepo`, and at the top of `#reconcile`. The subscription can no longer be registered by a dead session, and even a straggler reconcile drops the push instead of throwing. I left `WorkingTreeStore#commit` alone per your note that the session-side guard is the actual fix — wrapping listeners in try/catch there would paper over exactly this class of bug next time.

apps/os/src/components/repo-ide/repo-typescript.ts
mmkal: 🤖 Agreed — the worker's pre-env `updateFile` no-op plus `#reconcile` unconditionally advancing `#pushed` meant mid-seed keystrokes were recorded as delivered but never were. Fixed in 1f9b31343 with your first option: the subscription now attaches only after `initializeRepo` resolves, followed by one immediate `#reconcile()` — the diff against the seed-time `#pushed` snapshot picks up anything typed during the awaits. This also unified the two branches (initial seed and HEAD resync both end in subscribe-then-reconcile), which reads better than the old asymmetry.

apps/os/src/components/repo-ide/repo-typescript.ts
mmkal: 🤖 Agreed, and it would have been a rough silent failure given the tsconfig honoring is the headline stretch. Fixed in 1f9b31343: `repoSeedPaths` (now in the pure `repo-typescript-vfs.ts` module) pins `tsconfig.json` past the cap, with a spec case asserting a 600-file repo keeps it. On also pinning the open buffer's path: I left that out deliberately — the session is per-repo and doesn't know which file is open, and the miss self-heals on the first edit (a working-tree entry puts the full content into the reconcile diff regardless of the cap). Beyond-cap repos keep their generic console warning.

apps/os/src/components/repo-ide/repo-editor-pane.tsx
mmkal: 🤖 Agreed, and went one step further than `placeholderData: keepPreviousData` (which landed as suggested): the extension memo now keys on the reference-stable PARTS — the worker facade and the imported module namespaces — instead of the query-data envelope. Since those keep identity across refetches for the life of the session, the extension array identity survives the post-commit refetch entirely: zero marginal EditorView rebuilds and no squiggly gap from the TS chain on commit (the pre-existing `textBaseline` rebuild still happens, as before this PR). All in 1f9b31343. The once-per-file-open rebuild when the query first resolves is inherent to `SourceCodeBlock`'s recreate-on-extensions-change approach — I've filed the `Compartment` rework in the task file's follow-ups, since you're right that every future extension-changing feature keeps paying it.

apps/os/src/components/repo-ide/repo-typescript.worker.ts
mmkal: 🤖 Agreed on the narrower seam — moved in 1f9b31343. The worker no longer monkey-patches `getSemanticDiagnostics` (its diagnostics are fully unfiltered, so a future Problems panel inherits nothing silently); the suppression now lives in `workerFacade.getLints`, which appends 2347 to the `diagnosticCodesToIgnore` the lint extension already threads through. Your point 1 (untyped LOCAL js helpers called with type args) is real and remains suppressed in the editor's lint lane — I judged the until-typm noise (every template repo squiggles on the sdk's `kv.get<number>`) to outweigh that fairly exotic miss, but it's now documented at the suppression site and in the task file so the typm follow-up drops it wholesale. Also took the review-body notes: `extends` non-support is documented on `repoCompilerOptions`, the deliberately-absent route-unmount teardown got its docstring sentence, the `#desired`/`#reconcile` math moved to a pure module (`repo-typescript-vfs.ts`) with the suggested vitest spec against a real `WorkingTreeStore`, and the REPL's raw-proxy-in-state hazard is filed as a follow-up rather than expanding this PR into `itx-repl.tsx`.

mmkal:
🤖 Review handled — all five inline findings accepted and fixed in 1f9b31343 (plus a task-file dedupe in f166de2f0); details on each thread. The review-body items:

- **Vitest spec for the sync seam**: done — the pure math (seed-path capping, HEAD⊕working-tree overlay, push diff) moved to `repo-typescript-vfs.ts` and `repo-typescript-vfs.test.ts` drives it through a real `WorkingTreeStore` (overlay, working-over-staged precedence, diff batches, tsconfig pin).
- **No route-unmount teardown**: deliberate — the warm language service survives in-app navigation away and back, and the one-slot registry bounds the cost at a single worker. Now said explicitly in the registry docstring.
- **tsconfig `extends` ignored**: now documented on `repoCompilerOptions` and in the task file — chasing npm-hosted config chains is typm territory.
- **`itx-repl.tsx` raw proxy in React state**: agreed it is the same landmine this PR documents, but porting `workerFacade` there means touching (and re-verifying) the REPL, so it is a task-file follow-up rather than scope creep here.

CI green on the new head; live re-verified (squiggly on demand in `main.ts`, template `worker.ts` still quiet through the host-side filter).

mmkal:
🤖 Closing: this landed on main via #1781, which was retargeted from this branch to `main` and squash-merged with **both** changesets — the TypeScript language server from this PR (diagnostics, hover, autocomplete via the @typescript/vfs web worker) and typm on top of it. All five review threads here were addressed and resolved before the merge (commits 1f9b31343, f166de2f0). Nothing from this branch was dropped.

PR #1773, opened 2026-07-08
Repo IDE: SVG preview + preview in the Index view

mmkal:
> [!NOTE]
> **Stacked on #1767** (`repo-ide-html-preview`) — review/merge that first; this diff is only the follow-up commits.

The two follow-ups the HTML preview PR deliberately left out of scope:

## Demo

SVG rendered through the sandboxed iframe (`specs/repo-ide-svg-index-preview.spec.ts`, `VIDEO_MODE=1`):

svg preview

The Code/Preview toggle riding into the readonly Index view — previewing the staged snapshot:

index preview

## 1. SVG preview

`.svg` files get the same Code/Preview toggle as `.html`. They render through the **same sandboxed `<iframe srcdoc>` lane** — raw svg markup is valid in an html body (the parser switches to foreign content at `<svg>`; an XML prolog degrades to an ignored comment), so this is a one-line predicate change (`isHtmlPreviewPath` now matches `/\.(html?|svg)$/i`), zero new rendering code, and one sandbox stance to reason about.

Why not an `` with a blob URL, which also neuters scripts? Because it neuters them *entirely*: script-driven animation and interactivity — legitimate things to preview — never run. The iframe's existing stance (`allow-scripts`, **no** `allow-same-origin` → opaque origin) keeps user-supplied `<script>` working AND inert against the dashboard. (An earlier version of this rationale also claimed `` breaks sizing and SMIL animations; review testing showed both behave identically in the two lanes, so the decision rests on the script story alone.)

**Code view** — note the embedded `alert("pwned")` + `window.top.document.title` hijack attempt:

**Preview** — shapes + SMIL animation render; the script ran inert (no dialog appeared, the dashboard title stayed "OS" — asserted from the driving Playwright script):

## 2. Preview in the readonly Index view

The staged pseudo-file view (`path (Index)`) gets the identical toggle for html/svg, rendering the **staged snapshot** readonly — same `CodePreviewToggle`, same `leading` slot, header suffix flips `(Index)` ↔ `(Index Preview)`.

**Index code view** (readonly HEAD↔staged diff, toggle now present):

**Index preview** — the staged-only heading, never committed:

## Verified live

Playwright walkthrough against a local dev server (`getin`-minted session, repo seeded via the itx CLI): svg toggle + inert-script assertions above; staged an edit to `hello.html` through the UI, opened it from Staged Changes, toggled Preview and back. `typecheck`, `lint`, `format` green. Task file: `tasks/complete/2026-07-08-repo-ide-preview-followups.md`.

## Environment Config Lease

<details>

| Auth | released | `b5040aa` | [https://auth.iterate-preview-1.com](https://auth.iterate-preview-1.com) | 811.8 KiB | 18.9s | 455ms |  | 2.9s | [Workflow run](https://github.com/iterate/iterate/actions/runs/477375666435079) | 2026-07-09T08:45:21.007Z | Preview app released. |
| OS | released | `b5040aa` | [https://os.iterate-preview-1.com](https://os.iterate-preview-1.com) | 3.14 MiB (±0 vs main) | 65.1s | 139.1s | 10 retried: itx > Authenticated internal auth itx can create project and append to stream (vitest x1) · itx > Project describe exposes self-describing builtin capabilities (vitest x1) · DESIRED: a live-capability fetch handler serves WebSockets at an app host (vitest x1) · creates a project and uses project streams through v4 ITX (vitest x1) · stream getEvents defaults to a bounded page and supports event type filters (vitest x1) · stream rules cross-post matching events with source provenance (vitest x1) · onboarding agent replies to a chat message in the feed (specs x1) · reactivity page repaints from a stream subscription after a page action (specs x1) · toggle an html file between Code and its sandboxed Preview (specs x1, still failed) · toggle an svg file between Code and its sandboxed Preview (specs x1) | 12.4s | [Workflow run](https://github.com/iterate/iterate/actions/runs/477375666435079) | 2026-07-09T08:45:30.546Z | Preview app released. |

</details>

---

> [!NOTE]
> **Low Risk**
> Reuses the existing opaque-origin iframe preview path for SVG and staged snapshots; no new auth or data APIs, with behavior covered by Playwright.
> 
> **Overview**
> Extends the repo IDE **Code | Preview** flow in two places: **`.svg`** files now preview through the same sandboxed `srcdoc` iframe as HTML (via a one-line `isHtmlPreviewPath` regex change—no new renderer), and the readonly **Index** staged view gets the same toggle for markdown/html/svg, showing **`staged.content`** with the header suffix **`(Index Preview)`** while staying non-editable.
> 
> Playwright coverage adds `repo-ide-svg-index-preview.spec.ts` and bumps Preview/Code tab click timeouts in the existing html/markdown preview specs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 32fed29b13f2e1091265f6ffe55c3a4924f80511. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

| --- | ---: | ---: |
| UI components | +49 -27 | +28 -14 🟩🟩🟥⬜⬜ |
| Tests | +84 -4 | +63 -4 🟩🟩🟩🟩⬜ |
| Docs | +96 -0 | +84 -0 🟩🟩🟩🟩🟩 |
| **Total** | **+229 -31** | **+175 -18** 🟩🟩🟩🟩🟩 |

<sub>Lines counts every changed line; Significant ignores blank lines and JS comments. Between 07e3161 and 32fed29, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>


apps/os/src/components/repo-ide/repo-file-kinds.ts
```diff
@@ -75,8 +75,12 @@ export function isBinaryRepoPath(path: string): boolean {
...
- * also opens as html-highlighted text but is an image format, not a page. */
+/** Paths whose Code/Preview toggle renders through the sandboxed html iframe:
+ * real html documents, plus .svg — raw svg markup is valid in an html body
+ * (the parser switches to foreign content at `<svg>`; an XML prolog degrades
+ * to an ignored bogus comment), so the same srcdoc lane previews it with the
+ * same script-inert sandbox. An `<img>` would also neuter scripts but breaks
+ * height-less/percentage sizing and never runs legitimate svg animations. */
 export function isHtmlPreviewPath(path: string): boolean {
-  return /\.html?$/i.test(path);
+  return /\.(html?|svg)$/i.test(path);
```
mmkal, line 87: 🤖 Reusing the srcdoc lane is the right call, but two of the three claims recorded here don't survive an empirical check (Chromium via the repo's playwright, viewport 800px, iframe 600px):

- **Sizing**: `` does *not* collapse height-less SVGs when a `viewBox` is present — it stretch-fits to the container width using the viewBox ratio, exactly like inline svg in an html body does. Measured: viewBox-only svg → `` 784×784 vs srcdoc-inline 584×584 (both = their available width); `width="100%" height="100%"` + viewBox → same stretch-fit in both; *no* viewBox → identical 300×150 default in both; `width="200"` only → 200×150 in both. I couldn't construct a case where the two lanes size differently.
- **Animations**: SMIL animates inside `` just fine (verified by pixel-diffing two element screenshots 230ms apart on an `<animate>` circle). What `` actually suppresses is scripts — script-driven animation and interactivity.

The claims that hold: raw svg parses as foreign content ✓, the XML prolog degrades to a bogus comment ✓ (and srcdoc documents never enter quirks mode even with a legacy SVG DOCTYPE — `compatMode` stayed `CSS1Compat`), scripts run but inert against the dashboard ✓, and it's zero new rendering code with one sandbox stance ✓. Those alone justify the choice — no code change wanted. But this comment is the load-bearing record of the decision, so I'd trim the sizing/SMIL parts to just "an `` would also neuter scripts, but then script-driven animation and interactivity never run" so a future reader doesn't inherit the false premises. Same correction applies to the PR body's "height-less/percentage-sized SVGs collapse to the 300×150 replaced-element default" and "scripted/SMIL animations never run", and to the task file's design-decisions bullet.

mmkal (commented):
🤖 Reviewed the delta over #1767 only. The Index-view half is solid: the preview branch renders `<HtmlPreview html={staged.content}>` with no onChange path, the `SourceCodeBlock` stays `editable={false}`, and the toggle only patches the `preview` URL param — I couldn't find any edit path this opens. `onOpenStaged` resets `preview: undefined`, so no stale preview state leaks into a freshly opened Index file, and the `(Index)` ↔ `(Index Preview)` suffix flip mirrors the working view's suffix precedence. Predicate change and toggle placement are consistent with the parent branch's patterns, and the PR body/task file describe the diff accurately — with one exception: I empirically tested the srcdoc-over-`` justification and two of its claims are wrong in Chromium (details in the inline comment on `repo-file-kinds.ts`). The *decision* still stands on its remaining legs, so it's a comment/PR-body fix, not a code fix.

One verified-good claim worth calling out: the XML prolog + legacy SVG DOCTYPE case renders fine in the srcdoc lane, and `document.compatMode` stays `CSS1Compat` — iframe srcdoc documents never enter quirks mode regardless of the doctype, which is a stronger guarantee than the comment even claims.

One non-blocking question: `onOpenWorking` (repo-ide.tsx:298) resets `staged` and `diff` but not `preview`, so Index Preview → "Open file" lands directly in the working view's Preview over the working buffer. That reads as coherent — arguably the nicest behavior — but since `diff` is explicitly reset on the same transition, checking it's deliberate rather than an accident of `preview` not existing in the staged view until now.

apps/os/src/components/repo-ide/repo-file-kinds.ts
mmkal: 🤖 You're right, and thanks for actually measuring it — accepted. Trimmed both false premises in f2b4dfd79: the doc comment on `isHtmlPreviewPath` now claims only what held up (scripts are what `` suppresses; the opaque-origin sandbox keeps script-driven animation/interactivity working AND inert, zero new rendering code, one sandbox stance). Task file and PR body corrected to match, with a note that the earlier sizing/SMIL claims were review-tested and removed, so the record shows the decision standing on its true legs.

mmkal:
🤖 On the non-blocking question — `onOpenWorking` keeping `preview` across Index → "Open file": honest answer, it's emergent (the handler predates this PR and I didn't add `preview` to its reset list), but having thought it through I endorse it as the right behavior and am leaving it. Preview means the same thing in both views — "render the current buffer" — so carrying the mode across the transition is coherent: you were previewing the staged snapshot, you keep previewing, now over the working buffer you came to edit-or-look-at. `diff` is different in kind, which is why resetting it there is right: the Index diff is a readonly HEAD↔staged inspection while the working diff is an interactive staging surface, so carrying *that* flag over would silently change what the view does. No code change.

PR #1781, opened 2026-07-08
typm: package.json-driven type acquisition for the repo IDE

mmkal:
> **Includes #1771 in full.** This PR was stacked on #1771 (`repo-ide-typescript-lsp`) and has been retargeted to `main` to land both together: the TypeScript language server (in-browser diagnostics, hover, autocomplete via a `@typescript/vfs` web worker — see #1771 for details) plus typm, described below. #1771 is closed in favor of this one.

## What

`typm` — like pnpm, but just for types. The repo IDE's TypeScript language service (from #1771) currently types every bare npm import as `any` via a `declare module "*"` wildcard. This PR reads the repo's **package.json** (lockfiles deliberately not respected), recursively pulls its npm dependencies over jsdelivr, throws away everything except `.d.ts` files + `package.json`s, and drops the result into the language-service worker's `/node_modules` — so imports get their real types.

## Before / after

A repo with `"dependencies": { "zod": "^3.24.0" }` and:

```ts
import { z } from "zod";

const user = z.object({ name: z.string() });
user.parse({ name: 42 });
```

- **Before**: `z` is `any`. No completions, no hover docs, no errors — `z.objceт(…)` typos sail through.
- **After**: `z.` autocompletes the real zod surface with docs, hover shows real signatures, and type errors squiggle — while `user.parse({ name: 42 })` still runs (zod checks at runtime), the compiler now knows `user` is `ZodObject<{name: ZodString}>` and `user.parse("nope").name.toUpperCase()` type-checks accordingly.

Acquisition failures (offline, unpublished package, budget exceeded) degrade to exactly today's wildcard-`any` behavior — the editor never blocks or breaks on typm.

## Live verification (local dev)

A seeded repo with `zod`, `react`, and `@tanstack/react-query` in package.json. Console: `[typm] acquired 6 packages (392 files, 2597 kB): @tanstack/react-query@5.101.2, zod@3.25.76, @types/react@18.3.31, @tanstack/query-core@5.101.2, @types/prop-types@15.7.15, csstype@3.2.3` — note `@types/react@18` (major-matched to the repo's react 18, where ata would fetch 19) and the transitive/peer pulls.

Hover shows the real acquired zod types:

…and a genuine cross-module type error squiggles (string→number through `user.parse(…).name`):

The react-query JSX file (`useQuery` + JSX under `strict`) is squiggle-free with the acquired types.

Live verification also caught a real bug: `forceLinting` only flushes a *pending* lint query, and an untouched buffer has none — so acquired types never repainted stale diagnostics until a keystroke. Fixed with a `linter(null, {needsRefresh})` extension watching a `typesAcquired` StateEffect (dispatch marks the lint state stale, then `forceLinting` flushes immediately).

## How (decisions detailed in `tasks/complete/2026-07-08-repo-ide-typm.md`)

- **Rebuilt the acquisition loop rather than wrapping `@typescript/ata`**: ata is source-import-driven and resolves transitive references at `latest` (its own TODO admits it), which breaks version fidelity — e.g. react 18 repos would get `@types/react@19`. We keep its proven jsdelivr endpoint strategy (resolve API for semver ranges, flat listings, CDN file fetches) with attribution.
- **`packages/typm`**: the acquisition core is a pure fetch-injected module (no node APIs, no react) in a source-consumed workspace package; the repo IDE worker is its first consumer.
- **Cache API** persistence for immutable `package@version` URLs; in-memory dedupe per session.
- Trigger: after worker seeding + debounced when package.json changes in the working tree; completed acquisition force-relints open buffers.

| --- | ---: | ---: |
| Product | +368 -0 | +257 -0 🟩🟩⬜⬜⬜ |
| UI components | +237 -10 | +142 -6 🟩⬜⬜⬜⬜ |
| Tests | +615 -4 | +524 -4 🟩🟩🟩🟩🟩 |
| Config | +37 -0 | +37 -0 🟩⬜⬜⬜⬜ |
| Docs | +75 -0 | +58 -0 🟩⬜⬜⬜⬜ |
| Generated | +15 -0 | +14 -0 🟩⬜⬜⬜⬜ |
| **Total** | **+1,347 -14** | **+1,032 -10** 🟩🟩🟩🟩🟩 |

<sub>Lines counts every changed line; Significant ignores blank lines and JS comments. Between 7b17098 and c8bc02e, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>

## Environment Config Lease

<details>

| Auth | released | `c8bc02e` | [https://auth.iterate-preview-1.com](https://auth.iterate-preview-1.com) | 812.0 KiB | 16.6s | 858ms |  | 3.3s | [Workflow run](https://github.com/iterate/iterate/actions/runs/585860997902067) | 2026-07-09T13:49:05.611Z | Preview app released. |
| OS | released | `c8bc02e` | [https://os.iterate-preview-1.com](https://os.iterate-preview-1.com) | 4.07 MiB (+929.6 KiB vs main) | 78.3s | 187.3s | 7 retried: catalogue example "sandbox-exec" runs identically across runtimes (vitest x1) · itx > Agent scripts can send web-chat messages (with file attachments) and call project tools (vitest x1) · itx > Project worker processEventBatch receives events from every project stream and can cross-post (vitest x1) · DESIRED: a live-capability fetch handler serves WebSockets at an app host (vitest x1) · stream getEvents defaults to a bounded page and supports event type filters (vitest x1) · runs "scheduler-basics" through the project REPL (specs x1) · flags a package.json schema violation with a red squiggle (specs x1) | 17.1s | [Workflow run](https://github.com/iterate/iterate/actions/runs/585860997902067) | 2026-07-09T13:49:19.416Z | Preview app released. |
| Semaphore | released | `e3e0249` | [https://semaphore.iterate-preview-1.com](https://semaphore.iterate-preview-1.com) | 440.7 KiB | 13.2s | 6.0s |  | 800ms | [Workflow run](https://github.com/iterate/iterate/actions/runs/257511147905723) | 2026-07-09T13:49:03.117Z | Preview app released. |
| Streams Example App | released | `e3e0249` | [https://streams.iterate-preview-1.com](https://streams.iterate-preview-1.com) | 1.29 MiB | 14.9s | 19.3s |  | 765ms | [Workflow run](https://github.com/iterate/iterate/actions/runs/257511147905723) | 2026-07-09T13:49:03.088Z | Preview app released. |

</details>

---

> [!NOTE]
> **Medium Risk**
> Adds worker-side CDN fetches and large vfs writes, but acquisition is capped, failures fall back to existing wildcard-`any` behavior, and the editor integration is the main regression surface.
> 
> **Overview**
> Introduces **`@iterate-com/typm`**, which reads a repo’s **package.json** (not lockfiles), resolves dependencies over jsdelivr, keeps only declaration files plus **package.json** trees, and returns a **`/node_modules/...`** map for the in-browser language service.
> 
> The repo IDE **TypeScript worker** gains **`acquireTypes`**: cached CDN fetch, dependency-snapshot dedupe, serialized runs, and eviction of stale acquired paths when the manifest changes. The host session kicks acquisition after seed and on **debounced** effective **package.json** edits (including clearing types when the manifest is removed), with **retry** on transient CDN failures.
> 
> **`package.json`** is **pinned** past the 500-file seed cap so typm still runs on large repos. After types land, a **CodeMirror** **`needsRefresh` + `StateEffect`** path forces diagnostics to repaint (fixing stale squiggles without a keystroke). Vite pre-bundles **`@codemirror/lint`** and **`@codemirror/state`** to keep facet identity stable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c8bc02e27f282e98182352e87e74022cc2a93b71. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


apps/os/src/components/repo-ide/repo-typescript.worker.ts
```diff
@@ -198,6 +239,36 @@ const api = {
...
+   * trees into the vfs. Returns whether anything new landed (the host
+   * force-relints open buffers when it did). No-ops when the dependency maps
+   * are unchanged or the JSON doesn't parse (mid-edit).
+   */
+  async acquireTypes(input: { packageJsonText: string }): Promise<{ acquired: boolean }> {
+    const snapshot = dependencySnapshot(input.packageJsonText);
+    if (!snapshot || snapshot === lastAcquiredDependencies || !worker.getEnv()) {
+      return { acquired: false };
+    }
+    lastAcquiredDependencies = snapshot;
```
mmkal: 🤖 **Race: overlapping `acquireTypes` runs can leave stale types that never self-heal.** Nothing serializes calls here. Plausible sequence: the post-seed kick (delay 0) starts run A (a multi-MB download); the user edits a dep range (`zod ^3` → `^4`); 1s later the host fires run B. Both runs write into the vfs; if A finishes last, its `zod@3` files overwrite B's `zod@4` for every shared package. Worse, `lastAcquiredDependencies` already holds B's snapshot, so the stale result sticks until the dep maps change *again*.

Cheapest fixes: chain runs on a module-level promise (like the host's `#chain`), or take a run-id before the await and skip the vfs writes if a newer run started.

packages/typm/src/typm.ts
```diff
@@ -0,0 +1,250 @@
...
+
+  const fetchJson = async (url: string): Promise<unknown> => {
+    const response = await input.fetch(url);
+    if (!response.ok) return null;
+    return response.json();
+  };
+
+  const resolveVersion = async (dependency: PendingDependency): Promise<string | null> => {
+    for (const range of dependency.ranges) {
+      const url = `https://data.jsdelivr.com/v1/package/resolve/npm/${dependency.name}@${encodeURIComponent(range)}`;
```
mmkal, line 109: 🤖 **Package names are user-controlled but go unvalidated/unencoded into URLs and vfs paths.** `dependency.name` comes straight from repo package.json keys (and from fetched transitive package.jsons). A name containing `?`, `#`, whitespace, or `../` rewrites the jsdelivr request path/query here and at the `/flat` + CDN fetches — and later flows raw into vfs keys as `` `/node_modules/${name}${file.name}` ``. Blast radius is contained (fixed host, worker-side, types-only), but a hostile or garbage package.json shouldn't get to shape request URLs or write `..`-containing vfs paths.

Cheap hardening: validate names against npm's name grammar (`/^(@[a-z0-9-~][a-z0-9-._~]*\/)?[a-z0-9-~][a-z0-9-._~]*$/`) in the same place `unfetchableRangeReason` runs, and skip-with-warning on mismatch.

packages/typm/src/typm.ts
```diff
@@ -0,0 +1,250 @@
...
+      }
+      seen.add(dependency.name);
+      const unfetchable = unfetchableRangeReason(dependency.ranges[0]!);
+      if (unfetchable) {
+        warn(`skipping ${dependency.name}: ${unfetchable}`);
+        continue;
+      }
+      batch.push(dependency);
+    }
+    const nextDependencies = await Promise.all(batch.map((dependency) => acquireOne(dependency)));
```
mmkal: 🤖 **One rejected fetch aborts the entire acquisition, discarding every completed package.** Non-ok *responses* warn-and-continue, but a *thrown* fetch (network blip mid-run — likely during a multi-hundred-file pull; also the `new Response(body, {status})` replay in cached-fetch throws `TypeError` for null-body statuses like 204/304) propagates through `fetchJson`/`acquireOne` and rejects this `Promise.all`, so `acquireTypes` rejects wholesale. The worker catches it, so the editor is fine — but the PR body's "degrade to wildcard-any" becomes all-or-nothing rather than per-package, and combined with the host-side retrigger gap (see worker comment) one blip means no types for the session.

A try/catch inside `acquireOne` (warn + return `[]`) would make the per-package degradation story true.

apps/os/src/components/repo-ide/repo-typescript.worker.ts
```diff
@@ -198,6 +239,36 @@ const api = {
...
+        log: (message) => console.info(`[typm] ${message}`),
+        limits: TYPM_LIMITS,
+      });
+      for (const [path, content] of Object.entries(result.files)) {
+        worker.updateFile({ path, code: nonEmpty(content) });
+      }
+      return { acquired: Object.keys(result.files).length > 0 };
+    } catch (error) {
+      // Transient failure (offline, CDN hiccup): allow a later retrigger.
+      lastAcquiredDependencies = null;
```
mmkal: 🤖 **This retry path is unreachable through the only caller.** Resetting `lastAcquiredDependencies` lets a *later retrigger* succeed — but the host's `#maybeAcquireTypes` guards on `#lastRequestedPackageJson`, which still holds the current text, so no retrigger ever arrives until package.json's text actually changes. Net effect: transient failure at seed time (offline, CDN hiccup) leaves the session in wildcard-`any` mode for its whole life. If that's acceptable, worth saying so honestly in the comment; otherwise the host needs to reset `#lastRequestedPackageJson` when the call resolves `{acquired: false}` after a failure (or the worker result could distinguish "failed" from "no-op").

apps/os/src/components/repo-ide/repo-typescript.ts
```diff
@@ -118,10 +125,48 @@ class RepoTypeScriptSession {
...
+
+  /**
+   * Kick typm (npm type acquisition) in the worker whenever the effective
+   * package.json content changes. Debounced — package.json edits arrive per
+   * keystroke — and cheap to over-call: the worker no-ops unless the parsed
+   * dependency maps actually changed. Failures degrade to the prelude's
+   * `declare module "*"` wildcard (bare imports stay `any`), never the editor.
+   */
+  #maybeAcquireTypes(desired: Map<string, string>, delayMs: number): void {
+    const packageJsonText = desired.get("package.json");
```
mmkal: 🤖 **`package.json` can silently fall past the 500-file seed cap.** `repoSeedPaths` (base branch) caps at `MAX_SEED_FILES` and pins `tsconfig.json` past the cap precisely because dropping it would silently change behavior — but `package.json` gets no such pin, and it sorts after `apps/…`, `lib/…`, etc. In a repo with >500 TypeScript-relevant files, `desired.get("package.json")` returns `undefined` here and typm never runs, with no warning. Same one-line fix as tsconfig: pin `package.json` past the cap.

packages/typm/src/typm.ts
```diff
@@ -0,0 +1,250 @@
...
+function unfetchableRangeReason(range: string): string | null {
+  const match = /^(workspace:|file:|link:|portal:|git|github:|https?:)/.exec(range);
+  return match ? `unsupported version range "${range}"` : null;
+}
+
+function createLimiter(max: number) {
+  let active = 0;
+  const waiters: Array<() => void> = [];
+  return async function limited<T>(fn: () => Promise<T>): Promise<T> {
+    if (active >= max) await new Promise<void>((resolve) => waiters.push(resolve));
```
mmkal: 🤖 Minor: the limiter can exceed `max`. A finishing task decrements `active` and resolves one waiter, but that waiter resumes in a later microtask — a fresh synchronous caller can slip in first (`active < max` → proceeds), and then the woken waiter *also* increments, so `active` lands above `max`. Bounded overshoot and harmless against a CDN, so fine to leave — flagging in case typm grows other consumers. The standard fix is to loop the wait (`while (active >= max) await …`) or increment before resolving the waiter.

mmkal (commented):
🤖 Reviewed the stacked diff (typm core, cached fetch, worker/session integration, lint-refresh plumbing). Solid work overall: the fake-registry tests are readable and cover the interesting semver/fallback/budget paths, the budget check-and-add is atomic (sync between awaits), the `linter(null, {needsRefresh})` claim checks out against `@codemirror/lint`'s types, hover rendering uses `innerText` so acquired d.ts content can't inject HTML, and the 17-test claim is accurate. Findings inline — one race worth fixing (overlapping `acquireTypes` runs), plus a handful of lower-severity robustness gaps (unvalidated package names in URLs/vfs paths, whole-run abort on a single rejected fetch, an unreachable retry path, and package.json falling past the 500-file seed cap).

apps/os/src/components/repo-ide/repo-typescript.ts
```diff
@@ -118,10 +125,48 @@ class RepoTypeScriptSession {
...
+    if (packageJsonText === undefined) return;
+    if (packageJsonText === this.#lastRequestedPackageJson) return;
```
mmkal, line 157: 🤖 **Medium: the worker's "allow a later retrigger" on transient failure is unreachable from here.** On a caught acquisition failure (offline, CDN hiccup) the worker resets `lastAcquiredDependencies = null` specifically so a retry can happen — but the host gates every call on this text-equality check, and `#lastRequestedPackageJson` is never reset on failure (the worker swallows the error and returns `{acquired: false}`, so the `.catch` below never fires either). Net effect: open the IDE while offline and types never arrive for the rest of the session, even after coming back online, unless the user happens to edit package.json. The PR body's "degrades to wildcard-any" story makes this defensible, but the two layers' comments contradict each other — either surface failure in the return value (`{acquired: false, failed: true}`) and clear `#lastRequestedPackageJson`, or drop the worker-side "allow a later retrigger" reset/comment so the no-retry behavior is stated honestly.

mmkal (commented):
🤖 Reviewed the stacked diff (typm core, cached fetch, worker/host integration, repaint mechanism). The design holds up well under scrutiny — things I verified rather than took on faith:

- `linter(null, {needsRefresh})` is legit on the installed `@codemirror/lint@6.9.5` (`linter(source: LintSource | null, …)`), and `needsRefresh` combines across configs with OR — so the extra config cleanly marks the valtown linter's results stale. `@codemirror/state` is already in the vite `optimizeDeps.include` set, so the `StateEffect.define` instance can't split across copies.
- The byte-budget check-then-add in `acquireOne` is microtask-atomic (no await between check and increment), so concurrent packages in a wave can't race past the budget.
- Egress is pinned to `data.jsdelivr.com` + `cdn.jsdelivr.net`; fan-out is fused by `maxPackages` (checked before `seen.add`) and the byte budget — with one gap noted inline.
- Terminate/debounce lifecycle is sound: the timer is cleared in `terminate()`, the callback re-checks `#terminated`, and acquisition only ever fires after `initializeRepo` resolved, so the worker env exists.
- "17 unit tests" checks out (14 + 3); tests follow repo conventions (no vi.mock, helpers at the bottom, `toMatchObject`).

One non-inline observation: **the pnpm-lock.yaml churn goes beyond adding `packages/typm`**. Neither main nor the base branch's lockfile has the top-level `catalogs:` section, and both have 18 occurrences of `typescript@6.0.3` vs 104 on this branch — pnpm 10.24 rewrote the lockfile format and re-resolved ~86 peer keys from `typescript@5.9.3` to `6.0.3` for unrelated packages (vue, trpc-cli, msw, tsdown, opentui…). CI is green so it's probably benign catch-up that any dep-touching branch would trigger, but it changes auto-installed peers for unrelated packages and will conflict with sibling branches — flagging so it's a known trade rather than a surprise.

Inline: one medium finding (the transient-failure retrigger is unreachable in practice), and three smaller ones.

apps/os/src/components/repo-ide/repo-typescript.worker.ts
```diff
@@ -198,6 +239,36 @@ const api = {
...
+    lastAcquiredDependencies = snapshot;
+    try {
+      const result = await typmAcquireTypes({
+        packageJson: input.packageJsonText,
+        fetch: typmFetch,
+        log: (message) => console.info(`[typm] ${message}`),
+        limits: TYPM_LIMITS,
+      });
+      for (const [path, content] of Object.entries(result.files)) {
+        worker.updateFile({ path, code: nonEmpty(content) });
```
mmkal: 🤖 **Stale acquired types are never removed from the vfs.** Two consequences:

1. Remove `zod` from package.json → the new acquisition's `result.files` no longer contains zod, but the previously-written `/node_modules/zod/**` stays — imports of a removed dep keep their (now-phantom) real types instead of degrading to the wildcard, and `acquired` comes back `false` so no relint happens even though effective types should have changed.
2. Bump `zod` `^3` → `^4` → new files overwrite same-named paths, but files that existed in v3 and not v4 linger, leaving a mixed-version tree that resolution can trip over (stale relative imports inside old `.d.ts` files).

Cheap fix within the existing seam: remember the previous acquisition's file list and `env.deleteFile` the paths absent from the new result before writing. Related nit while you're here: two `acquireTypes` calls can overlap (snapshot is taken before the await, so two *different* snapshots run concurrently and interleave their writes) — serializing runs on a promise chain would close both.

packages/typm/src/typm.ts
```diff
@@ -0,0 +1,250 @@
...
+    if (declarationFiles.length === 0) {
+      // No shipped types: fall back to DefinitelyTyped, preferring the @types
+      // major that matches the resolved package version over blindly-latest.
+      const typesPackage = `@types/${definitelyTypedName(name)}`;
+      const major = version.split(".")[0]!;
+      return [{ name: typesPackage, ranges: [major, "latest"] }];
+    }
+    const packageJsonEntry = tree.files.find((file) => file.name === "/package.json");
+    const wanted = packageJsonEntry ? [...declarationFiles, packageJsonEntry] : declarationFiles;
+    const bytes = wanted.reduce((sum, file) => sum + (file.size || 0), 0);
```
mmkal: 🤖 **Low (adversarial): the byte budget doesn't bound the number of file fetches.** `file.size || 0` means a package whose listing reports thousands of tiny/zero-size `.d.ts` files sails through the budget precheck while triggering an unbounded number of CDN requests (only concurrency is capped, at 8). An adversarial transitive package.json is in the threat model here — the fuse comments say as much. Cheap hardening: cap `wanted.length` per package (skip-whole like the byte case), or charge a fixed per-file overhead into `bytes`.

packages/typm/src/typm.ts
```diff
@@ -0,0 +1,250 @@
...
+    }
+    const declarationFiles = tree.files.filter((file) => isDeclarationPath(file.name));
+    if (declarationFiles.length === 0) {
+      // No shipped types: fall back to DefinitelyTyped, preferring the @types
+      // major that matches the resolved package version over blindly-latest.
+      const typesPackage = `@types/${definitelyTypedName(name)}`;
+      const major = version.split(".")[0]!;
+      return [{ name: typesPackage, ranges: [major, "latest"] }];
+    }
+    const packageJsonEntry = tree.files.find((file) => file.name === "/package.json");
```
mmkal: 🤖 **Low: only the root `/package.json` is kept — nested package.json files are dropped.** Packages that route subpath types through per-directory package.jsons (the pre-`exports` pattern: `pkg/sub/package.json` with a `types` field pointing outside `sub/`, e.g. older rxjs/date-fns layouts) will fail subpath resolution even though their `.d.ts` files were acquired. Bundler-mode resolution finds `sub/index.d.ts` without help, so this only bites the redirecting layouts — but keeping every `*/package.json` from the listing is nearly free (they're tiny and already size-counted) and removes the edge entirely.

apps/os/src/components/repo-ide/repo-typescript.worker.ts
mmkal: 🤖 Agreed — real race, and the never-self-heals part made it worth serializing rather than run-id-guarding (a dropped newer run would ALSO have needed snapshot repair). Fixed in babc2ea84: runs chain on a module-level promise (the worker-side sibling of the host's `#chain`), and the dep-snapshot dedupe moved inside the chain so a queued duplicate no-ops once its predecessor acquired the same deps. Writes now land strictly in request order.

packages/typm/src/typm.ts
mmkal: 🤖 Agreed, cheap hardening. Fixed in babc2ea84: names are validated right next to `unfetchableRangeReason` and skip-with-warning on mismatch. One deviation from the suggested regex: the unscoped part keeps uppercase (`[a-zA-Z0-9~-]`) so legacy packages like `JSONStream` still resolve — scopes stay lowercase-only. `..`, `/`, `?`, `#`, and whitespace are all excluded; the new unit test asserts nothing invalid ever becomes a request URL or a vfs path.

packages/typm/src/typm.ts
mmkal: 🤖 Agreed — the per-package degradation claim was false for *thrown* fetches. Fixed in babc2ea84: each `acquireOne` in the wave is caught individually (warn + continue), so a network blip costs only its own package. The cached-fetch replay hazard you spotted is fixed too: null-body statuses (101/204/205/304) replay bodyless instead of throwing from the `Response` constructor. Unit tests for both.

apps/os/src/components/repo-ide/repo-typescript.worker.ts
mmkal: 🤖 Agreed — the retry path was dead code through the only caller, and offline-at-seed sticking for the session's life wasn't the intent. Fixed in babc2ea84 with your suggested shape: the worker result now distinguishes `failed` (threw, or nothing acquired with warnings — total wipeout is transient-shaped now that failures are per-package) and the host forgets `#lastRequestedPackageJson` on failure (guarded against a newer request made meanwhile), so any later reconcile retriggers. The worker's snapshot dedupe keeps those retries free once one succeeds.

apps/os/src/components/repo-ide/repo-typescript.ts
mmkal: 🤖 Agreed — same silent-drop hazard the tsconfig pin exists for. Fixed in babc2ea84: `PINNED_SEED_PATHS = ["package.json", "tsconfig.json"]` are both kept past the cap, and the seed-cap spec now uses `app/` filler so both files genuinely sort past the cap and the pin is actually exercised.

packages/typm/src/typm.ts
mmkal: 🤖 Correct, and the fix is a one-liner so I took it rather than leaving the trap for future consumers: the wait is now a `while (active >= max)` loop, so a woken waiter re-checks the gate and a barging synchronous caller can no longer push `active` past `max`. In babc2ea84.

apps/os/src/components/repo-ide/repo-typescript.ts
mmkal: 🤖 Agreed (this landed twice across review rounds — same finding as the worker-side thread). Fixed in babc2ea84 with the suggested shape: the worker result is now `{acquired, failed}` (`failed` = threw, or nothing acquired with warnings) and on failure the host forgets `#lastRequestedPackageJson` — guarded against a newer request made while the run was in flight — so any later reconcile retriggers. The worker's snapshot dedupe keeps retries free once one succeeds, and the two layers' comments now tell the same story.

apps/os/src/components/repo-ide/repo-typescript.worker.ts
mmkal: 🤖 Both parts agreed and fixed. Stale removal (7cc0d2110): the worker keeps the set of vfs paths the last acquisition wrote and deletes the ones a new successful acquisition doesn't produce — removed deps degrade back to the wildcard instead of keeping phantom types, version bumps can't leave a mixed tree, and deletions count as a change so the host still relints (`acquired` is true when anything was written OR removed). Failed runs deliberately keep the previous types (a blip shouldn't degrade working buffers). Overlap (babc2ea84): runs are serialized on a worker-module promise chain with the snapshot check inside the chain, so interleaved writes can't happen at all.

packages/typm/src/typm.ts
mmkal: 🤖 Agreed — adversarial transitive package.jsons are in the threat model and the budget only counted content bytes. Fixed in 7cc0d2110 with the per-file-overhead flavor: every wanted file charges a flat 512 bytes against the budget, so the 25 MB budget also bounds fan-out (≤ ~51k requests worst case, realistically far fewer since real files have size). A test feeds a hostile listing of 5000 zero-size .d.ts files and asserts the package is skipped whole with zero CDN fetches.

packages/typm/src/typm.ts
mmkal: 🤖 Agreed — nearly free and removes the edge entirely. Fixed in 7cc0d2110: every `*/package.json` from the flat listing is kept (they're tiny and size-counted, incl. the new per-file overhead), with the root one still driving dependency recursion. Test covers the pre-`exports` layout (`pkg/sub/package.json` with a `types` field pointing outside `sub/`).

packages/typm/src/typm.ts
mmkal: 🤖 Agreed — the size-only precheck was blind to request count. Fixed in 7cc0d2110: the budget now charges a flat 512-byte per-file overhead (`PER_FILE_OVERHEAD_BYTES`), so a listing reporting thousands of zero-size files busts the budget in the precheck and the package is skipped whole before a single CDN fetch. Test: "the byte budget bounds request fan-out, not just content size" (a 5000-file zero-size listing produces zero `cdn.jsdelivr.net` requests).

packages/typm/src/typm.ts
mmkal: 🤖 Agreed, and it was nearly free as predicted. Fixed in 7cc0d2110: `acquireOne` now keeps every `*/package.json` from the flat listing (they're tiny and already size-counted), while dependency recursion still reads only the root one. Test: "keeps nested package.json files (pre-exports subpath types routing)" covers the `pkg/sub/package.json` → `types` redirect layout.

mmkal:
🤖 Re the review summary's lockfile-churn flag: confirmed it was an install artifact, not a requirement. Regenerated in 470b2c4cf from the base branch's lockfile with `pnpm install --lockfile-only` (same pnpm@10.24.0) — the ~86 unrelated `typescript@5.9.3 → 6.0.3` peer-suffix flips are gone and the diff is now just the `packages/typm` importer plus the `catalogs:` section. That section is unavoidable: pnpm 10.24 deterministically writes it for the workspace's `cloudflare` catalog on any lockfile touch (main's lockfile predates that normalization, so any branch touching the lockfile will grow it). Validated with `pnpm install --frozen-lockfile`.

mmkal:
I have some concerns about perf but this is pretty isolated/doesn't meaningfully make the interesting bits of iterate more complicated:

image

So I think I'll merge this so I don't have to keep track of it in my head

mmkal:
🤖 Heads up before you click merge: **#1781's base is `repo-ide-typescript-lsp` (the #1771 branch), not `main`** — it's stacked on the TS-language-server PR you're still reviewing. So squash-merging it now lands typm into the LSP branch, not into main.

Two ways to get typm into main:

1. **Merge #1771 → main first**, then GitHub auto-retargets #1781's base to `main` and you squash-merge it cleanly. (Cleanest, but needs #1771 in first — the one you wanted more time on.)
2. **Retarget #1781's base to `main` now** — but `repo-ide-typm` is stacked *on top of* the LSP branch, so its diff-to-main includes the entire #1771 LSP change too. Merging it would pull the language server in as well, defeating the point of holding #1771.

So option 1 is the real path: #1771 has to go in before typm can land in main independently. Want me to do anything, or are you driving both merges yourself?

mmkal:
🤖 Agreed on the isolation — it stays inside `packages/typm` + the repo-IDE worker seam and never touches the itx/domain core. For the perf concern, the safeguards already in the code (from the review round) so you have them in one place:

- **Bounded fan-out**: the byte budget charges a **512 B/file floor**, so a hostile or huge dep tree stops fetching once the budget is hit rather than fanning out unboundedly (there's a test: 5000 zero-byte files → skipped whole, zero CDN calls).
- **No redundant work**: first-wins dedupe across the wave, `@types` major-matched (won't pull `@types/react@19` for React 18), cached-fetch wrapper, and `acquireTypes` is **serialized** on a promise chain so a debounced re-run can't overlap the seed kick.
- **Runs are gated**: acquisition only fires on seed + debounced (1s) package.json edits, and no-ops when the dep snapshot is unchanged.

The residual cost is the initial jsdelivr fetch on repo open (types aren't persisted across sessions — re-acquired each open, though within a session cached-fetch covers repeats). If that's the bit that worries you, the natural follow-up is persisting acquired `.d.ts` in the DO/KV so reopens are free. Happy to spin that up as a separate task whenever — no need to hold this PR for it.

mmkal:
🤖 Agree it's isolated — the whole thing lives in `packages/typm` + the repo-ide worker seam, nothing in the hot path of the interesting bits.

On perf, the cost is bounded by design (all landed in the review round):
- **Off the editor's critical path**: acquisition is async in the TS worker; the editor renders and stays interactive whether or not types have arrived — they just sharpen diagnostics/hover once they land.
- **Bounded fan-out**: the byte budget charges a 512-byte floor per file, so even a hostile `package.json` (thousands of zero-byte deps) can't trigger unbounded jsdelivr fetches — it skips the whole acquisition instead.
- **Serialized + cached**: overlapping `acquireTypes` runs are chained (no thundering herd), first-wins dedup across the wave, and the cached-fetch wrapper memoizes per session so re-opening a repo is free.

If the concern is the *eager* acquisition on repo open, the cheapest win would be to defer it until the first `.ts`/`.tsx` file is actually opened (right now it kicks off from the seed). Happy to file that as a follow-up if you want it — not doing it unprompted since you're merging.

mmkal:
🤖 One heads-up before you merge: this PR is **stacked on #1771** (its base is `repo-ide-typescript-lsp`, not `main`) — typm builds on the TS-worker `setFiles`/`deleteFiles` seam that #1771 adds. So:

- If you merge **#1771 first**, GitHub auto-retargets this one to `main` and it merges cleanly after.
- If you merge **this one as-is**, it lands into the `repo-ide-typescript-lsp` branch, not `main`.

Both are updated off latest main and green. Want me to retarget #1781 to `main` (once #1771 is in) or otherwise sequence them for you?

packages/typm/src/typm.ts
```diff
@@ -0,0 +1,291 @@
...
+    // FAN-OUT: a hostile listing reporting thousands of zero-size files would
+    // otherwise pass the precheck while triggering a fetch per file.
+    const bytes = wanted.reduce((sum, file) => sum + (file.size || 0) + PER_FILE_OVERHEAD_BYTES, 0);
+    if (totalBytes + bytes > input.limits.maxTotalBytes) {
+      warn(
+        `skipping ${name}@${version}: ${bytes} bytes (incl. per-file overhead) would exceed the ${input.limits.maxTotalBytes} byte budget`,
+      );
+      return [];
+    }
+    totalBytes += bytes;
```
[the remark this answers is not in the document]
mmkal: 🤖 False positive — no race here. The budget check and the charge are consecutive **synchronous** statements:

```js
if (totalBytes + bytes > input.limits.maxTotalBytes) { warn(...); return []; }  // L142
totalBytes += bytes;                                                            // L148
```

There is no `await` between them, so in single-threaded JS that check→charge is atomic. The concurrency from `Promise.all` only yields at the `await` points (the tree listing fetch, then the downloads at L149) — never between L142 and L148. A second `acquireOne` cannot read a stale `totalBytes`, pass its precheck, and slip its charge in before this one commits, because it can only resume at an await boundary. So the 25 MB budget holds. Resolving.

apps/os/src/components/repo-ide/repo-typescript.ts
```diff
@@ -118,10 +125,64 @@ class RepoTypeScriptSession {
...
```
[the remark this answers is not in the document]
mmkal: 🤖 This one is real, though narrow. If `package.json` is deleted (or renamed away) *within the 1s debounce window* of a just-armed acquisition, `#maybeAcquireTypes` early-returns at the `packageJsonText === undefined` guard (L151) without clearing `#acquireTimer` — so the pending timer still fires and acquires types for the now-deleted manifest, and the previously-written `/node_modules/**` types are never evicted (the worker only prunes stale files during a *new successful* acquisition, which no longer runs). It self-heals on the next package.json edit, and deleting package.json in the IDE is rare, so I read it as low-harm rather than Medium — but it is in keeping with the other edge-cases this PR pins (tsconfig/package.json seed pin, stale-type removal), so worth a one-liner.

Cheap fix in the same guard: when `packageJsonText === undefined`, clear the pending timer and reset `#lastRequestedPackageJson = null` (and optionally ask the worker to drop `/node_modules` types). @mmkal — since you're driving this merge yourself I have **not** pushed anything; say the word and I'll land just this fix in one commit, otherwise it's fine to merge as-is and follow up.

packages/typm/src/typm.ts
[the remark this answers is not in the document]
mmkal: 🤖 This one is a false positive — the check-and-increment is atomic. `bytes` is computed synchronously (the `reduce` on line 141), and there is no `await` between the budget check (`totalBytes + bytes > maxTotalBytes`, line 142) and the increment (`totalBytes += bytes`, line 148) — the first await is line 149's `Promise.all`. JS runs synchronous code to completion before yielding, so each concurrent `acquireOne` resumes from its `await fetchJson` (line 115) and then runs lines 118–148 in one uninterrupted synchronous stretch. Every package that reaches the check sees a `totalBytes` that already includes all previously-charged packages, and commits its own charge before any other can interleave. So concurrent packages cannot jointly bust the budget. (The only bounded overshoot in this file is in `createLimiter`, and that concerns fetch concurrency, not the byte budget.) Resolving as no-change.

apps/os/src/components/repo-ide/repo-typescript.ts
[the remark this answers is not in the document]
mmkal: 🤖 Real edge case — fixed in 4432bd000. Deleting package.json within the 1s debounce left the armed timer to fire with stale text. `#maybeAcquireTypes` now clears the pending timer and resets `#lastRequestedPackageJson` on removal, so no stale re-acquisition; re-adding a package.json acquires again. (Evicting the *already-acquired* /node_modules types on removal is a separate, larger change — left as a possible follow-up; the practical impact is stale `any`-vs-typed imports until the next manifest, not incorrectness in the editor.)

packages/typm/src/typm.ts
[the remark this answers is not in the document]
mmkal: 🤖 This one's a false positive. The budget check (`totalBytes + bytes > max`) and the charge (`totalBytes += bytes`) are in the *same synchronous block* — there's no `await` between them (typm.ts:142→148). JS is single-threaded, so once an `acquireOne` reaches the check it runs through the increment before any other `acquireOne` can execute; concurrent calls from the wave's `Promise.all` serialize through that block and each sees the previous one's charge. The only awaits are the file-tree fetch (before the check) and the downloads (after the charge), so no two packages can pass the precheck against the same stale `totalBytes`. Leaving the code as-is.

apps/os/src/components/repo-ide/repo-typescript.ts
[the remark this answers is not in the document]
mmkal: 🤖 Follow-up: the deferred eviction half landed too, in b4bf69e32 — it turned out to be a simplification, not a larger change. A removed (or empty) package.json now runs the normal acquisition flow with an empty manifest (`"{}"`), so the worker's existing stale-path diff evicts everything it previously wrote, open buffers relint, and the dedicated removal branch (timer cancel + guard reset) disappears into the ordinary debounce logic. Zero network cost — an empty manifest has no deps to resolve. Both halves of this finding are now fixed.

packages/typm/src/typm.ts
```diff
@@ -0,0 +1,291 @@
...
+            warn(`could not download ${name}@${version}${file.name}`);
+            return null;
+          }
+          const content = await response.text();
+          files[`/node_modules/${name}${file.name}`] = content;
+          return { file, content };
+        }),
+      ),
+    );
+    const succeeded = downloaded.filter((download) => download !== null);
```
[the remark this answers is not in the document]
mmkal: 🤖 This is by design, and in the safe direction. The upfront charge (`totalBytes += bytes` before the downloads) is deliberate — it is what makes the byte budget also bound request **fan-out**, which was the whole point of the `PER_FILE_OVERHEAD_BYTES` change from the earlier review round. Charging only on download *success* would reopen the adversarial-listing hole: a hostile listing of thousands of files could pass the precheck and trigger the fetches before any accounting caught up.

The failure mode you describe — a package whose downloads all fail still consumes its estimate, so a later package might be skipped — is real but errs toward **under-fetching**, which is the correct bias for a resource guard (better to acquire too few types than to blow the worker's memory/CDN budget). The types-only, degrade-to-wildcard design means the cost of skipping is just wildcard-`any`, never a broken editor. Leaving as-is; resolving.

apps/os/src/components/repo-ide/repo-typescript.worker.ts
```diff
@@ -198,6 +251,62 @@ const api = {
...
+        });
+        const nextPaths = new Set(Object.keys(result.files));
+        // Warnings with NOTHING acquired usually means offline/CDN-down (the
+        // core degrades per-package, so total wipeout is transient-shaped).
+        // Reset so a retry is possible — retries stay cheap, they re-enter
+        // this dedupe — and keep the previous acquisition's types in place
+        // rather than degrading working buffers over a blip.
+        if (nextPaths.size === 0 && result.warnings.length > 0) {
+          lastAcquiredDependencies = null;
+          return { acquired: false, failed: true };
```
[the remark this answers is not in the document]
mmkal: 🤖 Real, though narrower than stated — I checked the branches:

- **Remove all deps** → `rootDependencies` is empty → the wave loop never runs → `result.warnings` is `[]`. So `nextPaths.size === 0 && warnings.length > 0` is **false**, and it correctly falls through to stale eviction (L291). Removed deps *do* degrade back to the wildcard. ✅
- **Change deps to only *unfetchable* specifiers** (e.g. `zod: ^3` → `zod: workspace:*`) → each is skipped-with-warning, so `0 files + warnings > 0` → this hits the transient-failure branch (L284) and keeps the old phantom types. ❌ That is the genuine gap: the `0 files + warnings` heuristic conflates "offline/CDN blip (transient, keep + retry)" with "all deps deterministically unfetchable (should evict)".

It only bites a fetchable→all-unfetchable flip, and the stale types are for versions the user just stopped depending on, so harm is low. The clean fix is *not* a one-liner: typm would need to report *why* it acquired nothing — e.g. return a `transient` flag (any thrown/`5xx` fetch) distinct from deterministic skips — so the worker can evict on deterministic-empty and only keep+retry on transient-empty. @mmkal — flagging rather than fixing since you're driving the merge; happy to do the typm-return-shape change as a scoped follow-up if you want it, but it's reasonable to merge and defer.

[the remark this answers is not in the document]
mmkal: 🤖 Agreed — the no-files-plus-warnings heuristic conflated "offline" with "the manifest genuinely has nothing acquirable anymore". Fixed in c8bc02e27, per-cause rather than run-level: the typm core now counts packages lost to THROWN fetches separately (`failures` on the result — offline/DNS/dropped connections), while answered non-ok responses (resolve 404s, unfetchable specifiers, no-types packages) stay mere warnings. The worker's transient-failure branch keys on `failures > 0`; with zero failures an empty result is authoritative, so it falls through to the stale-path diff and evicts types the manifest no longer supports. Core tests assert the distinction (answered 404 → `failures: 0`, thrown fetch → `failures: 1`).

packages/typm/src/typm.ts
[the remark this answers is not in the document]
mmkal: 🤖 Deliberate, so resolving without a code change — the upfront charge is what bounds request fan-out, which is spent whether or not downloads succeed. The 512-byte per-file floor exists precisely so a hostile flat listing can't trigger unbounded CDN fetches (see the earlier fan-out finding); refunding on failed downloads would hand that budget back and reopen the vector (a listing whose files all 404 would charge ~nothing while still costing a request per file). The estimate models attempted traffic, not stored bytes — slightly conservative in the all-downloads-fail case, but that case already warns per file and the whole run degrades safely. The budget is a fuse, not an allocator.

mmkal:
> 🤖 One heads-up before you merge: this PR is **stacked on #1771** (its base is `repo-ide-typescript-lsp`, not `main`) — typm builds on the TS-worker `setFiles`/`deleteFiles` seam that #1771 adds. So:
> 
> * If you merge **[Repo IDE: TypeScript language server (diagnostics, hover, autocomplete) #1771](https://github.com/iterate/iterate/pull/1771) first**, GitHub auto-retargets this one to `main` and it merges cleanly after.
> * If you merge **this one as-is**, it lands into the `repo-ide-typescript-lsp` branch, not `main`.
> 
> Both are updated off latest main and green. Want me to retarget #1781 to `main` (once #1771 is in) or otherwise sequence them for you?

Can you change the base branch for this to main and merge this PR - i.e. get both sets of changes in in one go. Then close the other PR with a note.

mmkal:
🤖 Done — retargeted to `main`, updated the body to note it includes #1771 in full (that's the squash commit message), and merged. #1771 closed with a pointer here. Both feature sets are on main in one commit.

mmkal:
🤖 All done — and you beat me to every step: retargeted to `main`, merged as `463087f96` (both the TS language server and typm in one squash), and closed #1771 with a note. Your three pre-merge commits (`4432bd0` cancel-timer-on-removal, `b4bf69e` evict-on-empty-manifest, `c8bc02e` transient-vs-honest-empty) fixed exactly the two real edge cases I flagged in the cursor threads (deleted-package.json timer; all-unfetchable-flip stale types); the other two cursor findings were the false-positive budget race and the by-design upfront charge. Both worktrees are already gone, so nothing to clean up. Standing down the monitor on these two.

mmkal:
🤖 Done — though you (or a concurrent session) had already carried it out by the time I picked this up, so this is really a confirmation of the end state I verified:

- **#1781 retargeted to `main` and squash-merged** (commit `463087f96`) with **both** changesets — the TypeScript language server (#1771) and typm — in one go. `main` now has both `packages/typm/**` and the repo-IDE TS worker (`repo-typescript.worker.ts` / `repo-typescript.ts`). I confirmed the merge was conflict-free via `merge-tree` against latest main.
- **#1771 is closed** with a note pointing at #1781 as the landing PR.
- Local worktrees for `repo-ide-typm` and `repo-ide-typescript-lsp` are cleaned up.

One tidy-up left if you want it: the remote `repo-ide-typescript-lsp` branch still exists (it backed the now-closed #1771 and wasn't auto-deleted since #1771 was closed, not merged). Its content is in `main` via #1781, so it's safe to delete — say the word and I'll remove it.

PR #1792, opened 2026-07-09
Streams deep-review fixes: all 8 bugs verifiably closed, perf, leak, dead code

jonastemplestein:
## What

Every issue from the three-agent deep review (#1787), verifiably fixed — all 8 bugs with a test pinning each, the top measurable perf items, the userspace state leak, and the dead-code sweep. Plus the two hazards the bugs reviewer flagged in passing.

## Bugs → tests

| Finding | Fix | Pinned by |
|---|---|---|
| HIGH-1: parked subscriptions hot-loop the DO alarm forever | `#park` clears the row's backoff; `#armAlarmFromStore` skips parked + in-flight rows | scenario `r`; storage test |
| HIGH-2: wake delivery failures never back off (close→re-poke spin; poke ack reset the counter) | `onDurableDeliveryError` runs the shared failure machine before closing; poke success advances the watermark via new `advanceWatermark` (keeps the streak) and only a PROGRESSED checkpoint resets it — sustained deterministic failure now parks at the same threshold as every lane | scenario `s` |
| MED-HIGH-3: selector-condition error facts self-feed on `error-occurred` events (unbounded log growth the pause door can't stop) | condition failures about error facts are never recorded (event still skipped) | scenario `v` |
| MED-4: in-flight drain acks silently clobber seeks (cursor-set / resume / replacement-deliver / remove+recreate) | cursor rows gain an `epoch` (bumped by every seek, fresh per row); drain acks are FENCED on the epoch they read; webhook re-checks config per event so a replaced URL stops mid-batch | scenarios `t`, `u`; storage tests |
| MED-5 (== perf 9, found independently twice): rebuild reads the whole log in one `.toArray()` → constructor OOM bricks big streams | catch-up/rebuild pages at 500 | (structural; exercised by every rebuild) |
| LOW-6: idle teardown over-acks past a wedged sink's unsettled batch | sinks count `pendingDeliveries`; wedged connections keep their watermark and get re-poked immediately | teardown suite |
| LOW-7: `#armAlarmNoLaterThan` get/set race | in-memory earliest-arm floor, reset when the alarm fires | (logic-level) |
| LOW-8: NaN/negative offsets → live-looking dead subscriptions | `subscribe` validates `replayAfterOffset`; the wake handshake requires an integer checkpoint ≥ 0 | lifecycle e2e |

[...]

PR #1766, opened 2026-07-08
Telegram integration: bot connect, webhook ingress, chat agents

mmkal:
Adds a **Telegram** built-in integration: connect a bot (token from @BotFather), and messages sent to it become agent conversations — the agent replies in the Telegram chat. Mirrors the Slack integration shape (named connections, directory-routed webhooks, router+agent processor pair, write-only token in a Secret DO, `itx.integrations.telegram["<connection>"].sendMessage(...)` for outbound).

```ts
// what an agent does to reply, once connected:
await itx.integrations.telegram["mishas-helper-bot"].sendMessage({
  chat_id, text: "on it 👍",
});
```

Key design choices (details in `apps/os/tasks/telegram-integration-for-os.md`, first commit):

- **No OAuth** — Telegram doesn't have it. Connect = paste bot token → `getMe` validate → Secret DO → `setWebhook`. New `connectTelegram` verb instead of contorting the OAuth state machinery.
- **Per-bot webhook URL** (`/api/integrations/telegram/webhook/<botId>`) because Telegram updates don't identify the bot. Verified via `X-Telegram-Bot-Api-Secret-Token`, **derived** as `hmac(SECRET_ENCRYPTION_KEY, botId)` — stateless, so **zero new Doppler config** in any environment.
- **URL-path secret substitution** — the Bot API wants the token in the URL (`/bot<token>/sendMessage`), so the egress door's `getSecret({...})` substitution is extended from headers-only to URLs. The one change outside the integrations domain; reviewers please eyeball this part hardest.

## 📱 Testing from your phone (against the preview slot)

> Status: ✅ **Ready for review.** Part 1 (integration) verified end-to-end by @mmkal from his phone; Parts 2 (threading: `/new` session rotation, reply-as-hint, journaled sends) and 3 (steal-with-confirm for cross-project bot moves) implemented, CI-green, and deployed to preview-2. Full spec + implementation log: `apps/os/tasks/complete/2026-07-09-telegram-integration-for-os.md`.

1. In Telegram, message **@BotFather** → `/newbot` → follow prompts → copy the token it gives you.
2. Open the preview OS URL — see the **Environment Config Lease** section below in this PR body (added automatically once the slot is claimed), e.g. `https://os.iterate-preview-N.com`. Sign in and open/create a project.
3. **Integrations → Telegram → paste token → Connect.** The card should flip to connected with the bot's username.
[...]

PR #1797, opened 2026-07-09
Streams: sqlfu-managed migrations for the subscriptions table (fixes no-such-column: epoch on live DOs)

mmkal:
**Puts the streams `subscriptions` table's schema under sqlfu-managed migrations, replacing the hand-rolled `pragma_table_info` + `ALTER TABLE` backfill that main ships today.**

Background: #1784 introduced the `subscriptions` table without an `epoch` column; #1792 queried `epoch` but `create table if not exists` never upgrades an existing table, so any Durable Object that created its table under #1784-era code threw `no such column: epoch` on every subscription read (reproduced on preview-2, which broke live bot connections: https://github.com/iterate/iterate/pull/1766#issuecomment-4925591724). #1806 hotfixed that with an inline `pragma_table_info` check + `alter table add column` in the constructor.

This PR replaces that hand-rolled check: the table's desired schema, migration history, runtime migrator, **and every SQL statement the store runs** now live in one static [sqlfu](https://sqlfu.dev) inline config:

```ts
export class SqliteSubscriptionCursorStore implements SubscriptionCursorStore {
  static db = defineConfig({
    definitions: sql`create table subscriptions (…, epoch integer not null default 0, …);`,
    migrations: [
      { name: "20260709000001_create_subscriptions", content: sql`create table if not exists subscriptions (…#1784 shape…);` },
      { name: "20260709000002_add_epoch", content: sql`…rebuild with epoch…` },
    ],
    queries: {
      get: sql.nullableOne<{ parameters: { subscriptionKey: string }; result: SubscriptionCursorRowRecord }>`
        select … from subscriptions where subscription_key = :subscriptionKey
      `,
      ack: sql.run<{ parameters: { subscriptionKey: string; ackedOffset: number; updatedAt: string } }>`
        update subscriptions set acked_offset = max(acked_offset, :ackedOffset), … where subscription_key = :subscriptionKey
      `,
      // …list, ensure, ackFenced, advanceWatermark, nack, setCursor, delete, minNextAttemptAt
    },
  });

  constructor(sql: SqlStorage) {
    this.#db = SqliteSubscriptionCursorStore.db(createDurableObjectClient({ sql }));
    this.#db.migrate();
  }

  ack(subscriptionKey: string, ackedOffset: number, epoch?: number): void {
    const params = { subscriptionKey, ackedOffset, updatedAt: new Date().toISOString() };
    if (epoch === undefined) this.#db.ack(params);
    else this.#db.ackFenced({ ...params, epoch });
  }
  // …
}
```

`migrate()` runs per Durable Object instance on construction (a deploy updates Worker code, not any live DO's private SQLite), applying only what that object's `sqlfu_migrations` bookkeeping says is missing — sqlfu's intended Durable Object model. Future schema changes are appended migrations with checksummed history, instead of more constructor-time shape sniffing.

Two deliberate choices in the history:

- **Migration 1 is the #1784 table verbatim, `if not exists`** — it *adopts* the table live DOs already created from raw constructor DDL, and creates it fresh everywhere else.
- **Migration 2 is a table rebuild (rename → create → copy → drop), not `alter table add column`** — with empty sqlfu history it meets three live shapes: no table, the pre-epoch #1784 table, and the with-epoch table that #1792-era constructors (and #1806's backfill) created. A plain ALTER would throw `duplicate column name` on the last one. Rows and cursor progress are preserved; epochs restart at 0, which is safe at construction time because no in-flight delivery (the only reader of a stale epoch) survives the DO restart that reruns the constructor.

The store's methods no longer build SQL strings inline: each statement is a typed named-parameter query on the config, bound once per instance (the old dynamic epoch-fence string split into static `ack`/`ackFenced`). Tests cover all three live shapes plus idempotence across reconstructions (recorded history short-circuits; minted epochs survive). `sqlfu@0.1.1` — already the version pinned by apps/auth and apps/semaphore — becomes an apps/os dependency; its root export is workers-safe by design.

### Also included: DO-reset park no longer container-disables kept sandbox classes

Getting this PR's preview e2e green surfaced a preview-slot infra bug, fixed here in `scripts/lib/do-reset.ts`. The slot-handover park must keep exporting container-bearing classes whose objects still exist (Cloudflare rejects the upload otherwise, error 10064 — first hit on preview-5, independently hit and worked around on main via #1766 for preview-8). But the obvious workaround — wrangler-deploying the parked stub with those classes stub-exported — **permanently container-disables their namespaces**: a legacy-migrations upload that exports a container class without naming it in `containers` metadata strips enablement, and no later upload can restore it (`Containers have not been enabled for this Durable Object class` from then on; this is what temporarily broke preview-3 and preview-5's sandboxes today, both since repaired via queue-consumer remove → `wrangler delete` → redeploy).

The park is now a raw script upload (same shape as `ensureContainerClasses`' bootstrap) that stub-exports kept classes **and names them in `containers` metadata**. This replaces main's #1766 workaround, which has the disabling behavior. Verified live on preview-3: erase → park → redeploy → sandbox-exec e2e green. Exports-mode parking (the 100403 fallback) was also verified to leave enablement alone.

| --- | ---: | ---: |
| Product | +186 -71 | +161 -68 🟩🟩🟩🟩🟥 |
| Tests | +35 -0 | +23 -0 🟩⬜⬜⬜⬜ |
| CI & scripts | +76 -57 | +45 -40 🟩🟥⬜⬜⬜ |
| Generated | +66 -0 | +63 -0 🟩⬜⬜⬜⬜ |
| **Total** | **+364 -128** | **+293 -108** 🟩🟩🟩🟩🟥 |

<sub>Lines counts every changed line; Significant ignores blank lines and JS comments. Between 449fc58 and daee07b, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>

## Environment Config Lease

<details>
<summary>Lease: preview-8 | Doppler config: preview_8 | Type: environment-config-lease | Leased until: 2026-07-10T21:20:09.512Z</summary>

| Auth | awaiting tests | `daee07b` | [https://auth.iterate-preview-8.com](https://auth.iterate-preview-8.com) | 819.2 KiB | 28.1s |  |  |  | [Workflow run](https://github.com/iterate/iterate/pull/1797) | 2026-07-09T21:20:13.759Z |  |
| OS | awaiting tests | `daee07b` | [https://os.iterate-preview-8.com](https://os.iterate-preview-8.com) | 4.10 MiB (-29.8 KiB vs main) | 132.7s |  |  |  | [Workflow run](https://github.com/iterate/iterate/pull/1797) | 2026-07-09T21:20:13.641Z |  |
| Semaphore | awaiting tests | `daee07b` | [https://semaphore.iterate-preview-8.com](https://semaphore.iterate-preview-8.com) | 440.8 KiB | 26.9s |  |  |  | [Workflow run](https://github.com/iterate/iterate/pull/1797) | 2026-07-09T21:20:13.552Z |  |
| Streams Example App | awaiting tests | `daee07b` | [https://streams.iterate-preview-8.com](https://streams.iterate-preview-8.com) | 1.29 MiB | 25.0s |  |  |  | [Workflow run](https://github.com/iterate/iterate/pull/1797) | 2026-07-09T21:20:13.643Z |  |

</details>

---

> [!NOTE]
> **Medium Risk**
> Touches Stream DO SQLite migration paths and delivery epoch fencing at DO restart; DO reset behavior changes affect preview slot teardown and container DO enablement.
> 
> **Overview**
> Replaces the streams **`subscriptions`** table’s constructor-time `pragma table_info` + `ALTER TABLE` epoch backfill with **sqlfu** (`defineConfig`, versioned migrations, typed queries). On each Stream DO construction, `migrate()` adopts three live SQLite shapes (missing table, pre-epoch #1784 table, with-epoch table from later DDL) via a create migration and an **epoch rebuild** that avoids duplicate-column errors while keeping cursor rows; store methods call named sqlfu queries instead of dynamic SQL (including separate `ack` / `ackFenced`).
> 
> Adds **`sqlfu@0.1.1`** to `apps/os`. Unit tests cover the post-#1792 with-epoch adoption path and assert **epochs survive** when the store is constructed twice on an already-migrated DB.
> 
> **DO reset** (`scripts/lib/do-reset.ts`) parks legacy-migration workers with a **raw Workers script upload** that lists kept container classes in upload **`containers` metadata** (not only wrangler deploy), so sandbox DO namespaces are not permanently container-disabled; API **100403** still falls back to exports tombstones.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit daee07b7adf4c6fb67f842e5f3b32ab242ec0c80. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


apps/os/src/domains/streams/stream-storage.ts
```diff
@@ -233,24 +234,90 @@ export type SubscriptionCursorStore = {
...
+            updated_at text not null
+          );
+          insert into subscriptions (subscription_key, acked_offset, attempt, next_attempt_at, last_error, updated_at)
+          select subscription_key, acked_offset, attempt, next_attempt_at, last_error, updated_at
+          from subscriptions_pre_epoch;
+          drop table subscriptions_pre_epoch;
+        `,
+      },
+    ],
+    queries: {},
```
mmkal: Add the actual sql queries used in this class here

apps/os/src/domains/streams/stream-storage.ts
```diff
@@ -233,24 +234,90 @@ export type SubscriptionCursorStore = {
...
-
-  constructor(readonly sql: SqlStorage) {
-    this.sql.exec(`
-      -- Delivery cursors for durable subscriptions (the spine's rows). One row
-      -- per subscriptionKey; created on subscription-configured, dropped on
-      -- subscription-removed. See stream-subscribers.ts for the state machine.
-      create table if not exists subscriptions (
+  /**
+   * Inline sqlfu config for the delivery cursors table (the spine's rows —
+   * one per subscriptionKey; created on subscription-configured, dropped on
```
mmkal: No need for this jsdoc

apps/os/src/domains/streams/stream-storage.ts
mmkal: 🤖 Done in 1f9f5b5c2 — all ten statements the class runs now live in the config as typed named-parameter queries (`get`, `list`, `ensure`, `ack`, `ackFenced`, `advanceWatermark`, `nack`, `setCursor`, `delete`, `minNextAttemptAt`), and the methods call the bound instance instead of raw `this.sql.exec`. The one dynamic string (ack with/without the epoch fence) became two static queries, `ack` and `ackFenced`.

apps/os/src/domains/streams/stream-storage.ts
mmkal: 🤖 Removed in 1f9f5b5c2.

PR #1871, opened 2026-07-11
Ephemeral stream events: second-class rows — excluded from default reads, invisible to durable subscribers, evictable

jonastemplestein:
## What

`append({ ..., ephemeral: true })` commits a **second-class stored row**: an ordinary offset-ordered event (same synchronous commit turn, same idempotency dedup, same circuit breaker, same pause door) with one `ephemeral` column flag and two deliberate demotions:

- **Excluded from reads by default.** Range reads — `getEvents`, `readEvents`, processor catch-up — skip ephemeral rows unless the caller passes `includeEphemeral: true`. Point reads by offset or idempotency key (an explicit request) always return them.
- **Never delivered to durable subscribers.** The wake/push/webhook lanes drop ephemeral events from delivery exactly the way selectors already skip non-matching events (skip-not-defer: cursors advance over their offsets), so subscription-fed processors never fold or side-effect on one. Ephemeral `subscribe()` connections receive them, live **and on replay** — the browser mirror stays dense and streaming text survives reloads.

The demotions are a license the stream keeps: because nothing durable can depend on an ephemeral row, a future sweep may **evict** them (memory pressure, DO-startup cleanup). Two constraints on that sweep are pre-paid: the offset allocator survives head-row eviction (`highestAssignedOffset()` reads AUTOINCREMENT\'s `sqlite_sequence`, which row deletion does not reset — empirically verified in workerd), and the browser mirror\'s continuity trigger is relaxed to monotonic-append (schema v5) with a loud `console.error` gap detector so a gap today (= lost delivery, since nothing evicts yet) is diagnosable rather than silent.

`stream/*` control facts cannot be ephemeral — config, presence, and park state may never be forgotten.

## Design history

The first iteration of this PR skipped storage entirely (opencode-v2 style: never a row, live-tail-only delivery). Review direction: since DO SQLite is synchronous, **store the rows and demote them at the read layer instead** — that deleted the whole accounting apparatus the unstored design needed (`maxDurableOffset` core-state split, `maxOffsetFloor` KV key, dual fresh-tail handoff, per-lane `streamMaxOffset` re-stamping, browser liveness-probe changes; `CORE_STATE_VERSION` stays 13) and turned at-most-once live delivery into replayable delivery for free. Net production diff: ~140 lines.

## Motivating use case, wired

`events.iterate.com/agent/llm-response-chunk` was already appended **durably** per SSE chunk — one row + chunk blob + idempotency index entry per streamed token, delivered to the project worker\'s `deliver:"all"` feed, forever. It is now `ephemeral: true` (idempotency key kept — dedup is rows again). Browsers/TUI still stream text live and on reload; default reads and the project-worker feed stop seeing chunk traffic entirely. **One deliberate behavior change**: userspace `processEventBatch` no longer receives chunk events (it was the one place a userspace author could observe them durably).

## Proof

- **Live on the PR preview (preview_7), driven by browser** ([agent-browser](https://os.iterate-preview-7.com)): created a project + web agent, ran two streamed turns. The browser mirror\'s raw view shows `#19–65 agent/llm-response-chunk ×47` arriving between `llm-request-started` and `output-added` (live chunk delivery on the ephemeral lane, in offset order); `itx` against the same stream returns `defaultCount: 60, rawCount: 463, chunksInDefault: 0, chunksInRaw: 403, ephemeralFlagged: 403` — 87% of that journal\'s rows are now excluded from every default read.
- New e2e (`streams.e2e.test.ts`) green against local dev + preview CI: consecutive offsets, self-describing flag, idempotency dedup on an ephemeral append, default-read exclusion vs `includeEphemeral`, live + replay delivery to ephemeral subscriptions, control-fact rejection, and a `kill()`-after-ephemeral-burst allocator proof.
- Unit: durable-lane drop with cursor advance (push + wake, storage and fresh-tail paths), combined `eventTypes`+`includeEphemeral` SQL, the pre-ephemeral live-table adoption path (pragma + `ALTER TABLE` — the riskiest deploy lines), `highestAssignedOffset` surviving head-row deletion, the strict emitted-input envelope accepting `ephemeral` (agent chunks depend on it). 1013+ root tests, typecheck, lint green.
- Two rounds of multi-agent thermo review (findings + fixes in the PR comment below).

No backwards compatibility needed; the one schema change on live streams is an idempotent `ALTER TABLE ... ADD COLUMN` adopted in the DO constructor, and the browser mirror rebuilds once (schema v5).

---

> [!NOTE]
> **Medium Risk**
> Touches stream commit, delivery, SQLite migration on live DOs, and agent streaming behavior; userspace processors no longer receive durable `llm-response-chunk` delivery, but core invariants are heavily tested.
> 
> **Overview**
> Introduces **`ephemeral: true`** on stream appends: rows stay offset-ordered and idempotent like any commit, but **default range reads omit them** (`includeEphemeral` opts in) and **wake/push/webhook delivery skips them** while cursors still advance (skip-not-defer). **`subscribe()`** connections still get ephemeral events live and on replay.
> 
> Storage adds an `ephemeral` column (live streams migrate via constructor `ALTER TABLE`), range queries filter with `ephemeral = 0`, and **`highestAssignedOffset()`** uses `sqlite_sequence` so a future head-row eviction cannot reuse offsets. Core replay/rebuild and spine reads use `includeEphemeral: true` where the fold needs those rows; **`stream/*` control events cannot be ephemeral**.
> 
> **`agent/llm-response-chunk`** is now appended ephemeral so browsers/TUI keep streaming while durable catch-up and the project-worker feed stop ingesting per-token rows; durable truth remains **`output-added`** / **`llm-request-completed`**.
> 
> The browser raw-events mirror (schema **v6**) relaxes the insert trigger from strict continuity to **monotonic offsets** (gaps for evicted chunks), logs unexpected gaps before eviction exists, and drops unused `maxOffset` / continuity summary helpers on the browser DB client.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e07a9d773f37bbdaaf80c1870e774f3465ced6aa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

| --- | ---: | ---: |
| Product | +153 -61 | +70 -46 🟩🟩🟥⬜⬜ |
| Tests | +292 -8 | +230 -2 🟩🟩🟩🟩🟩 |
| Docs | +32 -2 | +28 -2 🟩⬜⬜⬜⬜ |
| Generated | +33 -7 | +9 -3 🟩⬜⬜⬜⬜ |
| **Total** | **+510 -78** | **+337 -53** 🟩🟩🟩🟩🟥 |

<sub>Lines counts every changed line; Significant ignores blank lines and JS comments. Between 7671d16 and e07a9d7, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>

## Environment Config Lease

<details>

| Auth | released | `e07a9d7` | [https://auth.iterate-preview-7.com](https://auth.iterate-preview-7.com) | 819.5 KiB | 17.5s | 501ms |  | 2.9s | [Workflow run](https://github.com/iterate/iterate/actions/runs/215234101363278) | 2026-07-11T12:38:54.548Z | Preview app released. |
| OS | released | `e07a9d7` | [https://os.iterate-preview-7.com](https://os.iterate-preview-7.com) | 4.19 MiB (+1.6 KiB vs main) | 73.7s | 173.2s | 1 retried: DESIRED: a live-capability fetch handler serves WebSockets at an app host (vitest x1) | 11.0s | [Workflow run](https://github.com/iterate/iterate/actions/runs/215234101363278) | 2026-07-11T12:39:02.651Z | Preview app released. |
| Semaphore | released | `e07a9d7` | [https://semaphore.iterate-preview-7.com](https://semaphore.iterate-preview-7.com) | 440.8 KiB | 14.1s | 6.2s |  | 870ms | [Workflow run](https://github.com/iterate/iterate/actions/runs/215234101363278) | 2026-07-11T12:38:52.562Z | Preview app released. |
| Streams Example App | released | `e07a9d7` | [https://streams.iterate-preview-7.com](https://streams.iterate-preview-7.com) | 1.34 MiB | 15.2s | 18.4s |  | 725ms | [Workflow run](https://github.com/iterate/iterate/actions/runs/215234101363278) | 2026-07-11T12:38:52.421Z | Preview app released. |

</details>


jonastemplestein:
## Thermo-nuclear review — two rounds, findings + what changed

Ran two rounds of adversarial multi-agent review (server core / delivery spine / clients+docs / code-judo), each finding refute-verified before being applied. Summary:

### Round 1 findings (all fixed in `068ad6794`)

**MAJOR — the eviction license was false as written.** The design licenses a future sweep to delete ephemeral rows, but `highestOffset()` was `MAX(offset)` over surviving rows: evict the highest row, hit a core-state rebuild (version bump, or a crash inside the 64-event/1s checkpoint debounce), and the allocator reissues offsets that live subscribers already saw — every browser mirror then hard-ABORTs on "same offset, different JSON" until its next schema bump. Fix: `highestAssignedOffset()` reads AUTOINCREMENT's `sqlite_sequence`, which is updated by every insert (explicit offsets included) and **survives row deletion**; applied as a floor after the catch-up fold. Round 2 verified this empirically **in workerd itself** (miniflare probe, the repo's pinned version): the table exists immediately after `CREATE TABLE ... AUTOINCREMENT` (no first-insert race), reads are permitted, explicit inserts bump `seq`, deletion doesn't reset it. Pinned by a delete-the-head unit test.

**MAJOR — relaxing the browser mirror's continuity trigger silently deleted its lost-delivery detector.** Until eviction actually ships, every batch the browser receives is provably dense (the raw-events subscription is an unfiltered ephemeral subscribe), so any gap the new monotonic-append trigger accepts is a real bug that previously failed loudly into the resubscribe-from-checkpoint self-heal. Fix: a per-batch gap check that `console.error`s (local head vs batch first offset) while still accepting the batch — diagnosable today, tolerable after eviction ships.

**Comment-rot cluster**: four comments still described the strict continuity ABORT that no longer happens (the stale-checkpoint failure mode is now a *silent hole*, which is worse and needed saying); the schemas doc overclaimed "excluded from every read" (point reads always return them); the `local_index` comment described consumers that don't exist; the eviction license comments claimed safety nothing enforced. All rewritten to match the code.

**Dead code**: the diff had carefully re-commented `StreamBrowserDatabase.eventSummary()` — a method with zero callers. Deleted it, `maxOffset()`, their types, and the orphaned helper.

**Test gaps closed**: the pragma+`ALTER TABLE` live-table adoption path (the riskiest lines of the deploy — every existing stream takes that branch exactly once, and no test executed it); combined `eventTypes`+`includeEphemeral` SQL clause composition; ephemeral-through-fresh-tail on the wake lane; the strict emitted-input envelope accepting `ephemeral` (without that key in `getEventInputSchema` every agent chunk append would throw "Unrecognized key" — only live e2e would have caught a regression).

**Public surface**: the type-alias JSDoc that the generated itx api prints verbatim didn't mention the flag at all — a userspace author would see a bare `ephemeral?: true`. Documented + regenerated.

### Code-judo verdicts (examined, current shape kept — with reasons)

- **SQL-side exclusion in the spine's read hook** instead of JS delivery filters: would break cursor semantics — an ephemeral-only tail returns an empty page, the drain reads "caught up" while `ackedOffset < maxOffset`, and reconcile re-pokes every subscriber on every wake forever. Raw reads advance cursors; delivery filters. The two one-line filters are the minimal shape (test z's trailing-ephemeral assertions are exactly this regression trap).
- **Merging the two lane filters into `#applySelector`**: the drain and pump deliberately differ (drain appends durable condition-error facts; live lanes must never) — merging adds a mode flag, i.e. a thin abstraction.
- **Pragma migration vs sqlfu in the same file**: adopting sqlfu for the `events` table would replay the exact three-live-shapes trap the cursor store's own migration history documents, for one column. The 4-line idempotent pragma guard matches the constructor-DDL posture.
- **The "ephemeral" name collision** (event flag vs subscription lane): kept — both mean "not durable, forgotten without ceremony" and they interlock (ephemeral events ride ephemeral subscriptions); a second word would add a concept, not remove one. One pre-existing README sentence that now misparsed was reworded.

### Preview proof (browser-driven, preview_7)

Created a project + web agent via the real UI and ran two streamed turns:
- Browser mirror raw view: `#19–65 agent/llm-response-chunk ×47` between `llm-request-started` and `output-added` — chunks delivered live on the ephemeral lane, in offset order, and persisted in the mirror (streaming text survives reload).
- Same stream via itx: `defaultCount: 60, rawCount: 463, chunksInDefault: 0, chunksInRaw: 403, ephemeralFlagged: 403` — the default read is 87% smaller than the raw journal, which is the feature.

### Known follow-ups (not this PR)

- The eviction sweep itself (the license + allocator floor are in place).
- Ephemeral events still charge the shared circuit breaker — deliberate (chunk storms must trip it), but chunk-heavy streams eat burst budget; a separate rate class is a follow-up if it bites.
- Wake-mode subscribers behind an ephemeral-only stream tail carry a persistent watermark lag → one redundant poke per wake cycle; identical to the pre-existing consumes-selector shape, noted for whoever tunes poke coalescing.

PR #2062, opened 2026-07-17
Stabilize Streams virtual row scrolling

jonastemplestein:
## Summary

- keep collapsed event metadata on one line so every row matches the virtualizer 40px estimate
- add a focused regression for long event metadata
- run the existing 1,500-row scroll regression at the narrowest desktop breakpoint, where wrapping pressure previously made the flake deterministic

## Root cause

TanStack Virtual estimates collapsed rows at 40px. Event types and timestamps could wrap, so newly realized rows sometimes measured taller and triggered scroll-position compensation while the user was moving upward. That compensation appeared as the intermittent forward jump.

Truncating the variable event type and preventing the timestamp from wrapping restores the existing fixed-height contract without adding scroll machinery or relaxing the assertion.

## CI base repair

`main` currently contains an unrelated `oxfmt` failure in `tasks/os-custom-domain-provisioning-saga.md`. A separate formatter-only commit repairs that inherited file so this PR can have meaningful green CI. It can be dropped once `main` receives the equivalent repair already present in #2033.

## Proof

- before fix: long metadata row measured 88px instead of 40px
- before fix: stressed 1,500-row test reproduced a 20px forward jump with retries disabled
- after fix: both regressions passed 20 times each with 4 workers and no retries (40/40)
- full Streams Playwright suite: 31 passed, 1 intentionally skipped
- `pnpm typecheck`
- `pnpm lint`
- `pnpm test` (1,811 OS tests plus all other root workspace suites)
- `pnpm format:check`

| --- | ---: | ---: |
| Product | +4 -2 | +4 -2 🟩⬜⬜⬜⬜ |
| Tests | +27 -0 | +22 -0 🟩🟩🟩⬜⬜ |
| Docs | +17 -17 | +17 -17 🟩🟩🟩🟥🟥 |
| **Total** | **+48 -19** | **+43 -19** 🟩🟩🟩🟥🟥 |

<sub>Lines counts every changed line; Significant ignores blank lines and JS comments. Between 8b51afc and ef9dfc3, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>

## Environment Config Lease

<details>

| Streams Example App | released | `ef9dfc3` | [https://streams.iterate-preview-8.com](https://streams.iterate-preview-8.com) | 1.44 MiB | 73.6s | 56.0s | 1 retried: stream capnweb protocol > appends events after the stream-created event over capnweb (vitest x1) | 13.2s | [Workflow run](https://github.com/iterate/iterate/actions/runs/536510939728868) | 2026-07-17T11:31:06.610Z | Preview app released. |

</details>

### DIFF
--- a/apps/streams-example-app/e2e/playwright/stream-browser.spec.ts
+++ b/apps/streams-example-app/e2e/playwright/stream-browser.spec.ts
@@ -45,6 +45,30 @@ test("stream page appends through the shared browser mirror", async ({ page }) =
   await expect(eventMeta(page, type).first()).toBeVisible();
 });
 
+test("collapsed event rows keep a stable height with long metadata", async ({ page }) => {
+  const streamPath = `/e2e/${crypto.randomUUID()}`;
+  await page.goto(streamRoute({ path: streamPath }));
+  await expect(eventMeta(page, "events.iterate.com/stream/created").first()).toBeVisible();
+
+  const type = `events.iterate.com/debug/${"long-segment/".repeat(24)}`;
+  await appendComposerEvent(page, {
+    type,
+    payload: { streamPath, value: crypto.randomUUID() },
+  });
+
+  const metadata = eventMeta(page, type).first();
+  await expect(metadata).toBeVisible();
+  await expect
+    .poll(() =>
+      metadata.evaluate((element) => {
+        const row = element.closest('[data-testid="virtual-row"]');
+        if (!(row instanceof HTMLElement)) throw new Error("event metadata must be inside a row");
+        return Math.round(row.getBoundingClientRect().height);
+      }),
+    )
+    .toBe(40);
+});
+
 // The pre-itx-v4 hosted circuit-breaker processor (via the removed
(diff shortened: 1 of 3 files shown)


jonastemplestein:
Preview retry classification:

- Both virtual-scroll regressions passed on their first preview attempt; the full Streams browser suite was 31 passed / 1 intentionally skipped with zero Playwright retries.
- Preview telemetry also recorded one retry in the unrelated `stream capnweb protocol > appends events after the stream-created event over capnweb` Vitest. That test is deliberately the fleet fresh-Durable-Object weather canary.
- I immediately reran that exact test against the same Preview 8 deployment 20 times, with Vitest retries disabled and a fresh stream path each run: 20/20 passed, zero retries.

Classification: an isolated fresh-DO canary retry outside this diff, with no recurrence under a 20-run zero-retry probe. The retry remains visible in the PR evidence; no compatibility or recovery code was added to hide it.

PR #2161, opened 2026-07-20
Add preview slots 10 through 19

mmkal:
## Summary

- Expands the typed preview fleet from 9 to 19 slots, with real D1/KV IDs and slot-specific hostnames, Workers, routes, queues, buckets, and container caps.
- Adds targeted Auth/Doppler provisioning for slots 10–19 and updates the GitHub App manifest/runbook to match GitHub's accepted permission and event keys.
- Fixes Cloudflare Artifacts event wiring: deploy setup backfills exact-repo subscriptions instead of attempting an impossible account subscription per Worker; runtime repo creation and replacement install the exact subscription before pushing; seeded repository lookup skips the Cloudflare subscriptions API; atomic public imports recover their otherwise-missed initial push from the authoritative Artifacts head.
- Splits inbound Email Routing into a pre-deploy enable step and a required post-upload catch-all reconciliation, matching Cloudflare's requirement that the target Worker already exist.
- Declares missing React type peers for `lucide-react` and `streamdown`, preserving mobile's React 19.1 types while UI/web resolve 19.2 consistently.
- Requires preview clients to declare their allowed inventory and lets an eligible PR request one exact slot with `preview_environment=preview-17`, without turning the directive into an opt-in or bypassing Semaphore ownership; directive-driven moves are labelled accurately in the PR, interrupted moves resume without self-blocking, failed requests preserve the recorded working lease while releasing unrelated adopted leases, and fenced/commented examples stay inert.

> Compatibility prerequisite [#2163](https://github.com/iterate/iterate/pull/2163) is merged into `main`.

## Why each change was needed

| Change | Why it was needed | How we got by before / why it surfaced now |
| --- | --- | --- |
| Add slots in `envs.ts` | `envs.ts` is the typed source of truth for every deployed environment. Adding each `preview_<n>` entry and its real D1/KV IDs lets Wrangler generation, resource reconciliation, and Semaphore inventory recognize the ten new stacks without `UNPROVISIONED` placeholders. | The existing nine slots were already fully represented, so their generated configuration and leases worked. Slots 10–19 were intentionally invisible until this expansion provisioned their resources and committed the resulting IDs. |
| Target Auth/Doppler provisioning (`scripts/preview/preview.ts` and its test) | The existing command operated on every known preview slot. Adding ten entries would therefore have touched slots 1–9 too. `--slots` constrains mutations to the approved batch and rejects malformed or unknown selections. | “All known slots” matched the original bootstrap use case, where every slot was meant to be initialized together. During an incremental expansion it would unnecessarily revisit existing credentials, so the new batch needed an explicit mutation boundary. |
| Correct the GitHub App manifest | GitHub's manifest API uses different permission keys from several settings-UI labels, rejects `installation_repositories` as an explicit event, and requires `merge_queues: read` for `merge_group`. The corrected manifest is the shape GitHub accepted for these apps. | Existing GitHub Apps were already created and did not re-run manifest validation. Creating apps 10–19 exercised the current manifest endpoint again, which rejected the stale UI-derived keys and event list. |
| Mark deployed Workers and require their Cloudflare token (`generate-wrangler-config.ts`, `env.ts`, and tests) | Runtime repository creation must install an exact Artifacts event subscription. Deployed Workers therefore need the API token, while an explicit deployment marker prevents local development from mutating the shared Cloudflare account. | The old account-level subscription model did not require every Repo Durable Object to call Cloudflare. Moving to exact-repository subscriptions introduced that runtime responsibility, and therefore needed both credentials in deployments and a hard boundary around local development. |
| Replace per-Worker account subscriptions with exact-repository subscriptions (event subscription helpers, queue setup/entrypoint, Repo Durable Object, and tests) | Cloudflare permits only one account-level `artifacts` subscription, so creating one per preview Worker failed as soon as another slot owned it. Deploy setup now backfills existing repos only within its destination queue, and repo creation/replacement synchronously installs the exact subscription before pushing. Runtime lookup scopes to that queue and probes the name-sorted pages exponentially, then by binary search, so the ordering guarantee does not require a full account scan. | Existing environments were provisioned before this path was exercised fleet-wide, and one preview Worker already occupied the account singleton. Routine use did not attempt to create another account subscription; running clean resource ensures for the new slots exposed the invalid per-Worker assumption with Cloudflare’s 405 response. The first preview-6 proof then exposed a second historical assumption: with 29,018 subscriptions across 291 account pages (mostly preview-5), full pagination delayed fresh repository seeding until the 60-second request deadline. |
| Split Email Routing across resource setup and deployment (`email-routing-resources.ts`, `ensure-resources.ts`, `deploy.ts`, and tests) | Cloudflare rejects a catch-all Worker action before the target script exists. Resource setup now enables routing and explicitly defers a missing Worker—or a zone that has not been created yet—so create-only D1/KV reconciliation can finish. The post-upload step requires both and installs the catch-all, so deployment cannot silently accept missing inbound email. | Existing slots already had their OS Worker and zone when resource reconciliation ran, so the ordering requirement happened to be satisfied. A genuinely new slot exposed the first-deploy 404. The extracted helper initially made the formerly skippable pre-deploy missing-zone case fatal before ID reconciliation; the phase-specific outcome preserves bootstrap progress while keeping deployment strict. |
| Declare missing React type peers (`pnpm-workspace.yaml` and lockfile) | `lucide-react` and `streamdown` use React types without declaring the peer. In this workspace that let web/UI packages resolve mobile's React 19.1 types and broke the full typecheck. The package extensions make each importer resolve its matching React types while leaving mobile on 19.1. | The undeclared peer made success depend on the installed/hoisted dependency layout; the previous layout happened to supply compatible types. A fresh full-workspace verification for this PR resolved the mixed React graph and exposed the latent type conflict. |
| Gate expanded acquisition and add an exact PR-body selector (`allowedSlugs`, `preview_environment`) | Seeding slots 10–19 would otherwise make them immediately eligible for every old client. The compatibility server keeps omitted input on slots 1–9; this branch explicitly allows configured inventory and can pin a canary to slot 17 without force or fallback. Fenced Markdown examples and HTML comments are excluded before directive matching. | Before expansion, “all inventory” and “the safe inventory” were the same nine slots, so acquisition needed no rollout boundary. Adding unverified inventory separates those concepts and requires a server-first compatibility step. |
| Report requested slot moves accurately | A successful `preview_environment` move is intentional, so its PR notice names the directive instead of claiming the old lease lapsed and someone else took it. Unrequested moves retain the takeover warning. | Before exact selection existed, every slot move came from normal reacquisition and the takeover explanation was appropriate. The new directive introduced a second, intentional reason for changing slots. |
| Harden interrupted preview and repository creation recovery | A cancelled requested-slot acquire can leave a same-holder lease whose data was never recorded, so assignment now erases it before reuse. If a run dies after acquiring a requested slot but before releasing the old one, the retry reissues its own requested lease and completes the old-slot release instead of blocking on itself. If a requested acquisition fails, the recorded working lease is preserved while any unrelated lease adopted earlier in that attempt is released before the error propagates. Artifact subscription setup runs before the first ordinary push, while seeded repository lookup skips the Cloudflare API. An atomic public import necessarily pushes before its exact-repo subscription can exist, so OS subscribes, reads the authoritative imported head, and appends an idempotent equivalent push fact. | Normal random preview claims already erased unknown-provenance leases; the new requested-slot path exposed the missing equivalent check. Exact-subscription setup was also newly placed on a general get-or-create path, which made ordinary lookup depend on Cloudflare until the creation boundary was separated. Bugbot then identified the distinct atomic-import window, where no ordering change alone could capture the first push, the interrupted two-lease window introduced by exact slot moves, the failed-request lease leak, and the need to distinguish that unrelated lease from the recorded working lease. |

## Provisioning state

GitHub and Slack Apps, 50 Doppler configs, Auth credentials, D1/KV/R2/Queue resources, DNS, and Email Routing enablement already exist for previews 10–19. Every Auth, Semaphore, Dummy Petshop, and OS resource command passed a clean second ensure with the IDs committed here.

This PR is ready for review and its current head is running through the normal preview checks. Merging does **not** deploy the five apps, install the final email catch-alls, publish production Semaphore leases, or run the lease lifecycle proof. Those remain post-merge approval boundaries recorded in `tasks/create-preview-slots-10-19.md`.

## Verification

- `pnpm test` — all workspace projects green; OS: 2,116 passed, 6 expected failures, 1 skipped
- `pnpm typecheck`
- `pnpm lint`
- `pnpm format`
- 191 scripts tests (including 125 preview-orchestrator tests), 149 focused preview/config tests, and 65 Auth tests
- Generated five-app Wrangler blocks audited for every slot 10–19
- Four Cloudflare second-ensure commands passed for every slot 10–19

| --- | ---: | ---: |
| Product | +237 -66 | +202 -56 🟩🟩⬜⬜⬜ |
| Tests | +640 -19 | +588 -19 🟩🟩🟩🟩🟩 |
| CI & scripts | +341 -133 | +309 -118 🟩🟩🟩🟥⬜ |
| Config | +6 -0 | +6 -0 🟩⬜⬜⬜⬜ |
| Docs | +508 -22 | +463 -22 🟩🟩🟩🟩⬜ |
| Other | +60 -0 | +60 -0 🟩⬜⬜⬜⬜ |
| Generated | +216 -52 | +203 -52 🟩🟩⬜⬜⬜ |
| **Total** | **+2,008 -292** | **+1,831 -267** 🟩🟩🟩🟩🟥 |

<sub>Lines counts every changed line; Significant ignores blank lines and JS comments. Between cdc370e and 03cedba, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>

## Environment Config Lease

<details>

| Auth | released | `03cedba` | [https://auth.iterate-preview-6.com](https://auth.iterate-preview-6.com) | 810.1 KiB | 17.0s | 10.8s |  | 3.8s | [Workflow run](https://github.com/iterate/iterate/actions/runs/376226058053719) | 2026-07-21T10:19:50.370Z | Preview app released. |
| Dummy Petshop | released | `03cedba` | [https://dummy-petshop.iterate-preview-6.com](https://dummy-petshop.iterate-preview-6.com) | 198.2 KiB | 6.3s | 5.3s |  | 789ms | [Workflow run](https://github.com/iterate/iterate/actions/runs/376226058053719) | 2026-07-21T10:19:47.392Z | Preview app released. |
| OS | released | `03cedba` | [https://os.iterate-preview-6.com](https://os.iterate-preview-6.com) | 4.55 MiB (-9.3 KiB vs main) | 93.9s | 138.3s |  | 20.9s | [Workflow run](https://github.com/iterate/iterate/actions/runs/376226058053719) | 2026-07-21T10:20:07.458Z | Preview app released. |
| Semaphore | released | `ea33ed9` | [https://semaphore.iterate-preview-6.com](https://semaphore.iterate-preview-6.com) | 440.9 KiB | 13.9s | 20.8s |  | 1.1s | [Workflow run](https://github.com/iterate/iterate/actions/runs/555026251400949) | 2026-07-21T10:19:47.709Z | Preview app released. |
| Streams Example App | released | `ea33ed9` | [https://streams.iterate-preview-6.com](https://streams.iterate-preview-6.com) | 1.49 MiB | 104.7s | 57.8s |  | 17.7s | [Workflow run](https://github.com/iterate/iterate/actions/runs/555026251400949) | 2026-07-21T10:20:04.338Z | Preview app released. |

</details>

---

> [!NOTE]
> **Medium Risk**
> Changes deploy-time and runtime Cloudflare integration (Artifacts subscriptions, email catch-all, required API token) and preview slot leasing across the fleet; behavior is heavily tested but mistakes could break repo events, inbound mail, or preview assignment.
> 
> **Overview**
> Adds **preview slots 10–19** to `envs.ts` (OS/Auth KV/D1 IDs and Semaphore inventory) and updates docs/runbooks for a 19-slot fleet, targeted Doppler provisioning (`--slots`), and the corrected GitHub App manifest keys.
> 
> **Cloudflare Artifacts events** no longer create per-worker account-level subscriptions (only one per account). Deploy and runtime now use **per-repo** subscriptions: queue-scoped listing, name-ordered lookup without full scans, subscription before first push via `getOrCreateArtifact`, and **synthetic initial push** after GitHub import. Deployed workers get required `APP_CONFIG_CLOUDFLARE__API_TOKEN` and `DEPLOYMENT_ENV`.
> 
> **Inbound email** is split: `ensure-resources` enables routing and defers the catch-all until the worker exists; **post-deploy** installs and verifies the catch-all.
> 
> **Preview leasing** passes `allowedSlugs` to Semaphore, supports `preview_environment=preview-N` in the PR body (with accurate move notices and safer interrupted/failed slot assignment), and declares React type peers for `lucide-react` / `streamdown`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 03cedbac67c770293fa8a87dbcebd151df9319a7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


apps/os/docs/github-preview-app-manifest.md
```diff
@@ -38,7 +38,9 @@ current task.
 ## Manifest
 
 Replace every `N` with the slot number. Preview OS URLs use `.com`; `.app` is
-the hosted-project domain.
+the hosted-project domain. `default_permissions` uses GitHub's manifest/API
```
mmkal, line 41: 🤖 This correction was needed because GitHub's manifest API keys do not consistently match the permission labels in the settings UI. The previous names were rejected; `installation_repositories` cannot be explicitly subscribed, and `merge_group` also requires `merge_queues: read`. This is the manifest shape GitHub accepted for the new apps.

apps/os/scripts/deploy.ts
```diff
@@ -56,6 +56,7 @@ import {
 } from "./generate-wrangler-config.ts";
 import { ensureWorkerEventsQueue } from "./event-queue-resources.ts";
 import { ensureR2Bucket } from "./ensure-resources.ts";
+import { ensureInboundEmailRouting } from "./email-routing-resources.ts";
```
mmkal, line 59: 🤖 Email Routing can target a Worker only after its script exists. Running this reconciliation in `afterDeploy` gives the first deployment the required ordering, and `require-deployed-worker` makes a missing script or catch-all a deployment failure rather than tolerated drift.

apps/os/scripts/email-routing-resources.test.ts
```diff
@@ -0,0 +1,93 @@
+import { expect, it } from "vitest";
```
mmkal, line 1: 🤖 These tests make the ordering contract explicit: pre-deploy setup may defer only the catch-all for a genuinely missing Worker, post-deploy must install it, and post-deploy absence is fatal. That prevents the initial Cloudflare 404 from becoming a silent fallback.

apps/os/scripts/email-routing-resources.ts
```diff
@@ -0,0 +1,78 @@
+import { CloudflareApiError } from "../../../scripts/lib/env-context.ts";
```
mmkal, line 1: 🤖 This helper centralizes the two valid Email Routing phases. Cloudflare lets us enable routing before upload but rejects a catch-all Worker action until the script exists, so the caller must explicitly choose whether a missing Worker is temporarily allowed or a hard error.

apps/os/scripts/ensure-resources.ts
```diff
@@ -28,7 +28,7 @@ import {
 } from "../../../scripts/lib/deploy-helpers.ts";
 import { resolveEnvContext } from "../../../scripts/lib/env-context.ts";
 import { reconcileResources } from "../../../scripts/lib/wrangler-config.ts";
-import { emailDomainForDeployment } from "../src/domains/email/utils.ts";
+import { ensureInboundEmailRouting } from "./email-routing-resources.ts";
```
mmkal, line 31: 🤖 A brand-new slot has no OS Worker yet, so the old all-in-one setup failed while installing its catch-all. Resource setup now enables routing and records the one intentional deferral; `deploy.ts` completes it after upload.

apps/os/scripts/event-queue-resources.test.ts
```diff
@@ -1,6 +1,7 @@
 import { expect, it } from "vitest";
 
 import {
+  ensureArtifactRepoEventSubscriptionForWorker,
```
mmkal, line 5: 🤖 This regression coverage includes another Worker's account-level subscription and verifies that setup leaves it alone while creating only the exact-repository subscription for this Worker. It captures the Cloudflare singleton constraint that surfaced during slot provisioning.

apps/os/scripts/event-queue-resources.ts
```diff
@@ -47,33 +46,18 @@ async function ensureArtifactEventSubscriptions(
...
     });
-    logEnsure(
-      await ensureArtifactEventSubscription(ctx.cf, {
-        desired,
-        existing,
-        queueId: input.queueId,
-      }),
-      desired.name,
-    );
+    const result = await ensureArtifactEventSubscription(ctx.cf, {
```
mmkal, line 56: 🤖 The account-level subscription was removed because Cloudflare permits only one `artifacts` source per account, not one per Worker. Setup now backfills exact subscriptions for the repositories already present in this Worker's namespace.

apps/os/scripts/generate-wrangler-config.test.ts
```diff
@@ -173,6 +173,15 @@ it("requires the first-party PostHog project token in every deployed environment
   expect(OPTIONAL_SECRETS).not.toContain("APP_CONFIG_POSTHOG");
 });
 
+it("requires the Cloudflare API token and marks only deployed environments", () => {
```
mmkal, line 168: 🤖 This proves two operational invariants needed by runtime subscription setup: deployed environments always receive the Cloudflare token, and only deployed env blocks get `DEPLOYMENT_ENV`, so local development cannot mutate the shared account.

apps/os/scripts/generate-wrangler-config.ts
```diff
@@ -39,6 +39,7 @@ import { workerEventsQueueName } from "../src/queue-names.ts";
  */
 export const REQUIRED_SECRETS = [
   "APP_CONFIG_ADMIN_API_SECRET",
+  "APP_CONFIG_CLOUDFLARE__API_TOKEN",
```
mmkal, line 42: 🤖 The Repo Durable Object now creates Cloudflare subscriptions at runtime, so a missing API token would leave deployed repository events unwired. Making it a required deployed secret fails configuration early; the per-env marker separately keeps this behavior disabled locally.

apps/os/src/domains/events/cloudflare-event-subscriptions.ts
```diff
@@ -1,23 +1,13 @@
 import { workerEventsQueueName } from "../../queue-names.ts";
 
 /**
- * Cloudflare Artifacts event subscriptions expose account-level repo lifecycle
- * events and exact-repo Git activity events. The `artifacts.repo` source needs
- * the concrete artifact repo name, so we backfill existing repos at deploy time
- * and add repo subscriptions when account-level create/import/fork events land.
+ * An Artifacts account permits only one account-level `artifacts` source, so
```
mmkal, line 4: 🤖 Cloudflare allows only one account-level `artifacts` source, which cannot route independently to every preview Worker. The reusable exact-repo helper reconciles queue lookup, desired subscription shape, and existing state identically from deploy setup and runtime creation.

apps/os/src/domains/events/event-queue-entrypoint.ts
```diff
@@ -5,10 +5,7 @@ import { RepoArtifactNameCodec } from "../repos/utils.ts";
 import { workerEventsQueueName } from "../../queue-names.ts";
 import {
   createCloudflareAccountApi,
-  desiredArtifactRepoEventSubscription,
-  ensureArtifactEventSubscription,
-  listArtifactEventSubscriptions,
-  queueIdForWorkerEventQueue,
+  ensureArtifactRepoEventSubscriptionForWorker,
```
mmkal, line 8: 🤖 This path now uses the same exact-repository reconciliation helper as deploy setup and the Repo Durable Object. Centralizing it prevents the event consumer's recovery path from drifting onto a different subscription shape.

apps/os/src/domains/repos/repo-durable-object.ts
```diff
@@ -14,6 +14,10 @@ import { filterWorkerSnapshotPaths } from "../workers/source-masks.ts";
 import { walkWorkspaceFiles, wipeWorkspace } from "../../lib/shell-fs.ts";
 import { DurableObjectNameCodec } from "../durable-object-names.ts";
 import { parseConfig } from "../../config.ts";
+import {
```
mmkal, line 17: 🤖 Exact subscriptions must exist for repositories created or replaced after deployment. Installing the subscription synchronously before the first replacement/new push closes the event-loss window; `DEPLOYMENT_ENV` ensures local repos do not call the shared Cloudflare API.

apps/os/src/env.ts
```diff
@@ -21,6 +21,10 @@ export interface Env {
    * internal error).
    */
   WORKER_SELF: string;
+  /** Present only in generated deployed env blocks; absent in local dev. */
```
mmkal, line 24: 🤖 These bindings model the intentional split between runtimes: generated deployments always receive the marker and Cloudflare token, while local development omits them. The runtime guard uses that distinction to avoid shared-account mutations locally.

envs.ts
```diff
@@ -199,6 +199,56 @@ export const envs = {
     workerBuildCacheKvId: "3599fdcb79db418db0ead561f1ef85f7",
     authDbId: "ebf149cb-d3ed-48c5-a2d0-010166b25033",
   }),
+  preview_10: previewSlot(10, {
```
mmkal, line 202: 🤖 `envs.ts` is the typed source of truth consumed by config generation, resource reconciliation, and preview leasing. These entries use the provisioned D1/KV IDs so each new `preview_<n>` stack is addressable without placeholders or runtime discovery.

pnpm-lock.yaml
```diff
@@ -41,7 +41,7 @@ overrides:
   wrangler: 4.107.0
   ws: 8.19.0
 
-packageExtensionsChecksum: sha256-N5V8LT5a9y6jX4wQ8xKv5GrV6WtlFoB0HfQg8g4L82o=
+packageExtensionsChecksum: sha256-azEm6qUXEE53n66mxB6KYnijYIMdLGLjN0KkwrI2GsU=
```
mmkal, line 44: 🤖 This is the generated resolution produced by the package extensions in `pnpm-workspace.yaml`. It records web/UI on React 19.2 types while preserving mobile's React 19.1 graph, which is the separation the monorepo typecheck requires.

pnpm-workspace.yaml
```diff
@@ -87,6 +87,12 @@ packageExtensions:
     peerDependencies:
       '@typescript/vfs': '*'
       typescript: '*'
+  lucide-react:
```
mmkal, line 90: 🤖 `lucide-react` and `streamdown` consume React types without declaring that peer. In this multi-version workspace, pnpm otherwise allowed web/UI to inherit mobile's React 19.1 types. Declaring the missing peers makes each importer resolve the matching type version.

scripts/preview/preview.test.ts
```diff
@@ -46,6 +46,7 @@ const {
   releaseLeaseDespiteTeardownFailure,
   renderCloudflarePreviewPullRequestBody,
   resolveAuthPreviewRootSecret,
+  resolveProvisionAuthPreviewSlotNumbers,
```
mmkal, line 51: 🤖 The test protects the approval boundary: a selected range resolves only known slots, while an unknown slot fails before any Doppler mutation. That is what made it safe to provision 10–19 without touching existing slots.

scripts/preview/preview.ts
```diff
@@ -1235,21 +1235,59 @@ export async function reconcile() {
 type ProvisionAuthPreviewConfigsOptions = {
   /** Regenerate OAuth client secrets and app auth tokens instead of keeping existing values. */
   rotate?: boolean;
+  /** Comma-separated preview slots or ranges, for example `10-19` or `10-12,19`. */
```
mmkal, line 1265: 🤖 The previous provisioner iterated every slot known to `envs.ts`; merely adding ten entries would therefore have revisited slots 1–9. `--slots` scopes credential/config creation to the explicitly approved batch and validates the selection first.

mmkal:
In the table in the PR body can you explain how/why we were getting by before this change

mmkal (commented):
🤖 File-level rationale for the non-trivial changes.

mmkal:
🤖 Updated the rationale table with a third column explaining how the previous setup got by and why this expansion exposed each constraint. I also submitted the file-level rationale review, so those comments are no longer pending.

scripts/preview/preview.ts
```diff
@@ -236,27 +236,44 @@ async function deployPreviewApps({
...
+      ? (
+          await assignEnvironmentConfigLease({
+            eraseSlotData: makePreviewSlotDataEraser(runtime),
+            holder,
+            leaseMs: defaultPreviewLeaseMs,
+            recordedSlug,
+            semaphore,
+            wantedSlug: requestedEnvironment,
+          })
+        ).lease
```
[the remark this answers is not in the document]
mmkal: 🤖 Fixed in 6924fa619. Requested-slot assignment now applies the same provenance check as normal deployment claims: an adopted lease is kept without erasure only when it matches the PR’s recorded slot; otherwise the slot is erased before handover. Added a regression covering a cancelled acquire of requested preview-17, and all 112 preview tests plus scripts typecheck pass.

apps/os/src/domains/repos/repo-durable-object.ts
```diff
@@ -1557,19 +1562,40 @@ export class RepoDurableObject extends DurableObject<Env> {
...
     } catch (error) {
       // Only the race we mean to tolerate. The old blind catch masked real
       // failures (an INTERNAL_ERROR here fell through to get(), which then
       // reported a misleading NOT_FOUND).
       if ((error as { code?: string }).code !== "ALREADY_EXISTS") throw error;
       const existing = await this.requireArtifacts().get(name);
-      return { created: false, lastPushAt: existing.lastPushAt };
+      result = { created: false, lastPushAt: existing.lastPushAt };
     }
+    await this.ensureArtifactRepoEventSubscription(name);
```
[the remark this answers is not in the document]
mmkal: 🤖 Fixed in 074b1e184. Seeded existing repos now return without calling the Cloudflare subscriptions API. The extracted creation boundary still runs exact-subscription setup after a successful create and when ALREADY_EXISTS has no prior push, preserving recovery if a run died between creation and subscription. Public imports and replacements keep explicit setup at their creation boundaries. Added three boundary tests; all 2,048 OS tests, OS typecheck, and full lint pass.

apps/os/src/domains/repos/artifact-creation.ts
```diff
@@ -0,0 +1,27 @@
+/**
```
mmkal, line 1: 🤖 This boundary separates first-push obligations from ordinary repository lookup. It still runs subscription setup for a successful create or an unseeded recovery, but a seeded repo no longer depends on the Cloudflare subscriptions API.

apps/os/src/domains/repos/artifact-creation.test.ts
```diff
@@ -0,0 +1,64 @@
+import { expect, test, vi } from "vitest";
```
mmkal, line 1: 🤖 These tests distinguish all three states that matter: new creation, interrupted unseeded creation, and an already-seeded repo. The last case protects routine lookup from repeating first-push setup.

mmkal (commented):
🤖 File-level rationale for the Bugbot follow-up.

apps/os/src/domains/repos/repo-durable-object.ts
```diff
@@ -1475,6 +1481,7 @@ export class RepoDurableObject extends DurableObject<Env> {
         repo: input.repo,
       });
     });
+    await this.ensureArtifactRepoEventSubscription(artifactName);
```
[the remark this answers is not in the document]
mmkal: 🤖 Fixed in fad7a9026. Public imports now install the exact-repository subscription as soon as the atomic import is ready, read the authoritative imported branch head through the Artifacts binding, and append an idempotent queue-shaped initial-push fact. That preserves the processor’s single Artifacts-event normalization path without cloning history through the Worker. Added regression coverage for ordering and payload shape; all 2,114 OS tests, OS typecheck, and changed-file lint/format checks pass.

scripts/preview/preview.ts
```diff
@@ -241,27 +241,44 @@ async function deployPreviewApps({
...
+              notice: [
+                `All preview slots are leased — this PR is waiting in line for one (since ${new Date().toISOString()}).`,
+                holderTable,
+              ].join("\n"),
+            }));
+          },
+          recordedSlug,
+          semaphore,
+          waitTotalMs: resolveSlotWaitTotalMs(runtime.commandEnvironment),
+        });
```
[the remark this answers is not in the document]
mmkal: 🤖 Fixed in 4d3677167. Slot-move notices now classify a successful `preview_environment` move from the directive already parsed for acquisition: requested moves name that directive, while only unrequested moves use the lapsed/taken explanation. Added a regression for preview-6 → requested preview-17; all 187 scripts tests, scripts typecheck, and changed-file lint/format checks pass.

apps/os/src/domains/repos/artifact-import.ts
```diff
@@ -36,6 +36,68 @@ export async function importGithubArtifact(
...
+  await importGithubArtifactRepo(artifacts, input);
+}
+
+/**
+ * Import a public repo, then close the event-subscription race inherent in
+ * Cloudflare's atomic server-side import. The import's initial push happens
+ * before an exact-repo subscription can exist, so read the authoritative
+ * Artifacts head after subscribing and capture an equivalent push fact.
+ */
+export async function importGithubArtifactWithInitialPushCapture(
```
mmkal, line 48: 🤖 Cloudflare’s atomic import pushes before an exact-repository subscription can exist. This boundary subscribes immediately after readiness, reads the authoritative Artifacts head, and appends the equivalent idempotent push fact so commit/task processing keeps one event shape without cloning Git history through the Worker.

apps/os/src/domains/repos/artifact-import.test.ts
```diff
@@ -9,6 +12,73 @@ function fakeArtifacts() {
 }
 
 describe("importGithubArtifact", () => {
+  test("captures the imported head after installing the exact-repo subscription", async () => {
```
mmkal, line 15: 🤖 This regression locks down the only safe ordering for atomic imports: import and readiness, exact-repository subscription, authoritative head read, then queue-shaped capture. It also verifies the stable key and all-zero ‘new ref’ before SHA consumed by the existing processor.

mmkal (commented):
🤖 Import-event race rationale

apps/os/scripts/email-routing-resources.ts
```diff
@@ -0,0 +1,78 @@
...
+    throw new Error(`${ctx.name} has no project hostname base for inbound email`);
+  }
+
+  const zones = await ctx.cfV4<{ id: string; name: string }[]>(
+    `/zones?account.id=${ctx.env.cloudflareAccountId}&per_page=500`,
+  );
+  const zone = zones.find((candidate) => candidate.name === emailBase);
+  if (!zone) {
+    throw new Error(`no zone named ${emailBase} in account ${ctx.env.cloudflareAccountId}`);
+  }
```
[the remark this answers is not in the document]
mmkal: 🤖 Fixed in 366e69d9a. Pre-deploy resource setup now returns and logs an explicit `deferred-until-zone` outcome, preserving the create-only command’s ability to reach D1/KV reconciliation and print IDs. Post-deploy reconciliation still treats the absent zone as fatal, so deployment cannot silently succeed without inbound email. Added coverage for both phases; 14 focused tests, OS typecheck, and changed-file lint/format checks pass.

scripts/preview/preview.ts
```diff
@@ -4318,6 +4421,7 @@ async function assignEnvironmentConfigLease(input: {
     }
 
     const acquired = await input.semaphore.acquireSpecific({
+      allowedSlugs: previewEnvironmentSlugs,
```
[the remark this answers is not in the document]
mmkal: 🤖 Fixed in b6f91f08b. If Semaphore already attributes the requested slot to this PR, assignment now reissues that own lease with force (the same adoption mechanism already used for held leases), completes the normal erase boundary, and then releases the old recorded lease. Another holder is never force-evicted unless the caller explicitly passed `--force`. Added the interrupted two-lease regression; all 188 scripts tests, scripts typecheck, and changed-file lint/format checks pass.

scripts/preview/preview.ts
```diff
@@ -4305,24 +4408,28 @@ async function assignEnvironmentConfigLease(input: {
...
       leaseMs: input.leaseMs,
       slug: input.wantedSlug,
       type: ENVIRONMENT_CONFIG_LEASE_RESOURCE_TYPE,
       holder: input.holder,
-      force: input.force,
+      // Semaphore intentionally rejects every active lease without force,
+      // including another lease held by this same holder. Re-issue that one
+      // exactly like adoption, then release the old recorded slot below.
+      force: input.force || resumeInterruptedMove,
```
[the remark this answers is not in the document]
mmkal: 🤖 Fixed in 9d6b3288a. Exact-slot assignment now treats an unrelated adopted/retaken lease as scoped to the attempt: if requested acquisition or cleanup fails, it releases that lease before propagating the original error. If release also fails, an AggregateError preserves both failures instead of hiding the leak. Added the unrecorded-adoption/busy-request regression.

scripts/preview/preview.ts
```diff
@@ -1985,6 +2045,37 @@ export const environmentConfigLeaseInventory = previewEnvironmentSlotNumbers.map
...
+    throw new Error("PR body must contain at most one preview_environment directive");
+  }
+
+  const slug = matches[0]?.[1];
+  if (!slug || !previewEnvironmentSlugs.includes(slug)) {
+    throw new Error(
+      `Unknown preview_environment ${slug || "value"}. Expected one of: ${previewEnvironmentSlugs.join(", ")}`,
+    );
+  }
+  return slug;
```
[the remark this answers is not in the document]
mmkal: 🤖 Fixed in 9d6b3288a. The selector now removes HTML comments and fenced Markdown blocks before matching whole-line directives, so the documented example stays inert while one actionable prose line still works. Added coverage for fenced-only, commented-only, and fenced-plus-live bodies. All 190 scripts tests, scripts typecheck, and changed-file lint/format checks pass.

scripts/preview/preview.ts
```diff
@@ -4305,61 +4432,90 @@ async function assignEnvironmentConfigLease(input: {
...
+            ? `requested slot assignment failed; released unrelated held slot ${heldLease.slug}`
+            : `requested slot assignment failed; unrelated held slot ${heldLease.slug} was already gone`,
+        );
+      } catch (releaseError) {
+        throw new AggregateError(
+          [error, releaseError],
+          `Requested slot ${input.wantedSlug} could not be assigned, and releasing unrelated held slot ${heldLease.slug} also failed.`,
+        );
+      }
+      throw error;
```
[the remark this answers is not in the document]
mmkal: 🤖 Fixed in ea33ed902. A failed requested-slot move now preserves the PR’s recorded working lease and only releases a genuinely unrelated lease adopted during the attempt. The new regression failed before the guard and passes alongside the existing unrelated-lease cleanup regression; all 191 scripts tests, scripts typecheck, and changed-file lint/format checks pass.

apps/os/src/domains/events/cloudflare-event-subscriptions.ts
```diff
@@ -56,13 +63,70 @@ export async function listArtifactRepos(
...
   return await listCloudflarePages<ArtifactEventSubscription>(
     api,
-    `/event_subscriptions/subscriptions`,
+    `/event_subscriptions/subscriptions?queue_id=${encodeURIComponent(queueId)}`,
   );
 }
 
+/**
+ * Cloudflare exposes queue filtering and name ordering, but no name filter.
+ * Probe sorted pages exponentially, then binary-search the bracket: runtime
```
mmkal, line 75: 🤖 Why this lookup is bounded: preview-6 had to search an account containing 29,018 subscriptions across 291 pages before the first repository push, so the repository remained unseeded until the request timed out. Queue scoping removes cross-slot inventory; sorted exponential probing plus binary search bounds same-slot history while preserving the subscribe-before-push guarantee.

apps/os/scripts/event-queue-resources.ts
```diff
@@ -37,7 +36,7 @@ async function ensureArtifactEventSubscriptions(
   ctx: QueueResourceContext,
   input: { queueId: string; workerName: string },
 ): Promise<void> {
-  const existing = await listArtifactEventSubscriptions(ctx.cf);
+  const existing = await listArtifactEventSubscriptions(ctx.cf, input.queueId);
```
mmkal, line 39: 🤖 Why deploy reconciliation passes the queue ID: this path already knows its destination queue, and an unscoped account list would make every preview slot read every other slot’s exact-repository subscriptions.

apps/os/scripts/event-queue-resources.test.ts
```diff
@@ -86,14 +91,112 @@ it("paginates Cloudflare list endpoints with the accepted page size", async () =
...
+        repo_name: "project-10--repo",
+        type: "artifacts.repo",
+      },
+    },
+    method: "POST",
+    path: "/event_subscriptions/subscriptions",
+  });
+});
+
+it("finds one worker subscription without scanning every subscription page", async () => {
```
mmkal, line 151: 🤖 Regression purpose: the target is deliberately placed after multiple full pages. The assertion locks in page probes 1, 2, 4, 3 and no POST, preventing a return to linear full-account pagination or accidental duplicate creation.

apps/os/src/domains/events/event-queue-handler.test.ts
```diff
@@ -170,7 +170,7 @@ describe("event queue handler", () => {
         }
         if (
           path ===
-          "/client/v4/accounts/account-1/event_subscriptions/subscriptions?page=1&per_page=100"
+          "/client/v4/accounts/account-1/event_subscriptions/subscriptions?queue_id=queue-1&order=name&direction=asc&page=1&per_page=100"
```
mmkal, line 173: 🤖 Regression purpose: queue-delivered repository recovery must use the same queue-scoped, name-sorted lookup as synchronous creation; otherwise delayed events could reintroduce the account-wide scan.

mmkal (commented):
🤖 File-level rationale for the bounded subscription lookup.

apps/os/src/domains/events/cloudflare-event-subscriptions.ts
```diff
@@ -56,13 +63,70 @@ export async function listArtifactRepos(
...
+    if (current !== undefined) return current;
+    if (items.length === 0) {
+      upperPage = page - 1;
+      break;
+    }
+    const lastName = items.at(-1)?.name || "";
+    if (input.name <= lastName || items.length < pageSize) return undefined;
+    lowerPage = page;
+    page *= 2;
+  }
```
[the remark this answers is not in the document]
mmkal: 🤖 Fixed test-first in 03cedbac6. The regression places the existing subscription on skipped page 3 and later-sorting names on page 4; the old code attempted a duplicate POST. An overshooting probe now brackets the preceding pages and binary-searches them. Full OS tests (2,116 passing), typecheck, lint, and format are green.

apps/os/src/domains/events/cloudflare-event-subscriptions.ts
```diff
@@ -56,13 +63,75 @@ export async function listArtifactRepos(
...
+  let upperPage: number;
+  for (;;) {
+    const items = await readPage(page);
+    const current = match(items);
+    if (current !== undefined) return current;
+    if (items.length === 0) {
+      upperPage = page - 1;
+      break;
+    }
+    const firstName = items[0]?.name || "";
```
mmkal, line 108: 🤖 Why the first-name boundary is required: exponential page probes intentionally skip pages. Landing on names after the target proves only that the target is in an earlier bracket, not that it is absent; returning early here caused duplicate subscription creation.

apps/os/scripts/event-queue-resources.test.ts
```diff
@@ -86,14 +91,117 @@ it("paginates Cloudflare list endpoints with the accepted page size", async () =
...
```
mmkal, line 151: 🤖 Regression purpose: page 4 now contains later-sorting names rather than being empty, proving the search returns to skipped page 3 and finds the existing subscription without POSTing a duplicate.

mmkal (commented):
🤖 File-level rationale for the sorted-page boundary fix.

PR #2259, opened 2026-07-22
Move the GitHub AI linter into the iterate package

mmkal:
## Summary

Moves Iterate's GitHub pull-request linter out of project config and into the public `iterate` package.

Project config keeps its event hook explicit:

```ts
#aiLintApp = GithubAiLinter.create({
  policyVersion: "2",
  rules: { glob: "rules/**/*.md", repoPath: "/repos/iterate" },
});

protected override async processEvent(event: StreamEvent): Promise<void> {
  await this.#aiLintApp.processEvent(event, this.env);
}
```

The package owns subscription bootstrap, dynamic-worker hosting, webhook routing, and review-agent policy. Canonical review rules live in root `rules/**/*.md` and are loaded from one pinned repo commit per webhook. Existing durable keys, subscription keys, freshness gates, and idempotency keys remain stable.

## Companion config PR

https://github.com/iterate/config/pull/18 imports the packaged app and deletes its local bot copy. It is temporarily pinned to this branch's pkg.pr.new artifact and must merge after this PR.

## Test plan

- Package tests, typecheck, and build
- OS typecheck and config-template typecheck
- Existing GitHub review routing/processor suite
- Config template seed freshness
- Preview exercise after both PRs are available

Local results: full monorepo typecheck, lint, format check, and tests pass; the package build and 29 focused GitHub/template tests pass.

## Environment Config Lease

<details>
<summary>No preview slot recorded.</summary>

| app | status | commit | preview | size (gzip) | deploy duration | test duration | retries | cleanup duration | workflow run | updated | summary |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| Auth | released | `b9e1fc9` | [https://auth.iterate-preview-15.com](https://auth.iterate-preview-15.com) | 811.3 KiB | 29.4s |  |  | 3.9s | [Workflow run](https://github.com/iterate/iterate/actions/runs/517780948470181) | 2026-07-23T09:07:39.631Z | Preview app released. |
| Dummy Petshop | released | `b9e1fc9` | [https://dummy-petshop.iterate-preview-15.com](https://dummy-petshop.iterate-preview-15.com) | 198.3 KiB | 9.1s |  |  | 874ms | [Workflow run](https://github.com/iterate/iterate/actions/runs/517780948470181) | 2026-07-23T09:07:36.569Z | Preview app released. |
[...]

PR #2276, opened 2026-07-23
Publish review-bot package artifact

mmkal:
Temporary draft PR for pkg.pr.new recovery. It points at the exact head of #2259 so its SHA-scoped package artifact can be published after the original PR backend mapping began returning 404. Close after the publish job succeeds.


mmkal:
Exact-SHA package artifact published successfully; closing this temporary recovery PR. The real change remains in #2259.

PR #2309, opened 2026-07-24
Approvals batches: 1 event/push/faceid per Promise.all

mmkal:
A script run doing `Promise.all` over ~50 requests behind a `hold` egress rule used to mean ~50 pushes, ~50 taps, ~50 Face ID prompts. Now it means **one event, one push, one tap, one Face ID, one signature**.

https://github.com/user-attachments/assets/cb42c027-232d-47ca-b91f-f98021f52502

## What changes

Approvals are batches (ADR 0007):

- **The egress door batches.** Requests matching a `hold` rule park in an in-memory dataloader per (script run, rule) for the rule's new `debounceMs` (default **100ms**, each arrival extends, capped at 3×; `null` disables). The batch commits as ONE `human-approval-requested` event with a `requests[]` array. A lone request — or any hold without script-execution provenance — is a batch of one. Only concurrent requests can ever coalesce (a sequential caller's next fetch starts after the previous one is approved), so 100ms covers `Promise.all` bursts with no human-visible latency.
- **ONE signed decision.** `human-approval-decided` replaces granted+rejected: a verdict per index, `decidedBy: human|expiry`, at most one ECDSA signature over the **approval.v2** message (binds the batch offset, every request subject, and the verdicts). The door honors the FIRST decision, so grant-vs-stray-reject races are impossible by construction. All-reject (incl. expiry) is never signed; deny stays fail-safe. Settlement stays per released request (`{batch offset, index, status/error}`).
- **Everything downstream is simpler.** NotificationProcessor is stateless-per-event (one batch event → one push, `Script run waiting: 12 requests (12x gmail.googleapis.com)`, destination `{kind: "approvals", approvalRequestEventOffset}`). Mobile renders one card per batch with Approve all / Reject all and signs ONCE. `iterate approve`, `--json`, and the menubar all decide batches keyed by offset; the enclave dialog lists the batch and signs it in one Touch ID.

Example: what the approver signs for a 2-request batch —

```jsonc
// approval.v2 canonical message (signed once)
{ "v": "approval.v2", "projectId": "prj_…", "approvalRequestEventOffset": 35,
  "requests": [{ "method": "POST", "url": "https://gmail.googleapis.com/…", "bodySha256": "…", … },
               { "method": "POST", "url": "https://gmail.googleapis.com/…", "bodySha256": "…", … }],
  "verdicts": ["approve", "approve"] }
```

**Not backwards compatible, on purpose**: the `human-approval-granted`/`rejected` event vocabulary and the single-request payload shape are gone with no legacy parsing. Pre-merge approval events stop parsing and are silently ignored everywhere.

## Risk map — what could go wrong, and how to review

**Riskiest: the egress-door dataloader** (`project-durable-object.ts` `#pendingHoldBatches` / `#flushHoldBatch`). It's in-memory `Map` + `setTimeout` inside a Durable Object — the one piece whose correctness rests on a runtime argument rather than a test: parked fetches are in-flight requests, which keep the DO alive, so the timer always gets to fire; and if the DO dies pre-flush anyway, the queued fetch promises die with it — nothing recorded, nothing stranded. Review question: is there any path where a caller's promise is neither resolved, rejected, nor dead-with-the-DO? (`#flushHoldBatch` is written to settle every entry in all branches and never throw.)

**Trust boundary: `evaluateDecision` + `buildApprovalMessage`** (`egress-approvals.ts`). One signature now authorizes N requests, so the canonical message MUST bind every request subject and the verdicts, and the door must ignore (never fatal-fail) anything malformed: verdict-count mismatch, unsigned approve with keys enrolled, bad signature. This is the file to read slowly. The subtle policy choice: **all-reject decisions are accepted unsigned** (deny is fail-safe) — meaning anyone with stream-append access can veto a batch. Same as the old world (rejections were never signed), but now stated in one place.

**Settlement fan-out semantics changed**: approved indexes release concurrently and settle per index; approver UIs wait for ALL approved indexes before calling a batch done. A single hung upstream fetch keeps the whole batch "submitted — awaiting the egress door" until the settlement window lapses. Acceptable, but it's a behavior reviewers should know is intentional.

**On merge, expect**:
- Any pre-merge held approvals become invisible orphans (old event shape) — their parked fetches die with the deploy's DO restart, which they would have anyway. Approver screens on old data look empty; that's the no-legacy-parsing deal.
- Stale approver clients (an un-reloaded phone bundle, an old menubar) speak the dead vocabulary — their decisions are appended but ignored by the door. Symptom: "submitted" forever. Fix: reload/rebuild the client.
- The Swift changes (menubar rows, enclave batch dialog) are `swiftc -parse`-verified but not run — the menubar needs a real build before next use.

**Suggested review order**: ADR 0007 → `project-processor-contract.ts` (the three event schemas) → `egress-approvals.ts` (message + policy) → `project-durable-object.ts` (dataloader + decision wait) → the e2e (`egress-approvals.e2e.test.ts`, esp. the burst and mixed-verdicts lanes, which encode the intended semantics end to end) → UIs last, they're mechanical.

## Try it

Phone: **Approvals** → enroll → **Examples** → "Grouped approvals demo: a 12-request burst, ONE approval" (replaces the project's egress rules; use a disposable project). One push right after the burst → **Approve all 12** → one Face ID → twelve 200s.

## Follow-up queued (post-merge, first thing)

Rejection reasons: reject prompts for an optional reason that rides the decided event and comes back to the calling script/agent in the 403 body — spec'd in the task file.

Task: `tasks/complete/2026-07-28-grouped-approvals.md` · ADR: `apps/os/docs/adr/0007-approval-batches-at-the-egress-door.md`

Session: `4a701c88-5931-49e1-a0a5-970ccedfdd2a`

## Environment Config Lease

<details>

| Auth | released | `1b7593b` | [https://auth.iterate-preview-7.com](https://auth.iterate-preview-7.com) | 814.1 KiB | 16.9s | 9.6s |  | 4.1s | [Workflow run](https://github.com/iterate/iterate/actions/runs/527002300118345) | 2026-07-29T05:18:44.384Z | Preview app released. |
| Dummy Petshop | released | `1b7593b` | [https://dummy-petshop.iterate-preview-7.com](https://dummy-petshop.iterate-preview-7.com) | 198.3 KiB | 62ms | 6.0s |  | 1.0s | [Workflow run](https://github.com/iterate/iterate/actions/runs/527002300118345) | 2026-07-29T05:18:41.327Z | Preview app released. |
| OS | released | `1b7593b` | [https://os.iterate-preview-7.com](https://os.iterate-preview-7.com) | 4.91 MiB (-7.8 KiB vs main) | 47.2s | 191.5s | 1 retried: agent runs an itx script that appends a proof event, then replies (vitest x1) — expected [ 'stream/created', …(28) ] to include 'capability-host/script-run-settled' | 13.5s | [Workflow run](https://github.com/iterate/iterate/actions/runs/527002300118345) | 2026-07-29T05:18:53.771Z | Preview app released. |
| Semaphore | released | `5924ed8` | [https://semaphore.iterate-preview-7.com](https://semaphore.iterate-preview-7.com) | 441.1 KiB | 13.8s | 21.4s |  | 1.0s | [Workflow run](https://github.com/iterate/iterate/actions/runs/527002300118345) | 2026-07-29T05:18:41.317Z | Preview app released. |
| Streams Example App | released | `5924ed8` | [https://streams.iterate-preview-7.com](https://streams.iterate-preview-7.com) | 1.45 MiB | 67.5s | 48.8s |  | 8.4s | [Workflow run](https://github.com/iterate/iterate/actions/runs/527002300118345) | 2026-07-29T05:18:48.653Z | Preview app released. |

</details>

| --- | ---: | ---: |
| Product | +1,018 -606 | +660 -360 🟩🟩🟩🟥🟥 |
| Mobile | +800 -425 | +638 -354 🟩🟩🟩🟥🟥 |
| Tests | +659 -222 | +573 -192 🟩🟩🟩🟥⬜ |
| Docs | +284 -0 | +192 -0 🟩⬜⬜⬜⬜ |
| Generated | +51 -1 | +12 -1 🟩⬜⬜⬜⬜ |
| **Total** | **+2,812 -1,254** | **+2,075 -907** 🟩🟩🟩🟥🟥 |

<sub>Lines counts every changed line; Significant ignores blank lines, JS comments, and TypeScript lines with no runtime output. Between dad6a9a and 1b7593b, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>

---

> [!NOTE]
> **High Risk**
> Touches egress door batching, cryptographic approval policy, and a non-backward-compatible stream vocabulary; stale clients or in-flight holds on old shapes will see ignored decisions or empty UIs until reload/rebuild.
> 
> **Overview**
> **Human egress approvals are now batches end-to-end** (ADR 0007): concurrent `hold` traffic from one script run at one rule debounces at the Project DO (`debounceMs`, default 100ms) into a single `human-approval-requested` with a `requests[]` array; a lone or non-script hold stays a batch of one.
> 
> The event model is **breaking**: `human-approval-granted` / `human-approval-rejected` are replaced by **`human-approval-decided`** (verdict per index, `decidedBy`, optional one signature over **approval.v2**). Settlement stays **per approved index** (`human-approval-settled` adds `index`). Hold rules gain **`debounceMs`**; there is no legacy parsing of old events.
> 
> **Mobile** switches from `grant`/`reject` to **`decide`**, derives open/recent **batches**, and renders **`BatchCard`** with Approve all / Reject all and one sign path. **NotificationProcessor** is stateless again (one batch → one push with host summary). E2e and unit tests cover burst coalescing, mixed verdicts, and signature policy (`evaluateDecision`). Smaller UX fixes: foreground push banners, best-effort push revoke on sign-out, grouped-approvals example, docs/CONTEXT updates.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1b7593b1d0caad5f94fb78ac163b12321ca56a3c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mmkal:
🤖 Preview e2e failures on this PR are **not from this branch**. Root cause: freshly-created projects on the preview never become ready because the default project worker build fails with:

```
Failed to resolve 'iterate/github-ai-linter' from worker.ts: the installed 'iterate' package does not provide this entry
```

The config-repo template gained that import on main in #2277, but the `iterate` package installed by deployed project workers doesn't provide the entry yet — so every spec that creates/uses a project hangs on worker-ready (mass 1.5m timeouts, onboarding-smoke `stream-wait-timeout`). Verified by: reproducing with a bare project-create against preview 7 (`stream-wait-timeout` after 100s), tailing `os-preview-7` (build loop settling `source-failed`), fetching the failed project worker's overlay page (error above), and confirming the import exists in the template at the merge base. Local dev is unaffected because `iterate` resolves to repo source there.

#2307 (name-agnostic pkg.pr.new ref pinning for previews) is the fix — its preview passes the same project-creation specs this branch fails. Once it merges I'll merge main into this branch and re-run previews.

apps/os/scripts/cli.ts
```diff
@@ -6,6 +6,7 @@ import { createBuiltInPrompts, createCli, isAgent, yamlTableConsoleLogger } from
 import { isMainModule } from "@iterate-com/shared/dev/is-main-module";
 
 export * as configRepo from "./reset-config-repo.ts";
+export * as demoGroupedApprovals from "./demo-grouped-approvals.ts";
```
mmkal: let's rename to be more generic module

```suggestion
export * as approvals from "./approvals.ts";
```

And the "run" function becomes "demoGrouping"

So the cli would read better

iterate approvals demo-grouping

Don't do this change yet got some other requests coming

apps/os/scripts/approvals.ts
```diff
@@ -0,0 +1,110 @@
...
+  // The Cap'n Web WebSocket would otherwise keep the process alive.
+  process.exit(0);
+}
+
+if (isMainModule(import.meta.url)) {
+  void createCli({ ...import.meta, name: "demo-grouped-approvals", jsonInput: "auto" }).run({
+    logger: yamlTableConsoleLogger,
+    prompts: isAgent() ? undefined : createBuiltInPrompts(),
+  });
+}
```
mmkal: Why do we need this to be a separate cli

apps/os/src/domains/notifications/notification-intent-contract.ts
```diff
@@ -65,6 +65,24 @@ export const NotificationIntentContract = defineProcessorContract({
                   }),
               })
               .meta({ description: "The approvals screen, focused on one held egress request." }),
+            z
+              .strictObject({
+                kind: z.literal("approvals-group"),
```
mmkal: Hmm I wonder if this should now be the only thing, let's think on't

apps/os/scripts/cli.ts
mmkal: 🤖 Holding off per your note — will wait for the rest of your requests before touching code. Plan once you give the go: `scripts/demo-grouped-approvals.ts` → `scripts/approvals.ts` exporting `demoGrouping` (plus room for future approval-related subcommands), registered so the CLI reads `approvals demo-grouping`.

apps/os/scripts/approvals.ts
mmkal: 🤖 No strong reason — it's a thin wrapper that executes the same catalogue example the phone runs, and it got its own module only because it started life as a bigger standalone demo (localhost echo server) before being reduced to the wrapper. Folding it into a generic `approvals` scripts module as you suggest in the other thread is the right home; will do it together with that rename once your remaining requests are in.

apps/os/src/domains/notifications/notification-intent-contract.ts
mmkal: 🤖 Thought it through — I think yes, with one wrinkle. The obstacle to making `approvals-group` the only kind is scope-context holds: they have no `executionId`. But if the group key becomes "executionId for script runs, the request's own event offset for scope holds", then every push points at a group and a singleton is just a group of one — which the mobile screen already renders as a plain card. Scope holds would keep their immediate (non-debounced) push; only the destination shape unifies. Two costs: (1) already-delivered pushes carry the old `approvals` kind, so the phone must keep routing it for a while (cheap — leave the parser, delete in a later pass); (2) the notification body for a scope hold stays per-request, so the unification is purely in the destination contract, not behavior. Net: I'd do it — one destination kind, one deep-link path, no semantic change. Happy to fold it into the same pass as the CLI rename when you give the go.

apps/os/scripts/cli.ts
mmkal: 🤖 Done in e843bcee5: `scripts/demo-grouped-approvals.ts` → `scripts/approvals.ts` exporting `demoGrouping`, so it reads `pnpm cli approvals demo-grouping`. Room for future approval subcommands in the same module.

apps/os/scripts/approvals.ts
mmkal: 🤖 Folded into the generic `approvals` scripts module per the other thread (e843bcee5) — no more standalone CLI.

apps/os/src/domains/notifications/notification-intent-contract.ts
mmkal: 🤖 Superseded by the v2 rewrite (e843bcee5): the `approvals-group` destination kind is deleted entirely. The egress door now commits the burst as ONE `human-approval-requested` event, so the plain `approvals` kind (pointing at the batch event offset) is the only destination — no unification wrinkle left, scope holds are just batches of one. ADR 0007 has the full story.

apps/os/docs/adr/0006-debounced-notification-processor-for-approval-groups.md
```diff
@@ -0,0 +1,30 @@
+# Debounce Approval Group pushes in NotificationProcessor, don't rely on APNs threading or batch signatures
+
+**Superseded by ADR 0007**: batching moved upstream into the egress door —
```
mmkal: Don't add a new file that is already superseded by the time it's merged into main ffs

apps/os/scripts/approvals.ts
```diff
@@ -0,0 +1,116 @@
...
+// the "grouped-approvals-demo" catalogue example (src/itx/examples-source.ts)
+// — the SAME entry the phone's Examples screen runs, executed through the
+// same capabilityHost.runScript door, so there is one source of truth for
+// the demo. Works against any deployment: the burst targets our deployed
+// dummy-petshop service, not a laptop-local echo.
+//
+//   doppler run --config dev -- pnpm cli approvals demo-grouping --project prj_…
+//
+// CAUTION: the example REPLACES the project's egress rule list — use a
+// disposable/dev project.
```
mmkal: I don't think this script merits existing. We have the various buttons to "run examples" in both apps/os and apps/mobile

apps/os/docs/adr/0006-debounced-notification-processor-for-approval-groups.md
mmkal: 🤖 Deleted in 2079ad092 — ADR 0006 is gone; ADR 0007 now carries the rejected first-design context inline (no dangling reference), and the number stays unused on main (the gaps 0001/0003/0004 already exist, and main's config.ts has its own unrelated dangling "ADR 0006" refs from #1508 I didn't want to collide with).

apps/os/scripts/approvals.ts
mmkal: 🤖 Agreed and deleted in 2079ad092 — the catalogue example's Run buttons (web REPL examples panel + mobile Examples screen) are the demo surface; the CLI wrapper added nothing.

mmkal:
🤖 Latest preview run: the approvals suite is fully green (178 passing incl. all egress-approvals e2e; the OS playwright lane incl. the previously-failing guestbook spec now passes). Two remaining reds, both from main's new stream-subscriptions work (#2311), not this branch:

- `slack-agent.e2e.test.ts`: `stream-unavailable: deleted applied migration: 20260709000001_create_subscriptions` — that migration name exists nowhere in this tree or on main; slot 7's long-lived stream DOs applied it under an earlier deploy and current code no longer declares it. Stale-slot state vs migration churn; an `erase-data` on the slot (or a fresh lease) clears it.
- `streams-example-app` `stream-browser.spec.ts` "root stream route respects the selected view" — separate app, same feature area, fails with a retry consistently.

No action taken on either — flagging so red CI here isn't read as approvals breakage.

mmkal:
🤖 The one red check (preview deploy + e2e) is repo-wide main breakage, not this branch: the same two failures — `stream-browser.spec.ts › root stream route respects the selected view` and the OS Slack webhook e2e dying on `stream-unavailable: deleted applied migration: 20260709000001_create_subscriptions` — appear identically on other PRs' preview runs (e.g. depot run `nzgsdmt3vw`, different sha/branch). Likely fallout from #2311's subscriptions migration vs. preview slots' retained DB state. Everything else is green and all review threads are resolved.

apps/mobile/src/lib/approvals.ts
```diff
@@ -342,20 +331,75 @@ export async function awaitSettlement(
...
+    if (typeof ref !== "number") continue;
+    if (event.type === EVENT.decided) {
+      // Ascending order means the first decided we keep IS the one the door honors.
+      if (!decisions.has(ref) && Array.isArray(payload.verdicts)) {
+        decisions.set(ref, {
+          verdicts: payload.verdicts,
+          decidedBy: payload.decidedBy === "expiry" ? "expiry" : "human",
+          eventOffset: event.offset,
+        });
+      }
```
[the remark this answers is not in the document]
mmkal: 🤖 Confirmed and fixed in d30eff479. The door does length-check (#judgeDecision ignores a mismatched decision and keeps waiting), so the desync was real: mobile's indexApprovalEvents and the CLI's reconcileBacklog honored the first decided event with ANY verdicts array. Both now mirror the door's rule — a decided event only counts when its verdict count matches the batch it references, so a malformed decision leaves the batch visibly open and a later well-formed decision still wins. Regression tests added in both suites.

apps/os/src/domains/projects/project-durable-object.ts
```diff
@@ -407,113 +418,207 @@ export class ProjectDurableObject extends DurableObject<Env> {
...
+            entry.reject(error);
+          }
+        }),
+      );
     } catch (error) {
-      await settle({ error: error instanceof Error ? error.message : String(error) });
-      throw error;
+      // A failure before the verdicts fanned out (the requested append, the
+      // decision wait) fails every parked caller with the true error.
+      for (const entry of entries) entry.reject(error);
```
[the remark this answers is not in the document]
mmkal: 🤖 Confirmed and fixed in 1b7593b1d — Request reconstruction moved inside the per-entry try, so each fan-out callback is total: a throwing entry settles and rejects itself alone, the Promise.all can no longer reject, and the outer catch genuinely only sees pre-fan-out failures as its comment claims. No new test: an entry whose Request cannot be reconstructed is unreachable through the public surface (the gate parsed the original Request to park it at all), so this is structural hardening; the mixed-verdict e2e covers the fan-out behavior.

PR #2313, opened 2026-07-28
Harden rollout recovery, script execution, and preview readiness

jonastemplestein:
## Summary

- replace the fixed 90-second preview rollout clock with bounded recovery only for read-only or idempotency-keyed operations
- guard arbitrary scripts and credential-bearing egress at rollout boundaries before irreversible work can start
- preserve forward/backward RPC compatibility while accepting the same or a provably newer rollout target
- reacquire Project/Secret stubs after lifecycle resets and classify persistent receiver outages explicitly
- require two coherent, exact-version SSR HTML/asset cohorts before preview tests begin
- exponentially back off Durable Object rollout probes so readiness does not keep an old object continuously busy
- start every counted flakethon iteration from erased persistent preview data while retaining the canonical full-fleet Depot workflow, artifacts, and PostHog finalizer
- isolate every diagnostic retry's durable fixtures and remove a simultaneous cold-writer race from the follower test

## Latest evidence

The automatic exact-head preflight for merge head `d4d6dd995` was correctly rejected after 286 seconds (aggregate Depot run `zrkbw6q8r8`, canonical workflow `g06pnrrgxw`, preview job `k7vfndgxdq`, attempt `skhjf4bwnf`). Its retained artifacts normalized 7,191 PostHog events and contained six framework retries. Playwright finished 61 passed / 2 flaky in 86 seconds; Vitest finished 47 files passed / 1 failed / 3 skipped in 206 seconds. The telemetry finalizer and artifact upload both completed despite the terminal test failure.

The dashboard retry was a real rollout-coherence failure. The browser's HTML response came from the previous deployment, referenced 13 previous-version chunks, and every chunk returned 404 even though the request carried the expected version override. The existing readiness check had proved the new health endpoint and inventory assets, but never proved that a real SSR response and all assets emitted by that response belonged to the same cohort. Readiness now stamps real HTML with its immutable worker version and requires two consecutive exact-version HTML/asset cohorts.

The onboarding failures were also rollout-related. Cloudflare traces showed 91 probes against one Workspace Durable Object over 30 seconds, approximately every 250 ms, while every invocation remained on the old version. The probe loop now backs off from 250 ms to a four-second ceiling, giving the object quiet periods in which deployment handoff can occur.

The terminal cache failure was environmental contamination, not the named cache assertion: both attempts timed out waiting for the first project's onboarding greeting, so neither created the second project. In the same window Cloudflare emitted 810 hosted-processor callback timeouts across roughly 375 projects, mostly project IDs absent from the current run's artifacts, and the backlog continued after CI ended. Because a PR keeps the same preview slot for its lifetime, sequential flakethon dispatches had accumulated durable state across repetitions. A counted iteration now explicitly erases the full leased preview environment before deploying the full fleet.

The complete diagnosis and immutable-head history are in `docs/preview-e2e-flake-hunt.md`.

## Safety boundary

Read-only waits retain their cursor; keyed writes retain their idempotency key. Each may absorb one classified deploy/eviction reset on a freshly acquired stub. Explicit public `kill()`, application failures, raw overload, unkeyed writes, second resets, and already-started arbitrary work remain terminal.

Arbitrary script execution and credential-bearing egress are never blindly replayed. Their read-only rollout probes happen before the guarded side effect, and only the same or a provably newer target is accepted.

Fresh-data erasure is restricted to explicit manual `workflow_dispatch` input, requires a full-fleet run, and is enabled by the strict flakethon driver. Ordinary PR preview updates retain their existing state.

## Validation

- `pnpm test`: green across all workspaces; OS 2,486 passed, 12 expected failures, 1 skipped across 250 files
- `pnpm typecheck`: green across all 17 workspace projects
- `pnpm lint`: zero warnings and zero errors
- `pnpm format`, Knip, shell syntax, and `git diff --check`: clean
- focused preview/policy suite: 152/152 passed
- focused readiness/version-stamping suite: 17/17 passed
- immutable candidate: `cf7899f23f148ed66ec9af5ee745cc8ac8d37e67`

## Flake-athon

```bash
PR_NUMBER=2313 REF=flake-athon-rollout-recovery RUNS=25 EXPECTED_HEAD_SHA=cf7899f23f148ed66ec9af5ee745cc8ac8d37e67 ./scripts/preview/flake-hunt-loop.sh
```

Every counted iteration is a separate dispatch of canonical `.depot/workflows/cloudflare-previews.yml` on the normal 16-core Depot runner. It starts from clean persistent preview data and emits the ordinary artifacts and PostHog finalizer, including on failure. The exact-SHA `pkg.pr.new` GitHub Actions run is an external-package prerequisite and is never counted as a flakethon iteration. Any workflow failure, moved head, framework/transport retry, or duration at or above seven minutes rejects the streak.

Current strict streak: **0/25**, pending a clean-state exact-head preflight.

## Environment Config Lease

<details>
<summary>Slot: preview-12 | Doppler config: preview_12</summary>

| Auth | deployed | `c284b1d` | [https://auth.iterate-preview-12.com](https://auth.iterate-preview-12.com) | 814.1 KiB | 19.5s | 10.7s |  |  | [Workflow run](https://github.com/iterate/iterate/actions/runs/113935669160539) | 2026-07-29T07:34:02.960Z |  |
| Dummy Petshop | deployed | `c284b1d` | [https://dummy-petshop.iterate-preview-12.com](https://dummy-petshop.iterate-preview-12.com) | 198.3 KiB | 6.0s | 5.8s |  |  | [Workflow run](https://github.com/iterate/iterate/actions/runs/113935669160539) | 2026-07-29T07:33:58.046Z |  |
| OS | deployed | `c284b1d` | [https://os.iterate-preview-12.com](https://os.iterate-preview-12.com) | 4.93 MiB (+6.5 KiB vs main) | 118.8s | 196.8s | 1 retried: a second project's onboarding turn is served from the AI Gateway cache (vitest x1) — expected { cacheStatus: 'HIT', …(1) } to match object { Object (greeting) } (1 matching property omitted from actual) |  | [Workflow run](https://github.com/iterate/iterate/actions/runs/113935669160539) | 2026-07-29T07:37:09.079Z |  |
| Semaphore | deployed | `13cbabb` | [https://semaphore.iterate-preview-12.com](https://semaphore.iterate-preview-12.com) | 441.1 KiB | 12.3s | 22.2s |  |  | [Workflow run](https://github.com/iterate/iterate/actions/runs/113935669160539) | 2026-07-29T07:34:14.463Z |  |
| Streams Example App | deployed | `c284b1d` | [https://streams.iterate-preview-12.com](https://streams.iterate-preview-12.com) | 1.46 MiB | 75.4s | 50.6s |  |  | [Workflow run](https://github.com/iterate/iterate/actions/runs/113935669160539) | 2026-07-29T07:34:42.928Z |  |

</details>

| --- | ---: | ---: |
| Product | +3,532 -842 | +2,340 -473 🟩🟩🟥⬜⬜ |
| Tests | +4,501 -709 | +4,059 -616 🟩🟩🟩🟩🟥 |
| CI & scripts | +611 -121 | +510 -81 🟩⬜⬜⬜⬜ |
| Config | +0 -1 | +0 -1 🟥⬜⬜⬜⬜ |
| Docs | +777 -65 | +705 -65 🟩⬜⬜⬜⬜ |
| Other | +8 -19 | +7 -17 🟥⬜⬜⬜⬜ |
| Generated | +108 -30 | +41 -6 🟩⬜⬜⬜⬜ |
| **Total** | **+9,537 -1,787** | **+7,662 -1,259** 🟩🟩🟩🟩🟥 |

<sub>Lines counts every changed line; Significant ignores blank lines, JS comments, and TypeScript lines with no runtime output. Between d7f41f1 and c284b1d, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>


jonastemplestein:
## Small reliability fixes extracted

I mined this PR hunk-by-hunk and have now extracted eleven genuinely bounded improvements into four clean PRs:

- #2354 — the first six, squash-merged as `4decd182`.
- #2361 — the next three, squash-merged as `fbe65d21` after a green preview, zero review threads, a green production deploy, and production verification.
- #2362 — the focused hosted-processor catch-up/recovery candidate, independently implemented while this extraction was under way and squash-merged as `bdfb9abb`; I audited it against the candidate and completed its production proof.
- #2363 — the two-tab contract split, squash-merged as `d9ece64f` after a green preview, zero review threads, green mainline, and direct production execution.

### Taken in #2354

1. **Predicate event waits now await the real waiter promise.** The private helper was also renamed to make its registration-only return shape explicit, with a regression proving a predicate wait stays pending until a matching future event is delivered.
2. **The code-mode fence e2e now follows its own execution.** It derives the exact execution ID from the synthetic assistant event, selects only that settlement, parses the canonical nested settlement schema, and requires `succeeded`.
3. **Retry attempts get coherent fresh fixtures.** GitHub App, userspace/Ocado, and Petshop identities now originate inside each Vitest callback, so a retry cannot combine durable/write-only state from attempt 1 with new keys, OAuth clients, or egress origins from attempt 2. For Petshop I used a completely fresh attempt instead of the memoization in this PR.
4. **Egress race tests preserve the causal failure.** Their barriers race the actual request and a bounded 30-second synchronization deadline; early rejection is rethrown, early success fails explicitly, and held providers are released in `finally`.
5. **The ephemeral-events example opts into bounded append replay safely.** Both appends receive distinct idempotency keys under one invocation UUID; generated artifacts and an executable regression were updated.
6. **The stream composer e2e waits for the exact operation's terminal state.** The helper is scoped to its composer, waits for `appended` or `error`, fails immediately on the wrong terminal state, and the paused-gate case explicitly expects `error`.

### Taken in #2361

7. **Dummy-petshop RSA fixtures are generated once in bounded suite setup.** The valid and attacker 2048-bit keypairs are generated in parallel in `beforeAll`, preserving the distinct bad-signature key and every behavior assertion. The focused file fell from 1.33 s to 156 ms in the baseline comparison.
8. **Preview retry annotations now tell the truth.** They use the existing `passedAfterRetry` result: recovered retries stay notices, still-failing retries are warnings, and the output explicitly says the test command exit remains authoritative. Focused tests cover both outcomes and the high-retry threshold.
9. **Opaque hosted-processor/capnweb failures are now attributable without a hidden redial.** The originating ITX call ID survives the one-way wake boundary; safe Cloudflare reset references and exact project, stream, processor/version, connection generation, cursor, batch, delivered-through/head, and pending-deadline fields are logged. Retry, backoff, close, and delivery semantics are unchanged.

### Taken in #2362

10. **Hosted-processor catch-up is bounded to the work each receiver can acknowledge.** Each acknowledgement advances one matching hosted event, with a 100-row raw scan cap; ReviewBot receives only its lifecycle/mention protocol events, the linter receives its six protocol events, and the redundant raw-webhook replay subscription is removed. Existing restore cutoffs are preserved. This fixes the diagnosed replay overload rather than increasing the 20-second acknowledgement timeout or adding a hidden reconnect loop.

#2362 landed independently while I was preparing this extraction, so I did not duplicate it. I reviewed its exact diff against the candidate, waited for green CI and zero review threads, and supplied the production-shaped state/trace proof on the PR.

### Taken in #2363

11. **The two-tab test now has two independent, attributable contracts.** A dedicated simultaneous cold-bootstrap test opens both tabs concurrently, requires exactly one writer and one reader, proves a reader-originated append reaches both tabs, closes the writer, requires reader promotion, and proves appends continue after handoff. The separate 2,000-event/DO-kill test establishes writer then reader sequentially before exercising reader-originated appends and exact counts in both browser databases. This removes an unrelated setup race from the heavy recovery proof without deleting concurrency coverage.

These changes were reimplemented against current `main` without compatibility shims or PRD/rollout baggage.

### Production proof for #2361

- Preview deployed all five apps and passed 64 Playwright tests plus 48 OS Vitest files / 179 tests, including all 35 stream-connection cases. The real workflow emitted the new truthful recovered-retry notice.
- Main deployed #2361 to `os-prd` as Worker version `6020919c-875e-4ff0-94b5-85085b8045ee`; every deploy/check completed green.
- A targeted production stream failure-isolation e2e passed in 26.56 s with no retry (`os-vitest-run-20260729-182505`).
- Real rollout traffic exercised the new diagnostic fields immediately. The timeout burst existed on the pre-change Worker, crossed #2361's deploy and the immediately following main deploy, and then stopped; an exact production query from 17:29 through 17:32:36 UTC found zero further rows. Both sampled projects had reopened hosted connections at stream head with no pending delivery, cleared error state, and zero retry attempts.

Full preview proof and production classification are recorded on #2361.

### Production proof for #2362

- `bdfb9abb6` deployed to `os-prd` as Worker version `04d36fa1-0f59-4e7e-bed2-9573654cb85b`; every deploy smoke passed.
- Across that deployment and an immediately following unrelated main rollout, exact production queries found zero recurrence of `stream durable callback failed; backing off before waking it again`. The pre-existing overload tail drained and stopped; bounded reset/read recovery remained visible outside the error signal without becoming terminal failures.
- All five affected project roots were reachable and unpaused. All 15 live hosted connections were exactly at stream head with lag 0 and no pending delivery; stored subscription retry state was clear (`attempt: 0`, no deadline, no error).
- ReviewBot retained its restore cutoff, had the narrowed condition, and delivered to head after both rollouts.

The full query windows, trace link, per-project state, and alternative-cause falsification are recorded in [#2362's production proof](https://github.com/iterate/iterate/pull/2362#issuecomment-5121914325).

### Preview and production proof for #2363

- Local baseline: the unchanged coupled heavy test passed 10/10. The split contracts passed 20/20, plus 6/6 under CPU/network throttling. The full app lane passed 21/21 Vitest and 31 Playwright tests with one intentional skip and zero local retries; the full repository gate passed.
- Preview deployed `streams-example-app-preview-4` as Worker version `ed2ca664-0f87-4f64-8841-651335a90679`. Both changed tests passed first try; the complete Playwright lane passed 31 with one intentional skip. One disjoint node-hosted Vitest test used the repository's single modeled edge/platform retry after a WebSocket connection failure and then passed.
- #2363 had zero review threads, AI-linter approval with 0 errors/0 warnings, and every PR check green. Mainline deploy, Test, Lint and Typecheck, CI telemetry, autofix, publish, and notification checks also completed green.
- `d9ece64fb` deployed to `streams-example-app-prd` as Worker version `1ed84a67-bed1-435a-b359-8ee3b3554e86`; deploy smoke and `/api/__internal/health` returned 200.
- The two changed contracts were then run directly against production three times each: **6/6 passed with no Playwright retry**. The concurrent test proved exact writer/reader election plus handoff; the 2,000-event test proved exact convergence in both browser databases after the deliberate DO kill.

The production trace audit also found a real, pre-existing namespace defect: the two persistent `/` and `/e2e` ancestor Stream DOs still carry the deleted `20260709000001_create_subscriptions` migration from before #2311. Fresh `/e2e/<uuid>` test streams work, but their observed background ancestor announcements fail against those two old DOs. This is not caused by #2363's test-only diff and does not invalidate the exact-count assertions, but it means the Streams Example App production namespace is not telemetry-clean. The clean follow-up is destructive data recreation per #2311, not a compatibility shim or swallowed error. I did not erase production data as part of a test-only PR.

The full command, result, trace correlation, and two-DO classification are recorded in [#2363's production proof](https://github.com/iterate/iterate/pull/2363#issuecomment-5122187118).

### Not taken

- **The broad rollout-recovery system:** Durable Object rollout gates/readiness, same-or-newer compatibility, stub reacquisition, project birth, script settlement recovery, sandbox admission, append hedging, coherent SSR cohorts, preview erasure/flakethon machinery, and related docs/config. #2362 takes the one bounded, diagnosed catch-up defect; the rest remain coupled architectural changes, not small no-regrets extractions.
- **Relaxing lifecycle/wake counts to `>= 2`.** Without an upper or quiescence bound, that can hide reset/reconnect storms.
- **The Guestbook signing-state change.** Current `main` already exposes visible `Signing…` progress and `data-spinner`.
- **Generic CLI 10s/12s timeout increases, sandbox command timeouts that begin only after readiness, and manual capnweb redial/sleep.** They widen or add retry surfaces without fixing the causal failure.
- **The remaining diagnostic/phase formatting and PostHog UUID/comment changes.** The causal attribution gap was taken in #2361; the rest is measurement/reporting work rather than e2e runtime reliability.
- **Compatibility migrations or fallbacks for deleted Stream migrations.** The intended #2311 contract is clean data recreation. Reintroducing deleted schema history would conceal divergent production state.

### Best next candidates

There are no further changes in #2313 that I would honestly label *small, no-regrets reliability fixes*. The remaining original ideas may still be valuable, but each needs a separately designed operational PR with explicit invariants and production-shaped evidence:

1. **Script-settlement recovery across worker/DO rollout.** Isolate terminal settlement ownership and replay/idempotency invariants from the rest of the rollout machinery; prove no lost or duplicate settlement under a forced deployment.
2. **Sandbox admission/readiness boundaries.** Model package-install/cold-start readiness separately from command execution and give each phase a bounded, attributable failure. Do not start a larger generic timeout or hidden retry clock.
3. **Coherent SSR/worker-version cohorts.** If mixed-version rendering is still observed, design one version-cohort contract and verify it under an actual rollout rather than carrying the whole recovery stack over wholesale.

Two adjacent follow-ups are higher-confidence than those larger #2313 slices:

- **Production blocker discovered by #2363 (not a #2313 extraction): recreate the stale Streams Example App production stream namespace/ancestors and rerun the 6-case proof with a clean trace gate.** This is the immediate operational priority because the current errors are real, classified, and persistent.
- **Small test-only candidate not originally from #2313: split the oversized sandbox-egress proof into bounded cold-start/package-install and protocol assertions.** #2361 preview saw one 180-second first-attempt timeout followed by a roughly 42.5-second successful retry; separate clocks would localize the slow phase without raising the suite timeout.

That is the extraction boundary: eleven bounded fixes are merged; the rest should not be smuggled into reliability cleanup.

PR #2452, opened 2026-08-07
Let a hosted processor consume ephemeral events by naming them

jonastemplestein:
A hosted stream processor is woken only for durable events. That is why a processor that wants live input — microphone frames, turn edges — cannot have it, and why the voice lab needs a *second* stateful worker that opens its own session connection back to the same stream to catch them and relay. One call, two server-side owners, two recovery state machines.

The restriction was one ternary in the send loop: hosted connections filtered `event.ephemeral`, session connections did not. There is a real reason behind it — an ephemeral body lives only in the Stream DO's bounded buffer, so a restart or FIFO eviction leaves a permanent hole, and a durable cursor cannot be caught up across one — but it was applied to every hosted processor rather than to the ones that actually fold their input into reduced state.

## The rule

**An ephemeral event reaches a hosted processor when the processor's contract names that exact type in `consumes`.** The same list as everything else.

```ts
consumes: [
  "events.example.com/conversation-requested",  // durable
  "events.example.com/mic-frame",               // ephemeral — named, so delivered
],
```

**`"*"` never matches an ephemeral event.** That single carve-out is what makes one list safe, and it is what preserves today's behaviour: every processor in this repo consumes durable types or `"*"`, so without it they would all start receiving live traffic they never asked for the moment this shipped. A wildcard is written for durable facts. Naming a type explicitly is the decision to cope with an event whose body may be gone forever — a decision nobody makes by typing `"*"`.

No new field on the contract. None on the subscription. **`pnpm generate:itx-api` produces byte-identical output**, because there is no new public shape to describe. The whole change is the wildcard carve-out plus letting the hosted filter see ephemeral candidates.

## Proof

Two end-to-end tests against a real deployment, driving a processor written as userspace code, committed to a project's config repo, built as a dynamic worker, woken by a subscription:

```
✓ an ephemeral event reaches a hosted processor that named its type    9.3s
✓ a star-consuming hosted processor still receives no ephemeral events 6.5s
```

Unit tests could not answer this. They drive the connection API directly, which skips the wake — the stretch where the stream dials the worker, gets its announcement back, and only then decides what the connection may see. A capability can be green in every one of them and deliver nothing; an earlier revision of this PR was.

The processor proves receipt by appending a **durable** event, because an ephemeral one leaves no trace by design — exactly what made this unobservable.

## Risk map

**The riskiest line is the wildcard carve-out in `getConsumedEventDefinition`.** It is load-bearing rather than defensive: without it, `"*"` processors change behaviour on merge. Check it first.

**Nothing changes without a contract naming a type.** A processor that announces nothing keeps exactly today's delivery, pinned by a test. No stored state changes, no migration; `CORE_STATE_VERSION` untouched.

**Review order**

1. `packages/iterate/src/processors/processor-contracts.ts` — the carve-out.
2. `apps/os/src/domains/streams/stream-event-sender.ts` — the hosted filter.
3. `stream-event-sender.test.ts` — the rule four ways.
4. `stream-connections-and-subscriptions.e2e.test.ts` — the two end-to-end tests.
5. Unrelated tidy-ups riding along so `apps/os` does not drift behind the voice branch: an explicit serialization boundary for script results (`script-result-serialization.ts` + test + its use in `script-execution-entrypoint.ts`), and two connection tests using `ping()` as a round-trip barrier instead of awaiting a property.

Deliberately **not** included: everything under `apps/os/scripts/voicelab` (a ~20k-line voice lab, including its `config-repo/voice-agent.ts` template) and the `cli.ts` wiring that registers it. Those stay on `stream-for-audio`.

---

> [!NOTE]
> **Medium Risk**
> Changes core stream delivery and processor consume resolution; the `"*"` carve-out for ephemeral is load-bearing for every wildcard processor. Script serialization is a narrow RPC boundary fix with good test coverage.
> 
> **Overview**
> **Hosted processors can receive ephemeral events when their contract lists that exact type in `consumes`.** The send loop no longer strips ephemeral bodies before the hosted filter runs; admission is explicit per type, and **`"*"` never matches ephemeral** so existing wildcard consumers keep today’s durable-only delivery. Contract resolution mirrors that: ephemeral events skip the `"*"` fallback, and an ephemeral append of an ordinarily-durable type is rejected unless the catalogue marks the type ephemeral.
> 
> **Proof:** new unit coverage on the hosted filter rule plus two deployment e2e tests that wake a userspace echo processor and observe durable `saw` events (ephemeral leaves no log trace).
> 
> **Alongside (not ephemeral delivery):** script settlement now uses `serializeScriptResult` so nested RPC/`Promise.allSettled` rejections keep `message`/`name` instead of workerd’s `remote`/`durableObjectId` after `JSON.stringify`. Connection e2e barriers switch from awaiting `connectionKey` to `ping()`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d15b19e735b8a4944d05ca331f6664a7dea8e779. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

| --- | ---: | ---: |
| Product | +149 -14 | +32 -12 🟩⬜⬜⬜⬜ |
| Tests | +431 -2 | +238 -2 🟩🟩🟩🟩🟩 |
| **Total** | **+580 -16** | **+270 -14** 🟩🟩🟩🟩🟩 |

<sub>Lines counts every changed line; Significant ignores blank lines, JS comments, and TypeScript lines with no runtime output. Between 008c850 and d15b19e, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>

## Environment Config Lease

<details>

| Auth | released | `d15b19e` | [https://auth.iterate-preview-11.com](https://auth.iterate-preview-11.com) | 815.6 KiB | 16.6s | 10.7s |  | 4.5s | [Workflow run](https://github.com/iterate/iterate/actions/runs/184151447429941) | 2026-08-07T11:54:42.496Z | Preview app released. |
| Docs | released | `d15b19e` | [https://docs-preview-11.iterate-dev-preview.workers.dev](https://docs-preview-11.iterate-dev-preview.workers.dev) | 866.2 KiB | 11.3s | 1.9s |  | 3.8s | [Workflow run](https://github.com/iterate/iterate/actions/runs/184151447429941) | 2026-08-07T11:54:41.879Z | Preview app released. |
| Dummy Petshop | released | `d15b19e` | [https://dummy-petshop.iterate-preview-11.com](https://dummy-petshop.iterate-preview-11.com) | 198.3 KiB | 291ms | 5.3s |  | 1.2s | [Workflow run](https://github.com/iterate/iterate/actions/runs/184151447429941) | 2026-08-07T11:54:39.256Z | Preview app released. |
| OS | released | `d15b19e` | [https://os.iterate-preview-11.com](https://os.iterate-preview-11.com) | 5.09 MiB (-10.1 KiB vs main) | 87.6s | 213.7s |  | 13.2s | [Workflow run](https://github.com/iterate/iterate/actions/runs/184151447429941) | 2026-08-07T11:54:51.264Z | Preview app released. |
| Semaphore | released | `d15b19e` | [https://semaphore.iterate-preview-11.com](https://semaphore.iterate-preview-11.com) | 441.1 KiB | 13.2s | 38.2s |  | 1.2s | [Workflow run](https://github.com/iterate/iterate/actions/runs/184151447429941) | 2026-08-07T11:54:40.429Z | Preview app released. |
| Streams Example App | released | `d15b19e` | [https://streams.iterate-preview-11.com](https://streams.iterate-preview-11.com) | 1.51 MiB | 20.1s | 72.4s |  | 8.0s | [Workflow run](https://github.com/iterate/iterate/actions/runs/184151447429941) | 2026-08-07T11:54:46.083Z | Preview app released. |

</details>


jonastemplestein:
## Preview e2e is red, and it is not this diff

`Preview / deploy + e2e` fails on:

```
subscription "agent" does not exist
subscription "workspace" does not exist
```

**PR #2376 fails with the identical pair**, and that PR touches **0** files under `apps/os/src/domains/streams` or `packages/iterate/src/processors` — it is the ESP32 firmware consolidation. Two unrelated branches showing the same two missing subscriptions is an environment/preview-seeding failure, not either diff.

The deploy itself succeeded (`5 ok, 0 failed`); only the e2e assertions fail.

Everything else here is green: lint+typecheck, test (3,403), publish, LOC, autofix, continuous releases.

For completeness I checked the one way this diff could plausibly have caused it — the new optional `consumesEphemeral` being stripped by the `openedBy` parse, which would have made the second gate silently dead — and it survives `ConnectionOpenerDescriptor.parse` intact. Adding an optional field to a `strictObject` cannot reject a payload that previously parsed, and the 2,724 `apps/os` unit tests include the subscription-configuration reducers.

Not merging on red. Flagging it so whoever picks this up knows where the red came from.

jonastemplestein:
## Verified on the deployed preview

Same script, same payload, two environments — `preview-5` carries this PR, prd does not:

```
=== preview_5 (carries #2452) ===
ACCEPTED — committed at offset 5

=== prd (does not) ===
REJECTED — ZodError: unrecognized_keys ["includeEphemeral"] at payload.receiver
```

The payload was a real `subscription-configured` append with

```ts
receiver: {
  action: "processor-wake",
  expression: ["workers", "processor", "wakeStreamProcessor"],
  includeEphemeral: true,
}
```

and a filter matching a type that is never appended, so nothing is ever delivered through it.

This is worth stating precisely because the rejection is the interesting half: `SubscriptionConfiguredPayload` is strict, so a caller that sets this field against a deployment without the change does not degrade — its setup **throws**. Any consumer of the flag has to land after this deploys, which is exactly why the voice-lab consumer is not in this PR.

jonastemplestein:
### Narrowing the preview red a little further

Driving the failing path by hand against `preview-5` — the deployment this PR built — it **works**:

| | `agents.get(p).create({})` | `.message("hello")` |
|---|---|---|
| preview-5, existing project | ok | ok |
| preview-5, freshly created project | ok | ok |
| prd, for comparison | ok | ok |

So the agent facet subscription is configured and dialable on this deployment; the platform path in `agent-send-message` is not broken. Whatever the e2e is hitting is specific to the catalogue-example/REPL harness conditions, not to `agent.create` + `message`.

Combined with #2376 (firmware only, 0 stream files) failing on the identical `subscription "agent"/"workspace" does not exist` pair, I am confident this red is not from this diff — and I have not been able to reproduce it outside the harness. Leaving it for whoever owns the preview seeding rather than guessing at a fix.

jonastemplestein:
## Do not merge yet — I proved the delivery half and it does not work

I set this up end to end on `preview-5` (the deployment this PR built) to prove an ephemeral event actually reaches a hosted processor, not just that the schema accepts the flag. **It does not reach it**, and the reason matters for the design.

### What happens

Subscription committed correctly — `includeEphemeral: true`, ephemeral types in the filter. Appending an ephemeral `mic-frame` **does wake the processor** (four `connection-opened`/`connection-closed` pairs land right after the append). But the processor never sees the event, because the SDK runner drops it first:

```
StreamProcessor#isDeliverable
  -> #parseConsumedEvent
    -> getConsumedEventDefinition({ contract, eventType })   // looks in 
```

`consumesEphemeral` is not `consumes`, so the definition is `undefined`, the event is not deliverable, and it never reaches `processEvent`. **The platform change in this PR is necessary but not sufficient** — there is a second gate in `packages/iterate/src/processors/stream-processor.ts` that also has to know.

### And there is probably a better shape

The contract system already has a first-class notion of an ephemeral event, which I did not see before writing this:

```ts
// processor-contracts.ts — an event DEFINITION may be marked ephemeral
ephemeral: ephemeralEnvelopeSchema(args.ephemeral, StreamEventSchema.shape.ephemeral)
```

> for a definition marked `ephemeral: true`, absent defaults to `true` and an explicit `false` FAILS the parse — the contract, not the append site, decides that the event never becomes a durable stream fact.

So a contract can already say "this type is always ephemeral" in its `events` catalogue, and list it in `consumes` like anything else — at which point `getConsumedEventDefinition` finds it and the existing runner delivers it with no SDK change at all. That makes `consumesEphemeral` a second, parallel way to say something the contract system already expresses.

**The real trade-off, stated honestly, because I do not think it is my call:**

- **Catalogue (`events[t].ephemeral: true` + `consumes`)** — one concept instead of two, works with today's runner, no second gate. But an ephemeral type reaching `reduce` means a processor author *can* fold it into durable reduced state, which breaks replay determinism. Nothing structurally prevents it.
- **`consumesEphemeral` (this PR)** — structurally impossible to fold, because those types never reach `reduce`. But it needs a live delivery path in the SDK that does not exist yet, which is exactly why the proof above failed.

I would rather flag this than ship the wrong one. The schema half (`includeEphemeral` on the receiver) is needed either way; the second gate is the open question.

Marking this as **not ready to merge** until that is settled.

jonastemplestein:
### Narrowed one more level, then stopping

Ruled out the storage read. `StreamDurableObject`'s `readEvents` hook — the one the send loop uses — already passes `includeEphemeral: true`, and its own comment says it merges "SQLite rows with the current incarnation's memory-only ephemeral events **before subscription-specific visibility is applied**". Subscription-specific visibility is exactly the line this PR changes, so the events are present in the batch this diff filters.

Also ruled out the declaration route: I tried the catalogue shape too — `mic-frame` declared in the contract's `events` with `ephemeral: true` and listed in `consumes`, so `getConsumedEventDefinition` resolves it and the existing SDK runner would deliver it. **Same result: not delivered.** So this is not `consumesEphemeral` vs catalogue; something common to both is still dropping it.

What is left, in the order I would look:

1. **The opt-in wiring in this diff.** `includeEphemeral` is computed at `openHosted` and read from the `args` closure in the send loop. Worth confirming it is actually `true` on the connection that runs, rather than assumed — the reduced-state read of the receiver is the part I would distrust first.
2. **Wake/checkpoint timing.** The hosted connection opens at `replayAfterOffset: response.checkpointOffset`. If an earlier wake's scan already advanced the processor past the ephemeral offset without delivering it, a later wake starts after it and the body is gone — and ephemeral offsets are exactly the ones a scan is allowed to skip over.

The unit tests in this PR pass because they drive `openHosted` directly with `includeEphemeral: true` and a synthetic batch; they never exercise the wake path that computes it from reduced state. That gap is the likely home of the bug, and it is worth a test that goes through `wakeStreamProcessor` rather than around it.

Leaving it here rather than guessing further. The PR stays a draft: it is verifiably not doing what it says.

jonastemplestein:
## The likely answer, and it is a design problem rather than a bug in this diff

Three hypotheses are now dead, each with evidence:

- **Storage read** — `readEvents` already passes `includeEphemeral: true` and merges the buffer *before* subscription visibility.
- **Reducer** — pinned by a test in this PR: `includeEphemeral` survives into the stored configuration.
- **Announcement** — read off a live prd stream: the hosted `voice-agent` connection announces (`slug=voice-agent consumes=6`), so `ephemeralFilter` is defined and the opt-in computes `true`.

The plumbing is right. What is left is timing, and it is not fixable by moving a filter:

```
append(ephemeral mic-frame)        offset 33, body in the DO's in-memory buffer
  -> wake the processor            dials a USERSPACE dynamic worker: cold build, seconds
  -> worker returns a callback     + its checkpoint
  -> openHosted -> sendQueued      NOW reads the buffer
```

The ephemeral buffer belongs to "the current Durable Object incarnation". Seconds separate the append from the read, across a cold worker build, and any hibernation or restart in that window empties it. An ephemeral event cannot survive the latency of waking a cold processor — and a stream that is only woken *by* that event is cold by definition.

**Which means the no-bridge design needs a different shape than "wake the processor per frame".** For a hosted processor to consume live audio it has to be ALREADY WARM and holding an open connection for the duration of the call — which is precisely what `VoiceBridge` is, and precisely what the warm-up handshake in the voice lab already exists to guarantee.  is still necessary (a warm hosted connection needs permission to receive them), but it is not sufficient on its own, and the delivery it enables only works while a connection is already open.

I cannot close this out — preview-5 has been reclaimed (503; the ~3h shelf life), so I have no environment left to confirm the eviction directly. Stating it as the leading explanation rather than a proven one.

**Suggested next step for whoever picks this up:** prove or kill it with a test that appends an ephemeral event to a stream that ALREADY has a warm hosted connection open, versus one that has to be woken. If the warm case delivers and the cold case does not, the mechanism is confirmed and the design note above is the real deliverable of this PR.

jonastemplestein:
## Rebuilt off latest `main`, and the design changed

Force-pushed. This is now **one commit on top of `origin/main` (24fc4c4b2)**, nine files, and the earlier two-declaration design is gone entirely.

### The rule is now one sentence

An ephemeral event reaches a hosted processor when the processor's contract names that exact type in `consumes` — the same list as everything else. `"*"` never matches one.

Deleted versus the previous revision of this PR: `consumesEphemeral` on the contract and announcement, and `includeEphemeral` on the receiver, its envelope type and the connection plumbing. Two concepts for one question. Jonas called the old shape out and was right — the wildcard was the only real hazard, so carving `"*"` out is the whole safety argument and everything else was scaffolding around a problem I had invented.

It also very likely explains the delivery bug documented in the comments above: the SDK runner resolves consumed events through `getConsumedEventDefinition`, which searches `consumes` and nothing else. A type declared only in `consumesEphemeral` was invisible to it, so the platform could deliver to the connection and the processor would still never see it. Putting the type where the runner already looks removes that class of problem instead of adding a third place to keep in sync.

**`pnpm generate:itx-api` now produces byte-identical output** — there is no new public shape, so the generated surface is unchanged from main.

### Also riding along

So `apps/os` does not drift while the voice work continues on its own branch:

- `script-result-serialization.ts` (+ test) and its use in `script-execution-entrypoint.ts` — an explicit serialization boundary for script results.
- Two connection e2e tests use `ping()` as their round-trip barrier instead of awaiting a property.

Deliberately **not** included: everything under `apps/os/scripts/voicelab` (a ~20k-line voice lab, including its `config-repo/voice-agent.ts` template) and the `cli.ts` wiring that registers it. Those stay on `stream-for-audio`.

Gate, in a clean worktree off latest main: typecheck, lint 0/0, knip, format, **3,419 tests**.

### Still honest about what is unproven

The platform delivers to the connection and the SDK resolves the type, and both halves are tested — but this has not been watched carrying a real ephemeral event to a real userspace processor on a deployment. The only consumer is voice-lab code, deliberately out of this PR.

jonastemplestein:
## End-to-end tests added — and they do not pass yet

Two tests against a real deployment, in `stream-connections-and-subscriptions.e2e.test.ts`:

1. a hosted processor that **names** the ephemeral type in `consumes` receives it
2. a hosted processor consuming **`"*"`** still receives none

The second is the one guarding today's behaviour: every processor in this repo consumes durable types or `"*"`, so if the wildcard swept ephemeral events in, all of them would start receiving live traffic they never asked for the moment this shipped.

They build the real thing — a processor written as userspace code, committed to a project's config repo, built as a dynamic worker, woken by a subscription — because the unit tests structurally cannot answer this. They drive the connection API directly and skip the wake, which is the stretch where the flag is computed. A capability can be green in every one of them and deliver nothing, and this one was.

The processor proves receipt by appending a **durable** event, since an ephemeral one leaves no trace by design, which is exactly what made this unobservable.

### Status: unproven, deliberately visible

**They fail against local dev.** The harness has two barriers precisely so the failure is legible, and the second one fires: the wake never produces a hosted connection. The worker itself builds — the dev server logs a 702KB bundle for it, SDK included — so the dial happens and something inside the wake gives up.

I do not yet know whether that is my harness or local dev not supporting a wake that dials a userspace stateful Durable Object. Every other hosted-processor test in this file uses a **first-party** processor, which is a hint.

Committed unproven on purpose: e2e here runs against a deployed preview, where userspace dynamic workers demonstrably work (the entire voice lab depends on it), and CI is a better arbiter than more local guessing. If they fail there too, that is the signal — on a PR nobody is merging on red.

jonastemplestein:
## Fixed — both end-to-end tests pass, and the feature is proven

```
✓ an ephemeral event reaches a hosted processor that named its type   9.3s
✓ a star-consuming hosted processor still receives no ephemeral events 6.5s
```

This is the proof the PR has been missing since it was opened. A real userspace processor, committed to a project's config repo, built as a dynamic worker, woken by a subscription, receiving an ephemeral event and recording it in a durable fact.

### Both failures were mine, and neither reported itself where I was looking

**`contract` is a property, not a constructor argument.** `StreamProcessor` declares `abstract readonly contract`, so passing it through `super()` lands it in `deps` and leaves `this.contract` undefined — the host then throws on `contract.stateSchema` and the wake produces no connection at all.

**`readonly` is TypeScript; the worker source is JavaScript.** Fixing the first introduced the second, failing the esbuild step with `Expected ";" but found "contract"`.

The reason this cost so long is worth recording: **a build failure in a committed worker is not thrown at the caller.** It lands on the stream as `subscription-delivery-halted` carrying the esbuild message. So the test timed out on a barrier for ninety seconds while the answer sat one `getEvents` away, and I spent that time suspecting the platform. There is now a comment in the test saying exactly where to look.

One barrier was removed rather than fixed: waiting for a live hosted connection is the obvious way to prove the wake happened, and it **races** — the wake opens a connection, delivers, and closes it as idle within seconds, so polling `runtimeState` times out on a stream that worked perfectly. What the processor recorded is the durable proof, and the only evidence an ephemeral event can leave.

### One environment note

The dev server's RSS guard kills workerd mid-run once dynamic-worker isolates accumulate (`workerd RSS is 2993MB (limit 2048MB)`), which reads as an unrelated timeout. Restart the dev server before a run.

The PR is now: one commit for the feature off latest `main` (9 files, no public API change) plus these tests. Nothing about the platform needed fixing.

PR #2468, opened 2026-08-10
REPL runs through the capability-host script door

mmkal:
## What this becomes

The product REPL stops evaluating code in the browser tab and instead submits every Run through the **capability-host script door** — the same path agent scripts use (`itx.capabilityHost.runScript`): typechecked against the scope's capability types, journaled as `script-run-requested`/`script-run-settled` on a stream, and feeding the scope's derived `results` preamble.

- Each user gets a dedicated scope per project: `/repl/<user-id>`, born with the standard fallback to the project root host, so capability reads resolve project-wide.
- **History is durable**: the entry list renders from the scope stream's events — reload the page and your session is still there.
- **Continuity is `results`**: `results[0].data` (or `await results[0].load(itx)` for large outcomes) instead of the old mutable `$_`/`_`.
- The session REPL (browser-eval against the OS Session) is deleted; `/_app/itx-repl` becomes a minimal project chooser.

Spec: `tasks/repl-on-script-door.md` (first commit). Implementation follows on this branch.

Session id: `e89f2eb5-b303-47df-9e5d-5033942f37d8` (subagent)

| --- | ---: | ---: |
| Docs | +137 -0 | +121 -0 🟩🟩🟩🟩🟩 |
| **Total** | **+137 -0** | **+121 -0** 🟩🟩🟩🟩🟩 |

<sub>Lines counts every changed line; Significant ignores blank lines, JS comments, and TypeScript lines with no runtime output. Between f83499a and a57b5e9, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>

### DIFF
--- a/tasks/repl-on-script-door.md
+++ b/tasks/repl-on-script-door.md
@@ -1,0 +1,137 @@
+---
+status: in-progress
+size: large
+---
+
+# REPL on the script door
+
+## Status summary
+
+Spec fleshed out, implementation not started yet. Main pieces: script-door run
+path + per-user scope, stream-derived history, session REPL deletion, examples
+audit, editor types from the scope preamble, spec updates.
+
+Make the product REPL run scripts through the capability-host script door on a
+dedicated per-user scope, replacing in-browser evaluation. "Script door" =
+submitting code to a scope's capability host — the same path as agent scripts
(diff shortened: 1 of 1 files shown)


mmkal:
superseded by the durable-repl branch PR — renamed away from jargon

PR #2491, opened 2026-08-13
Lint: simple-truthiness-check + type-aware repo sweep

mmkal:
Adds `iterate/simple-truthiness-check` and applies it to ~4,600 sites. Follow-up to #2487.

**The rule** (commits 2 + 3, reviewable alone): flags `==`/`!=`/`===`/`!==` against `null`, `undefined`, or `NaN`, plus `.length` compared to 0. Suggestion-only, no autofix — the experiment below is why.

```ts
if (llm.durationMs == null) {}     // → !Number.isFinite(llm.durationMs)  (0 is meaningful)
if (llm.inputTokens == null) {}    // → !llm.inputTokens                  (plain truthiness)
const has = x !== undefined;       // → !!x   (bare `x` when already in an if/ternary test)
if (x !== null && x !== undefined) // → x     (dual check collapses to ONE report/fix)
if (items.length > 0) {}           // → items.length
if (items && items.length > 0) {}  // → items?.length  (only when the guard chain is call-free)
items.length === 0 ? a : b         // NOT flagged: explicit emptiness reads better as a ternary test
!x ? y : z                         // → x ? z : y   (flip — "if not x" makes the reader swap arms)
if (entry && entry.revokedAt) {}   // → entry?.revokedAt  (boolean contexts, side-effect-free chains)
```

Two suggestions per nullish report — truthiness first, `Number.isFinite` second — so `--fix-suggestions` bulk-applies the truthiness form and IDE users can pick per site.

**The experiment** (commits 4 + 5): rather than a blind sweep, a codemod fed oxlint's spans into each package's TypeScript program and chose the form by operand type: object/string → truthiness, number → `Number.isFinite`, boolean/bigint → skip for hand review. Verdict on the original question: **suggestion was the right call — autofix would have shipped real bugs.** What the sweep surfaced:

- **The codemod broke its own rule.** In the lint plugin, `node.value === null` (ESTree `Literal`) became `!node.value`, and the rule started flagging `x === 0` and `x === false`. AST tooling is a domain where null, 0, `false`, and `''` are four different real values — `.oxlintrc.json` now turns the rule off for `lint/**`.
- ~30 sites keep their precise comparison with a reasoned `oxlint-disable` line: script results (`null` is a value a script can return; `undefined` means no result — the itx-repl comment already said so), `kv.set` validation, OpenAPI params where `0`/`false` must serialize, `stableJson` hashing, evlog field dropping, live-state diffs (`undefined` reads as removed), JSON byte counting and type inference.
- **`Number.isFinite` doesn't narrow** — lib types it `(value: unknown) => boolean`, so the escape hatch un-narrowed ~60 nullable numbers. Fix: per-package `number-guards.d.ts` typing `isFinite`/`isInteger` as ts-reset-style type guards (deliberately "unsound" in the negative branch: NaN narrows away from `number`, which is exactly how this repo treats it). SDK files compiled by scaffolded config templates use native `typeof x === "number"` instead, since those programs don't see our ambient types.
- Mixed unions were the codemod's own bug class: `string | number` bucketed as number, so `Number.isFinite(part)` silently dropped strings (ci-telemetry artifact ids, sandbox `sleepAfter` duration strings like `"10m"`). A post-hoc type audit caught all 11 such sites.
- The preview e2e caught the best one: **validation doors**. `x !== undefined && (!Number.isSafeInteger(x) || x < 0)` throws on any DEFINED garbage; the sweep's `Number.isFinite(x) && ...` gate does the opposite — `isFinite(NaN)` is false, so NaN walked past the door, bound as SQL NULL downstream, and a stream DO burned CPU instead of throwing "must be a non-negative integer". All isFinite-gated throws were audited; stream offset/limit doors, optimistic-concurrency `expectedOffset` checks, and GitHub history depth restored with reasoned disables.
- The full test suite caught seven true regressions, all one shape — **domains where null/undefined/'' is a protocol**: agent summary updates (null CLEARS a field, omission preserves it — `applyAgentSummaryUpdate` ended up with a dead clear branch), the chat-reply title and secret `refresh` folds (same null-clears wire semantics), mobile repo-working-tree (`null` = absent vs `""` = real empty-file content — empty files became deletes), the template-reference parser (`github:o/r#main&path:` started parsing instead of rejecting), and two test fixtures whose tri-state params (undefined = valid default, null = omit the field) quietly turned "missing header → 401" into 200. Protocol files got file-level disables naming the rule; single lines got line disables.

## Risk map

- **Riskiest:** the ~3,200 mechanical truthiness rewrites conflate `''`/`0`-adjacent semantics wherever a type said "object/string" but the domain meant "sentinel". The repo-working-tree case above is the shape of the residual risk; typecheck + full test suite are the net, and both are green. Skim any file you know has null-vs-undefined tri-states.
- `number-guards.d.ts` changes what `Number.isFinite` means to the type checker repo-wide (23 copies, one per tsconfig — consolidating them is an open question worth a review comment).
- The `x?.length` guard collapse changes evaluation from two reads to one — restricted to call/side-effect-free chains.
- **Review order:** `lint/oxlint-plugin-iterate.ts` (rule) → `lint/oxlint-plugin-simple-truthiness.test.ts` → `number-guards.d.ts` + `.oxlintrc.json` → the ~30 disable-comment sites (grep `oxlint-disable-next-line iterate/simple-truthiness-check`) → hand-fix commits → mechanical sweeps last (they're 90% of the diff).
- On merge: no runtime behavior change intended anywhere; the disable comments and task file record where behavior deliberately stayed put.

Session: 4a4b91ba-35e9-42e2-a3d0-c268dd2e1ba4

| --- | ---: | ---: |
| Product | +3,854 -3,894 | +3,550 -3,861 🟩🟩🟥🟥🟥 |
| Mobile | +240 -225 | +220 -224 🟥⬜⬜⬜⬜ |
| UI components | +1,297 -1,365 | +1,267 -1,351 🟩🟥⬜⬜⬜ |
| Tests | +774 -394 | +725 -394 🟩⬜⬜⬜⬜ |
| CI & scripts | +249 -238 | +234 -238 🟥⬜⬜⬜⬜ |
| Docs | +148 -0 | +135 -0 🟩⬜⬜⬜⬜ |
| Other | +507 -78 | +397 -75 🟩⬜⬜⬜⬜ |
| Generated | +65 -49 | +57 -41 🟩⬜⬜⬜⬜ |
| **Total** | **+7,135 -6,243** | **+6,586 -6,184** 🟩🟩🟩🟥🟥 |

<sub>Lines counts every changed line; Significant ignores blank lines, JS comments, and TypeScript lines with no runtime output. Between b90ecb9 and fcd0ff5, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>

## 📱 Mobile preview

Channel `simple-truthiness-check` · runtime `023cbbc4d` · **JS-only** — the installed app can run it

<details open><summary>OTA — switch the installed app to this PR's channel (`fcd0ff5`)</summary>

**[Switch this phone to the PR channel](https://os.iterate.com/m/preview-channel/simple-truthiness-check)**

<a href="https://os.iterate.com/m/preview-channel/simple-truthiness-check">QR code</a>

</details>
<details><summary>Full install — if the app is uninstalled or the runtime differs (`c3b64f8`)</summary>

**[Open the EAS build install page](https://expo.dev/accounts/mishanustom/projects/iterate/builds/3fe243bc-7133-4020-9205-e6b67ab02ece)**

<a href="https://expo.dev/accounts/mishanustom/projects/iterate/builds/3fe243bc-7133-4020-9205-e6b67ab02ece">QR code</a>

<sub>Installing gets you a compatible binary on the default channel — then use the OTA link above to switch it to this PR's JS.</sub>

</details>

<sub>Back to main inside the app: Build info → Reset to default channel. Republishes on every push to this PR.</sub>

## Environment Config Lease

<details>

| Auth | released | `fcd0ff5` | [https://auth.iterate-preview-11.com](https://auth.iterate-preview-11.com) | 817.4 KiB | 29.4s | 19.4s |  | 3.6s | [Workflow run](https://github.com/iterate/iterate/actions/runs/419776653217293) | 2026-08-14T11:42:02.848Z | Preview app released. |
| Docs | released | `fcd0ff5` | [https://docs-preview-11.iterate-dev-preview.workers.dev](https://docs-preview-11.iterate-dev-preview.workers.dev) | 866.1 KiB | 24.9s | 2.9s |  | 2.8s | [Workflow run](https://github.com/iterate/iterate/actions/runs/419776653217293) | 2026-08-14T11:42:02.045Z | Preview app released. |
| Dummy Petshop | released | `fcd0ff5` | [https://dummy-petshop.iterate-preview-11.com](https://dummy-petshop.iterate-preview-11.com) | 198.3 KiB | 7.1s | 31.2s |  | 1.0s | [Workflow run](https://github.com/iterate/iterate/actions/runs/419776653217293) | 2026-08-14T11:42:00.258Z | Preview app released. |
| OS | released | `fcd0ff5` | [https://os.iterate-preview-11.com](https://os.iterate-preview-11.com) | 5.11 MiB (-11.7 KiB vs main) | 65.8s | 232.7s |  | 10.2s | [Workflow run](https://github.com/iterate/iterate/actions/runs/419776653217293) | 2026-08-14T11:42:09.404Z | Preview app released. |

</details>


apps/docs/src/lib/board-model.ts
```diff
@@ -71,7 +71,7 @@ export function taskPathInFolder(path: string, folder: string): string {
 
 function taskSummary(source: string): { text: string; from: number } {
   const frontmatter = /^---[ \t]*\r?\n[\s\S]*?\r?\n(?:---|\.\.\.)[ \t]*(?:\r?\n|$)/.exec(source);
-  const bodyOffset = frontmatter === null ? 0 : frontmatter[0].length;
+  const bodyOffset = !frontmatter ? 0 : frontmatter[0].length;
```
mmkal: this should also be banned. 
```suggestion
  const bodyOffset = frontmatter ? frontmatter[0].length : 0;
```
is obviously better. or probably better still:

```suggestion
  const bodyOffset = frontmatter?.[0].length ?? 0;
```

would also do the trick. i think we should ban confusing "if not x then y else z" expressions

apps/mobile/src/lib/approver.ts
```diff
@@ -111,7 +111,7 @@ export async function approverKeyStatus(
   if (!key) return { kind: "unenrolled" };
   const enrolled = await projectApprovalKeys(baseUrl, projectId);
   const entry = enrolled.find((candidate) => candidate.keyId === key.keyId);
-  if (entry && entry.revokedAt !== null) return { kind: "revoked", key };
+  if (entry && entry.revokedAt) return { kind: "revoked", key };
```
mmkal: this should also be `if (entry?.revokedAt)` no?

apps/docs/src/lib/board-model.ts
mmkal: 🤖 Done in fcd0ff538. The rule now flags negated ternary conditions (`!x ? y : z` → flip to `x ? z : y`; `!!x ? y : z` sheds its bangs) — scoped to ternaries rather than enabling eslint's `no-negated-condition`, which also flags `if/else` and `!==` tests. The sweep flipped ~600 of them, and the flips cascaded: the codemod's `...(!x ? {} : {x})` spreads went positive and then collapsed to `...(x && {x})` via prefer-logical-and-spread. This line took your 'better still' form verbatim: `frontmatter?.[0].length ?? 0`. Seven sites with comments lodged between the arms were flipped by hand (the rule reports without a suggestion there instead of dropping the comment).

apps/mobile/src/lib/approver.ts
mmkal: 🤖 Yes — done in fcd0ff538, this line is now `if (entry?.revokedAt)`. Generalized into the rule: `x && x.y` in a boolean context suggests `x?.y` (the nearest `&&` operand is the guard, so `a && b && b.c` becomes `a && b?.c`; call chains never collapse since that would change evaluation counts; value positions are left alone because the collapse changes the expression's non-boolean value). oxlint lists typescript/prefer-optional-chain but it currently loads as zero rules, hence the in-house check.

PR #2492, opened 2026-08-13
Bump middlewright to 0.1.5, enable all lint plugin rules for specs

mmkal:
Bumps `middlewright` from a pkg.pr.new pin (3 commits behind) to the released **0.1.5**, and turns on the full `middlewright/lint-plugin` rule set for `specs/**` — with every pre-existing violation actually fixed, no exclusions. Review feedback pushed this beyond lint config into real product work.

## What changes

- **Dependency**: `https://pkg.pr.new/middlewright@71c6865...` → `0.1.5`. The pin was missing [Prefer positive waits over detached waits (#28)](https://github.com/iterate/middlewright/pull/28) — the source of the new `prefer-positive-waits` rule. `patches/middlewright.patch` (spinner-waiter disappearance-goal fix) is **not** upstreamed in 0.1.5; it stays and re-applies cleanly.
- **Lint**: `specs/**` now gets all three plugin rules, with zero carve-outs:

```jsonc
"middlewright/prefer-locator-waits": "error",
"middlewright/prefer-positive-waits": "error",
"middlewright/require-timeout-comment": "error"
```

  All ~91 pre-existing `require-timeout-comment` violations were fixed: 6 repo-ide tab-click timeouts **removed** (those specs re-run green without them), the rest annotated with the real reason each timeout exists — `expect.poll` budgets outside the spinner-waiter's reach, `spinnerWaiter.settings.run({disabled})` blocks, raw popup pages with no middleware, and >30s cold-build/LLM budgets past the 30s spinner ceiling.
- **Rule dedup**: `iterate/spec-restricted-syntax` (bans all awaited `expect()` in specs) now skips `toBeVisible`/`toContainText` — `middlewright/prefer-locator-waits` owns those with an autofix, so one mistake reports once.
- **Product — decided approvals get positive UI** (mobile notifications screen): a batch decided from the screen used to vanish instantly on all-reject with zero acknowledgment. Decided batches now linger for the session with their outcome ("Rejected"/"Approved"), the post-decide button reads "Recording decision…" while the authoritative record rides the stream back (honest in-flight copy the spinner-waiter can also see), rejects no longer read "Signing…", and outcome badges carry `approval-decision-badge` testIDs.
- **Specs wait on positive UI instead of detachment**: `decideBatch` waits for the activity card's ✓/✗ approval glyph; `rejectFromExpansion` waits for the row-scoped outcome badge. The mobile signup preamble also tolerates the authorize re-entry skipping the `/project-access` Continue step (pre-existing race, surfaced while verifying locally).

## Risk map

- Riskiest: the notifications list derivation change (`deriveNotificationListRows` grew a `sessionDecidedOffsets` param) — unit-tested in `notifications.test.ts`, and both mobile specs pass locally end to end against the bumped middlewright.
- On merge: dev-dependency + lint config + one deliberate mobile UX change (decided rows linger for the session). New spec files fail lint for unexplained `timeout:` options — that's the point.
- Review order: `apps/mobile/src/lib/notifications.ts` + screen wiring → `approval-batch.tsx` labels → spec helper rewrites (`specs/mobile/*`) → `.oxlintrc.json` + `lint/oxlint-plugin-iterate.ts` → the mechanical timeout-comment annotations across `specs/**` → `package.json`/lockfile.

Session ID: `99f7d462-f777-4a97-8825-90660a32001a`

---

> [!NOTE]
> **Medium Risk**
> The notifications list derivation and approval decision wiring are user-visible mobile behavior with race-sensitive stream timing; the rest is test tooling and lint, though the broad spec timeout edits could mask or fix real flakes.
> 
> **Overview**
> **middlewright** moves from a pkg.pr.new pin to **0.1.5** (existing spinner-waiter patch still applies). All three **middlewright** lint rules are enabled for `specs/**` with no per-file carve-outs; ~91 `require-timeout-comment` issues are fixed (real inline reasons, or timeouts removed where specs proved unnecessary). `iterate/spec-restricted-syntax` no longer flags `toBeVisible`/`toContainText` so those overlap with **prefer-locator-waits** once.
> 
> On **mobile notifications**, batches decided on that screen no longer disappear under the deciding finger: session-scoped `sessionDecidedOffsets` keeps synthetic rows showing **Rejected** / **Approved** (human decisions only; expiry still drops orphans). `ApprovalBatchActions` adds `onDecideStarted` before append so live stream echoes cannot unmount the row before mutation callbacks run; reject in-flight copy is **Deciding…** / **Recording decision…** instead of misleading **Signing…**. Outcome badges get `approval-decision-badge` testIDs.
> 
> **Playwright specs** lean on spinner-waiter where loading UI exists (agent-chat, create-project drop manual disables), wait on positive UI (approval glyphs / badges instead of detached buttons), and annotate long timeouts where middleware cannot help (polls, popups, cold builds).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 736a927b7b1b692258f3a675b0bb9a12262cc32e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Environment Config Lease

<details>

| Auth | released | `73fca8f` | [https://auth.iterate-preview-1.com](https://auth.iterate-preview-1.com) | 817.7 KiB | 19.5s | 8.0s |  | 4.8s | [Workflow run](https://github.com/iterate/iterate/actions/runs/363396796906088) | 2026-08-13T14:56:31.779Z | Preview app released. |
| Docs | released | `73fca8f` | [https://docs-preview-1.iterate-dev-preview.workers.dev](https://docs-preview-1.iterate-dev-preview.workers.dev) | 866.2 KiB | 36.2s | 4.4s |  | 3.8s | [Workflow run](https://github.com/iterate/iterate/actions/runs/363396796906088) | 2026-08-13T14:56:30.773Z | Preview app released. |
| Dummy Petshop | released | `73fca8f` | [https://dummy-petshop.iterate-preview-1.com](https://dummy-petshop.iterate-preview-1.com) | 198.3 KiB | 58ms | 7.6s |  | 1.2s | [Workflow run](https://github.com/iterate/iterate/actions/runs/363396796906088) | 2026-08-13T14:56:29.092Z | Preview app released. |
| OS | released | `73fca8f` | [https://os.iterate-preview-1.com](https://os.iterate-preview-1.com) | 5.11 MiB (-11.4 KiB vs main) | 83.7s | 142.2s |  | 16.5s | [Workflow run](https://github.com/iterate/iterate/actions/runs/363396796906088) | 2026-08-13T14:56:43.476Z | Preview app released. |
| Semaphore | released | `32bbbe4` | [https://semaphore.iterate-preview-1.com](https://semaphore.iterate-preview-1.com) | 441.2 KiB | 25.7s | 16.4s |  | 872ms | [Workflow run](https://github.com/iterate/iterate/actions/runs/363396796906088) | 2026-08-13T14:56:27.885Z | Preview app released. |
| Streams Example App | released | `32bbbe4` | [https://streams.iterate-preview-1.com](https://streams.iterate-preview-1.com) | 1.53 MiB | 34.7s | 64.7s |  | 8.7s | [Workflow run](https://github.com/iterate/iterate/actions/runs/363396796906088) | 2026-08-13T14:56:35.755Z | Preview app released. |

</details>

| --- | ---: | ---: |
| Mobile | +177 -26 | +111 -19 🟩🟩🟩⬜⬜ |
| Tests | +190 -172 | +110 -112 🟩🟩🟥🟥🟥 |
| Config | +1 -1 | +1 -1 🟩⬜⬜⬜⬜ |
| Docs | +84 -0 | +74 -0 🟩🟩⬜⬜⬜ |
| Other | +25 -15 | +19 -15 🟩⬜⬜⬜⬜ |
| Generated | +5 -6 | +5 -6 🟥⬜⬜⬜⬜ |
| **Total** | **+482 -220** | **+320 -153** 🟩🟩🟩🟥🟥 |

<sub>Lines counts every changed line; Significant ignores blank lines, JS comments, and TypeScript lines with no runtime output. Between b90ecb9 and 736a927, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>

## 📱 Mobile preview

Channel `middlewright-0.1.5` · runtime `023cbbc4d` · **JS-only** — the installed app can run it

<details open><summary>OTA — switch the installed app to this PR's channel (`736a927`)</summary>

**[Switch this phone to the PR channel](https://os.iterate.com/m/preview-channel/middlewright-0.1.5)**

<a href="https://os.iterate.com/m/preview-channel/middlewright-0.1.5">QR code</a>

</details>
<details><summary>Full install — if the app is uninstalled or the runtime differs (`c3b64f8`)</summary>

**[Open the EAS build install page](https://expo.dev/accounts/mishanustom/projects/iterate/builds/3fe243bc-7133-4020-9205-e6b67ab02ece)**

<a href="https://expo.dev/accounts/mishanustom/projects/iterate/builds/3fe243bc-7133-4020-9205-e6b67ab02ece">QR code</a>

<sub>Installing gets you a compatible binary on the default channel — then use the OTA link above to switch it to this PR's JS.</sub>

</details>

<sub>Back to main inside the app: Build info → Reset to default channel. Republishes on every push to this PR.</sub>


specs/mobile/approvals.spec.ts
```diff
@@ -390,6 +390,7 @@ async function decideBatch(
     page.once("dialog", handler);
     try {
       await button.click().catch(() => {});
+      // detached decide button is the only landed-decision signal — the decided batch vanishes with no replacement UI
```
mmkal: is this true? we don't see a checkmark? why don't we add a checkmark?

specs/mobile/notifications.spec.ts
```diff
@@ -343,6 +343,7 @@ test("the approval push is suppressed in the watched thread and sent when you're
     await page.getByText("Awaiting decision").waitFor();
     await page.getByText("egress-echo?orphan=1").waitFor();
     await rejectFromExpansion(page, "orphans get decided here too");
+    // detached is the payoff: all-reject removes the synthetic row and no positive UI replaces it
```
mmkal: so let's add the ui?

specs/mobile/notifications.spec.ts
```diff
@@ -403,6 +404,7 @@ async function rejectFromExpansion(page: Page, reason: string) {
     page.once("dialog", handler);
     try {
       await button.click().catch(() => {});
+      // detached Reject button is the only landed-decision signal — the decided row vanishes with no replacement UI
```
mmkal: same as above

specs/repo-ide-json-schema.spec.ts
```diff
@@ -38,6 +38,6 @@ test("flags a package.json schema violation with a red squiggle", async ({
   // spinner-waiter clamps a locator.waitFor here to its 1ms no-spinner
   // fail-fast; the web-first assertion is not middleware-instrumented and
   // keeps the generous 20s budget the one-off fetch needs.
-  // oxlint-disable-next-line iterate/spec-restricted-syntax -- no spinner exists during the background schemastore fetch, so locator.waitFor gets fail-fasted to 1ms; expect().toBeVisible() polls the full 20s.
+  // oxlint-disable-next-line iterate/spec-restricted-syntax, middlewright/prefer-locator-waits -- no spinner exists during the background schemastore fetch, so locator.waitFor gets fail-fasted to 1ms; expect().toBeVisible() polls the full 20s.
```
mmkal: what's in `iterate/spec-restricted-syntax` again? is it overlapping with prefer-locator-waits? should we remove/reduce down if so?

.oxlintrc.json
```diff
@@ -279,20 +282,36 @@
...
+        "specs/repo-ide-markdown-preview.spec.ts",
+        "specs/repo-ide-svg-index-preview.spec.ts",
+        "specs/seeded-apps.spec.ts",
+        "specs/signup.spec.ts",
+        "specs/stream-cache-before-live.spec.ts",
+        "specs/stream-resume-after-suspend.spec.ts",
+        "specs/test-support/email-otp-signup.ts",
+        "specs/test-support/repo-tree.ts",
+        "specs/workspace-lens-board.spec.ts"
+      ],
```
mmkal: wtf why don't you just fucking do `specs/**/*.ts` or something

```suggestion
      "files": ["specs/**/*.ts"],
```

specs/mobile/approvals.spec.ts
mmkal: 🤖 You were right to push — the claim wasn't quite true. The activity card's header already renders ✓/✗ approval glyphs (aria-labels "approved"/"rejected"), visible whether the card is collapsed or expanded. In b8beacc03 decideBatch waits for that glyph instead of button detachment, and the post-decide button now reads "Recording decision…" — honest in-flight copy (the decision landed, the authoritative record hasn't ridden the stream back yet) that also keeps the spinner-waiter engaged through the gap. Approvals spec passes locally against this.

specs/mobile/notifications.spec.ts
mmkal: 🤖 Added in b8beacc03: batches decided from this screen now linger in the list with their outcome ("Rejected"/"Approved") instead of vanishing under the deciding finger. Session-scoped — a decided-here set fed by onDecided — so the documented no-device-independent-history rule still holds on the next visit. The spec now asserts the lingering row's "Rejected because: …" record instead of waiting for detachment, and notifications.test.ts covers the new derivation.

specs/mobile/notifications.spec.ts
mmkal: 🤖 Same product fix as the thread above: the decided row lingers with its outcome, so rejectFromExpansion in b8beacc03 waits for the row-scoped outcome badge (testID approval-decision-badge) instead of button detachment — scoped to the row so an earlier rejection's badge elsewhere can't vouch for this one. Notifications spec passes locally.

specs/repo-ide-json-schema.spec.ts
mmkal: 🤖 spec-restricted-syntax bans every awaited expect() in specs (plus toBe(true/false), waitForURL, and baseURL-in-goto) — so middlewright/prefer-locator-waits (toBeVisible/toContainText only, but with an autofix) was a strict subset and every such mistake reported twice. Reduced in b8beacc03: the iterate rule now skips those two matchers and defers to the middlewright rule, so one mistake → one report, autofixable; the broader awaited-expect ban still catches everything else (toHaveAttribute etc.). The three disable comments shrank accordingly.

.oxlintrc.json
mmkal: 🤖 The exclusion override is gone entirely — no file list, no glob, and the ratchet task file is deleted. All ~91 violations are fixed in b8beacc03: the 6 repo-ide tab-click timeouts were removed outright (re-ran those specs to prove they weren't load-bearing), and the rest now carry the actual reason each timeout exists — expect.poll budgets outside the spinner-waiter's reach, spinner-waiter-disabled blocks, raw popup pages with no middleware, and >30s cold-build/LLM budgets past the spinner ceiling.

specs/agent-chat.spec.ts
```diff
@@ -53,11 +53,11 @@ test("agent replies to a browser chat message in the feed", async ({ helpers, pa
     const composer = page.getByPlaceholder("Message this agent");
     // This is route readiness, not an LLM wait; bound it explicitly before
     // interacting with the deliberately tight default action budget.
-    await composer.waitFor({ timeout: 30_000 });
+    await composer.waitFor({ timeout: 30_000 }); // timeout: manual — spinner-waiter is disabled for this block
```
mmkal: i'm not actually sure it should be disabled for this block though. First, I do not love that we have a spec that hits the *real* LLM, but let's grant that for now. Why do we completely disable the spinner-waiter?  `"Thinking…" state renders two spinner-matching elements at once which trips spinner-waiter's strict-mode` sounds like a middlewright bug to me

mmkal: 🤖 Your instinct was right — it WAS a middlewright bug, and it's already fixed upstream: anySpinnerVisible has been multi-element-safe since iterate/middlewright#3 (filter({visible:true}).count() instead of a bare isVisible()), which landed 2026-07-09 — a week AFTER this disable was written (2026-07-02). The 0.1.5 bump makes the sit-out stale, so 7673ce5e0 removes it here and in create-project.spec.ts (same stale reason), drops the manual 15-30s timeouts (the "Thinking…"/pending states are honest spinners the middleware now follows), and keeps only create-project's 60s cold-slot budgets, which exceed the spinner-waiter's 30s ceiling. Both specs pass locally with the middleware live. The remaining settings.run({disabled}) blocks cite different, still-true reasons: unmarked skeletons (signup, seeded-apps, email-otp-signup) and asserting on spinner-shaped text itself (stream-cache-before-live). On hitting the real LLM: agreed it's not lovely — but it's this spec's entire subject (message→reply routing), so leaving that for another day.

specs/clients-os-app.spec.ts
```diff
@@ -99,7 +99,7 @@ test("two dashboard tabs are two clients; an itx caller navigates each independe
     // @ts-expect-error - dynamic capability member
     await host.capabilities.browser.navigate(`/projects/${slug}/settings`);
     await expect
-      .poll(() => page.url(), { timeout: 15_000 })
+      .poll(() => page.url(), { timeout: 15_000 }) // timeout: poll budget — expect.poll is outside the spinner-waiter's reach
```
mmkal, line 102: why are we doing expect.poll on just getting the url?

specs/repo-ide-svg-index-preview.spec.ts
```diff
@@ -37,13 +37,13 @@ test("toggle an svg file between Code and its sandboxed Preview", async ({
...
-  await page.getByRole("tab", { name: "Preview" }).click({ timeout: 10_000 });
+  await page.getByRole("tab", { name: "Preview" }).click();
   const preview = page.locator('iframe[title="HTML preview"]');
   // The tab changes URL-owned view state through a React transition while the
   // usable Code pane remains on screen. There is intentionally no loading
   // spinner, so assert the bounded eventual render rather than treating the
   // transition as a missing product loading state.
-  // oxlint-disable-next-line iterate/spec-restricted-syntax -- React retains the usable Code pane during this search-state transition, so there is intentionally no spinner for locator.waitFor to follow.
+  // oxlint-disable-next-line middlewright/prefer-locator-waits -- React retains the usable Code pane during this search-state transition, so there is intentionally no spinner for locator.waitFor to follow.
   await expect(preview).toBeVisible();
```
mmkal: i still don't see why we can't use `await preview.waitFor()` they are functionally almost identical

specs/seeded-apps.spec.ts
```diff
@@ -27,16 +27,16 @@ test("the seeded guestbook app works after creating a project", async ({
   // module. Seeing the signed note proves that the SDK's LiveState target and
   // the app's RPC root share one Cap'n Web class identity end to end.
   await page.goto(appUrl("guestbook", fixture.project.slug, baseURL!));
-  await page.getByRole("heading", { name: "Guestbook" }).waitFor({ timeout: 120_000 });
+  await page.getByRole("heading", { name: "Guestbook" }).waitFor({ timeout: 120_000 }); // timeout: cold worker-bundler build — far past the spinner-waiter's 30s ceiling
```
mmkal: we can use `await spinnerWaiter.settings.run({ spinnerTimeout: ... }, async () => {...})` to allow for even longer spinners. That's assuming we do have a proper spinner UI which we absolutely should

specs/repo-ide-svg-index-preview.spec.ts
mmkal: 🤖 You're right — tried it and it just works: `await preview.waitFor()` passes repeatedly (the transition lands well inside the spinner-waiter's 1s readiness quick-check, so the no-spinner fast-fail never triggers). The expect().toBeVisible() workaround and its disable comment are gone in 6bdf492da. The jsonc/json-schema siblings keep theirs: the background schemastore fetch takes multiple spinnerless seconds, which genuinely outlasts the 1s readiness check.

specs/seeded-apps.spec.ts
mmkal: 🤖 Adopted in 6bdf492da: the building page IS proper spinner UI (the status widget sets data-spinner="true" and the page titles itself "Building…"), so all three cold-build waits in this spec now use spinnerWaiter.settings.run({ spinnerTimeout: 130_000 }) with plain waitFor()s instead of hand-rolled 120s timeouts. Guestbook + todo tests pass locally (the docs one needs DOCS_APP_ORIGIN, preview-only). Didn't apply it to the todo-heading wait after "Continue with iterate" — that stretch spans app→OS→app auth-callback navigations where no spinner exists yet, so the spinner can't carry it edge to edge.

specs/clients-os-app.spec.ts
mmkal: 🤖 Pre-existing code (this PR only added the trailing comment), but the why: the navigation is driven REMOTELY — host.capabilities.browser.navigate() is an RPC whose resolution doesn't wait for the target tab to finish navigating, so page.url() flips asynchronously afterwards. The URL is the capability's actual contract here (and its sibling assertion is "tab two's URL did NOT change"), page.waitForURL is banned by iterate/spec-restricted-syntax, and a destination-locator wait would test the settings page rather than the navigation capability — so polling the URL is the honest observable. The spinner-waiter isn't in play either way: expect.poll isn't a locator action. Happy to switch it to a destination-locator wait + one-shot URL assert if you'd rather test through UI.

specs/mobile/approvals.spec.ts
```diff
@@ -390,7 +399,7 @@ async function decideBatch(
     page.once("dialog", handler);
     try {
       await button.click().catch(() => {});
-      await button.waitFor({ state: "detached" });
+      await page.getByLabel(outcome, { exact: true }).waitFor();
```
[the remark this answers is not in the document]
mmkal: 🤖 Fair catch — in the current spec the two lanes wait on distinct outcomes ("approved" then "rejected") so a leftover glyph couldn't misfire, but the helper is generic and a future same-outcome lane would be vacuously satisfied. Fixed in 32bbbe49a: decideBatch pins the batch's activity card before the press (via the "approval pending" ◷ it wears — the decide dialog is the card's sibling, so the button itself can't anchor the lookup) and waits for that same card's glyph to flip. Spec passes locally.

apps/mobile/src/app/project/[projectId]/notifications.tsx
```diff
@@ -398,12 +416,17 @@ function ApprovalNotificationDetail({
           baseUrl={baseUrl}
           offset={batchOffset}
           onDecided={() => {
+            // Keeps this batch's row visible (with its outcome) for the rest
+            // of the session — the row must not vanish under the finger that
+            // just decided it.
+            onSessionDecided(batchOffset);
```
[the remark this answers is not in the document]
Comment (mmkal) [reply apps/mobile/src/app/project/[projectId]/notifications.tsx]: 🤖 Real race, confirmed — worse than a flicker: if the websocket echo unmounted the row first, the mutation's observer callbacks died with it and the setState never ran at all. Fixed in the pushed commit: ApprovalBatchActions now fires a required onDecideStarted at the start of mutationFn — before the append — so the offset is pinned before any echo can close the batch. Only actually-closed batches act on the recorded offset, so a cancelled reject prompt recording is inert. Notifications spec passes locally.

apps/mobile/src/components/approval-batch.tsx
```diff
@@ -273,6 +279,11 @@ export function ApprovalBatchActions({
   const enrolledKey = key.data?.kind === "enrolled" ? key.data.key : null;
   const respond = useMutation({
     mutationFn: async (decision: "approve" | "reject"): Promise<"decided" | "cancelled"> => {
+      // Before the append: once the decided event is on the stream, the live
+      // subscription's echo can re-render the surface ahead of any of this
+      // mutation's observer callbacks (which die silently if that re-render
+      // unmounts us).
+      onDecideStarted();
```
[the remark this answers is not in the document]
mmkal: 🤖 Tightened in the pushed commit: onDecideStarted now fires after promptForRejectReason commits — a cancelled reject records nothing — and deriveNotificationListRows additionally requires decidedBy === "human" to linger, so an abandoned attempt's batch closing by EXPIRY never wears a decided row (unit-tested). The one remaining soft edge is the approve lane: the Face ID prompt lives inside the signing step, so a cancelled sign still records — with the expiry guard that only surfaces if another of this human's devices decides the batch, where showing the outcome seems right; documented on the prop.

PR #2474, opened 2026-08-11
Userland-first agent defaults: birth batches as data, prompts out of platform code

mmkal:
Stacked on #2423 (base: `codemode-tag-format`).

Two moves, one principle: **the event vocabulary is the platform; everything opinion-shaped is data a project owns.**

## 1. Birth defaults are now a list of plain events

`project/agent-birth-defaults-configured` was a hand-picked field bag (config patch + system prompt + subscription list, each with its own builder plumbing). It's now:

```ts
await itx.streams.get("/").append({
  type: "events.iterate.com/project/agent-birth-defaults-configured",
  idempotencyKey: `my-config/agent-birth-defaults:${contentHash}`,
  payload: {
    birthEvents: [
      { type: "events.iterate.com/agent/configured", payload: { config: { driver: "agent-headless" } } },
      { type: "events.iterate.com/agents/context-added", payload: { role: "system", key: "agent/system-prompt", content: prompt } },
      { type: "events.iterate.com/stream/subscription-configured", payload: { name: "agent-headless", receiver: { action: "facet-processor", source: { kind: "builtin" } } } },
    ],
  },
});
```

A prompt is just a keyed context event, a driver choice just `agent/configured` — the next defaultable thing needs zero platform changes. Guard rails, since this is userland data reaching a platform-assembled batch:

- **Fold-time validation** in the project processor (`state.agentBirthDefaults`): every event must parse against the agent-consumed vocabulary, except `stream/subscription-configured` with a `facet-processor` receiver whose name is on a tight allowlist (`agent-headless`). A malformed latest payload folds to `null` — degrade to platform-default births, never to stale defaults, never break creation.
- **Platform-minted idempotency keys** (`agent/birth-defaults:<hash>:<index>:<projectId>:<agentPath>`) — projects can't wedge creates with hand-rolled keys.
- The creation door reads the processor **snapshot** (the raw-event scan is deleted); `matches.pathPrefix` scopes which agents pick the defaults up.
- Precedence in one sentence: explicit call-site policy > project birth events > embedded-template fallback. A project prompt-slot event replaces the platform's fallback prompt slot; a router's explicit `systemPromptPolicy` drops project prompt-slot events entirely.

## 2. Prompts move out of platform code

`DEFAULT_AGENT_SYSTEM_PROMPT` (~17KB string constant) and the slack/telegram/email prompt builders now source from **`configs/default/prompts/*.md`** — the config-repo template itself:

[...]

PR #2497, opened 2026-08-13
Generic project/defaults-configured: the project processor no longer knows what an agent is

mmkal:
POC for de-agenting the project processor, per review feedback on #2474: **a project shouldn't know what an agent is.**

## Before / after

The project contract imported `AgentBirthDefaults` from the agents domain, declared `project/agent-birth-defaults-configured`, and validated birth events against the agent vocabulary at fold time. Now the project stream is a generic latest-wins-per-key fact store:

```ts
// any domain, userland or platform, publishes:
await itx.streams.get("/").append({
  type: "events.iterate.com/project/defaults-configured",
  idempotencyKey: `my-config/defaults:agents/birth-defaults:${contentHash}`,
  payload: { key: "agents/birth-defaults", value: { birthEvents: [...] } },
});
```

- Project fold: `state.defaults[key] = value`, stored raw. No schema, no agent imports, no validation.
- The agents domain owns the key constant (`AGENT_BIRTH_DEFAULTS_KEY`), the `AgentBirthDefaults` schema, and validation — all moved to the creation door's read site, with the same degrade-to-platform-defaults posture. "Never stale" survives: the raw latest always replaces the previous raw value, so a malformed latest degrades to *no* defaults, not to its predecessor.
- The next domain wanting a project-scoped default is a new key — zero project-contract changes.
- No backcompat (per review): the old `agent-birth-defaults-configured` event type is gone entirely. Events of that type already on streams become unrecognized.
- Both templates publish the new shape under **new idempotency-key prefixes** (the stream rejects same-key-different-body appends, so the old prefixes must not be reused).

## Risk map

- **Riskiest part: validation moved from fold to read.** The fold used to guarantee the creation door a pre-validated snapshot; now the door parses raw `unknown` on every create (`agentBirthDefaultsForProject` in rpc-targets.ts). The degrade path (schema fail / vocabulary fail / allowlist fail → warn + platform defaults) rests on that one function — review it first. The subscription allowlist check is a security gate and now runs only there.
- **On merge:** existing projects' config repos still publish the OLD event type, which is now unrecognized — their birth defaults stop applying until the config repo publishes the generic event. Default-template projects degrade to the byte-identical embedded fallback prompt (invisible); codemode-tag experiment projects fall back to the worker's corrective sweep path until updated. No data migration.
- **Not covered by a unit test:** the door-level raw-parse seam (`agentBirthDefaultsForProject` is module-private in the 5k-line rpc-targets.ts). Covered indirectly by builder-level degrade tests + validate tests; a preview-slot live check (fresh project → agent born with template prompt) is the remaining verification listed in the task file.
- **Review order:** `rpc-targets.ts` (the door), `project-processor-implementation.ts` + `project-processor-contract.ts` (fold/contract), `configs/*/worker.ts` (publishers), tests, generated files last (mechanical).

## Background: why the defaults are read at the door (the race this design kills)

The old reactive flow — the config worker can only react to `agent/created`, which is definitionally after birth; the in-DO classic processor dials the model before the reaction's delivery hop completes:

the reactive race: config worker's append arrives after the classic processor already dialed

The current flow — the project publishes a standing fact once at deploy; `create()` reads it inside its own call stack, before the one atomic birth append, so there is no window:

the constitutive flow: standing fact read inside create() before the atomic birth append

This PR doesn't change that mechanism — it changes *whose vocabulary the stored fact is spelled in* (the project stores it opaquely; the agents domain interprets it at the door).

Task file: [tasks/generic-project-defaults.md](https://github.com/iterate/iterate/blob/generic-project-defaults/tasks/generic-project-defaults.md) — includes the assumptions made (naming, shim, `matches` staying inside the value).

Verification: typecheck, lint, knip, format green; full suite 269 files / 2787 tests passed.

<details>
<summary>Aside: "couldn't we just make the userspace worker faster instead?"</summary>

Honest answer first: for today's default web chat, a reliably-~25ms reaction **would** win the ~550ms race (message arrival + 250ms debounce) essentially always, and the observed jank would disappear. Three reasons "mostly always" can't become "always":

1. **The deadline is a userland knob.** The budget being raced is `llmRequestDebounceMs`, which legitimately goes down — waiter-chef ships 100ms, and an agent born *with* its first message in the batch has only the debounce as slack. At debounce→0, any hop > 0 loses. Speed re-litigates the race every time someone makes agents snappier; it never retires it.

the deadline is a knob: 550ms today, 250ms born-with-message, 0 with debounce 0

2. **A fast hop is a p50; the race is decided at p99.9.** Cold starts, redeploys — and the failure mode that isn't latency at all: during a project's first build the worker doesn't exist and the project-worker lane *drops* deliveries (skip-on-failure, no retry). t=∞, correlated exactly with when new chats happen (fresh project, first deploy). A lost race isn't retried: the turn runs to settlement in the wrong format, user-visible.

latency is a distribution: green mass under the deadline, red tail beyond it, and a dropped-forever bar at infinity

3. **Brief wrong state is still representable wrong state.** Even winning at 25ms, the stream spends 25ms saying the wrong thing, and anything reading in that window (another subscriber, a script, a 0-debounce processor) acts on it. Guaranteeing the reaction beats every reader means delivering it on the commit path — holding the birth until userland answers — i.e. an interception/hold-gate mechanism: kernel machinery with liveness obligations.

fast reaction leaves a brief representable wrong state; birth defaults make wrong state have no offset to live at

The slowness isn't a bug in the worker; it's the defining property of an *observation* lane (async, batched, skip-on-failure, can't wedge a stream). A reaction lane fast and reliable enough to win this race is a different, heavier subsystem — and if that ever gets built, birth defaults folds into it as the declarative case. Until then, this is the workaround that costs one event instead of one subsystem.

</details>

Session id: `5a7db041-4282-4390-908c-6fe75128f1e8`


configs/codemode-tag/worker.ts
```diff
@@ -204,30 +206,37 @@ export default class ProjectWorker extends IterateWorkerEntrypoint {
...
-            },
-          ],
+              {
+                type: "events.iterate.com/stream/subscription-configured",
+                payload: {
+                  name: HEADLESS_PROCESSOR_SLUG,
+                  receiver: { action: "facet-processor", source: { kind: "builtin" } },
+                },
+              },
+            ],
```
mmkal, line 242: might be nice/reassuring to add a `satisfies` here like
```suggestion
            ] satisfies Array<WhicheverAgentEventTypeUnion>,
```

mmkal: or could just do the `satisfies` on the whole value so it's like

```ts
  payload: {
    key: "...",
    value: {
      birthEvents: [...],
    } satisfies AgentBirthDefaults,
```

mmkal: 🤖 Done in cd166c9d5, went with the whole-value form. One naming/scope note: the template can't reach the `AgentBirthDefaults` zod schema (it lives in apps/os, and its `birthEvents` is deliberately loose — `{type: string, payload?: record}` — so satisfying it would prove nothing). Instead `iterate/sdk` now exports `AgentBirthDefaultsValue`: `birthEvents` is `AgentEventInput` (the generated agent-consumed union, same one `agent.append()` takes) **plus** the one allowlisted exception (builtin facet-processor subscription), which `AgentEventInput` alone doesn't include. Both templates use it — codemode-tag via `satisfies` on the value, default via an annotation on the conditional `birthEvents` (+ `satisfies` at the append). Typo'd event names or bad payloads are now compile errors in config workers instead of runtime degrades at the door.

PR #2507, opened 2026-08-18
Agent birth moves to the config worker; headless-only; one-shot

mmkal:
Was stacked on #2506 (worker delivery obligations) — that merged; this PR now targets main and carries the whole birth refactor.

Agent personality stops being a platform decision. `create()` commits only the
mechanism; the project's config worker authors every birth and interprets every
response — and the platform's defaults become services the worker calls.

### What a template looks like now

```ts
// configs/default/worker.ts — the whole "headed agent" story is two jobs:
case "events.iterate.com/agent/created": {
  const agent = this.itx.agents.get(event.path);
  await agent.append(
    ...(await agent.getDefaultBirthEvents({ kind: "web" })), // prompt, model, boot context
    { // your project's opinions, as plain keyed events
      type: "events.iterate.com/agents/context-added",
      idempotencyKey: "iterate/config/house-style:v1",
      payload: { role: "system", key: "config/house-style",
                 content: "all responses should be in all-lowercase" },
    },
    { type: "events.iterate.com/agent/birth-finalized", idempotencyKey: "…" },
  );
  break;
}
case "events.iterate.com/agents/context-added": // …and the other interpretable events
  await this.itx.agents.get(event.path).interpretResponse(event);
```

### The pieces

- **Readiness hold**: the keeper holds every LLM trigger until
  `agent/birth-finalized` reduces — early messages accumulate, nothing
  dispatches on a half-authored personality. A ~10s platform deadline arms at
  the **first held message** (an idle unborn agent waits forever for free); on
  expiry a visible `agent/birth-timed-out` fact lands and its delivery appends
  the platform-default personality + finalize — all content-hash keyed, so a
  late worker's own birth events supersede in place instead of conflicting.
- **One keeper, one slug**: the classic interpreting `AgentProcessor` and the
  `config.driver` knob are deleted; the keeper registers under the vocabulary's
  own contract and slug, `agent` (the transitional `agent-headless` split
  contract is dissolved — "headless" survives only in a history comment).
  Contract v6.0.0.
- **`interpretResponse`**: the classic codemode component becomes a per-event
  service — response parsing, slash commands, settlement rendering (with
  workspace spill), preamble + stream-error transcription, waiting clears.
  Same `agent/` idempotency keys, so redelivered handlers converge. The
  platform never calls it on its own; vendoring your own interpreter (see
  `configs/codemode-tag`) is the pinning story.
- **Integration agents migrate**: Slack/Telegram/email routers record channel
  facts in the birth certificate (`{ channel: { type, connection, … } }`);
  `getDefaultBirthEvents({ kind: "slack" | … })` interpolates the channel
  prompts from them. MCP sessions become kind `"mcp"`; onboarding waits like
  everyone.
- **Cleanup, same change**: the generic project defaults store
  (`project/defaults-configured` + fold slot + SDK type, from #2497), the
  birth-defaults validation/allowlist, the default template's prompt-pinning
  publish, and codemode-tag's whole handover/driver-flip machinery.
- **Boundary specs** (`agent-birth-degraded-start.test.ts`): (a) keyed
  supersession heals a late-personality-after-degraded-start in place; (b) a
  wrong-personality turn already in model-visible history cannot be un-said —
  the standing, executable argument for a future logical/physical stream-path
  split, deliberately not built.

### Risk map

Riskiest first:

1. `agent-turn-loop.ts` — the hold + degraded-start state machine. The
   correctness argument: the deadline sleeper appends only the timed-out FACT;
   the fold applies it only while unfinalized, and the personality+finalize is
   a per-event consequence of the applied fact — so a worker finalize that
   commits first wins outright (single-stream total order), with no
   check-then-append window on the personality. Pinned by
   `agent-headless-processor.test.ts` and the boundary specs.
2. `agent-response-interpreter.ts` — the executable-output gate moved from
   "matches the open request" to "the fold accepted it into contextItems".
   Same verdict, read later; a raw append claiming an offset still reduces to
   nothing and never executes.
3. `rpc-targets.ts` (`create()` shrink, the two new services) and the router
   birth-certificate changes.
4. Behavior on merge: classic-era streams (subscribed `agent`) resolve against
   the current keeper again — their first NEW message rides the 10s degraded
   start to a fresh finalize, and old driver-carrying `agent/configured`
   events fail the strict patch parse and reduce to nothing. The residual gap
   is the short-lived `agent-headless`-subscribed streams (codemode-tag
   experiments, this PR's own pre-rename preview streams): that name now
   points at no registered processor — preview-only, erased with the slot.
   Old `agents/birth-defaults` publishes become inert data.
5. Every turn now costs a worker delivery hop before interpretation — the
   speed round is explicitly deferred.
6. Mechanical tail: test surgery (the unit harness now plays the worker's
   interpretation loop), generated files (`itx-api*`,
   `config-repo-template.generated.ts`).

Live preview-slot verification (fresh project first turn; broken-worker commit
→ visible degraded start) has not run yet — CI previews on this PR are the
place to do it.

Task file: `tasks/agent-birth-userland-refactor.md` (checklist + implementation
log of every mechanism/opinion judgment call).

Session id: `64e48e34-690e-42d0-b403-0646b3d626b0`

## Environment Config Lease

> [!CAUTION]
> Refusing to report preview e2e success: the semaphore leases no slot to pr-2507 and no runnable deployment is recorded. E2e was NOT run. Run preview deploy first.

<details>

| Auth | released | `577fc87` | [https://auth.iterate-preview-5.com](https://auth.iterate-preview-5.com) | 817.3 KiB | 35.1s | 18.2s |  | 4.2s | [Workflow run](https://github.com/iterate/iterate/actions/runs/212638169919775) | 2026-08-19T12:39:07.268Z | Preview app released. |
| Docs | released | `577fc87` | [https://docs-preview-5.iterate-dev-preview.workers.dev](https://docs-preview-5.iterate-dev-preview.workers.dev) | 866.2 KiB | 64.8s | 2.4s |  | 3.2s | [Workflow run](https://github.com/iterate/iterate/actions/runs/212638169919775) | 2026-08-19T12:39:06.293Z | Preview app released. |
| Dummy Petshop | released | `577fc87` | [https://dummy-petshop.iterate-preview-5.com](https://dummy-petshop.iterate-preview-5.com) | 198.3 KiB | 41ms | 6.1s |  | 921ms | [Workflow run](https://github.com/iterate/iterate/actions/runs/212638169919775) | 2026-08-19T12:39:03.979Z | Preview app released. |
| OS | released | `577fc87` | [https://os.iterate-preview-5.com](https://os.iterate-preview-5.com) | 5.11 MiB (-9.3 KiB vs main) | 66.9s | 220.6s | 2 retried: agents.liveState subscribed before any agent exists serves the empty catalog, then pushes the first agent and its title (vitest x1) — expected { birthCertificate: null, …(2) } to deeply equal { birthCertificate: {}, …(2) } · lazy commits, reads, and clone-lane writes share one history (vitest x1) — stream-wait-timeout: Timed out waiting for stream event after 98594ms (the public deadline expired while recovery re-armed one-shot waits). | 18.3s | [Workflow run](https://github.com/iterate/iterate/actions/runs/212638169919775) | 2026-08-19T12:39:21.341Z | Preview app released. |
| Semaphore | released | `577fc87` | [https://semaphore.iterate-preview-5.com](https://semaphore.iterate-preview-5.com) | 441.3 KiB | 32.6s | 74.5s |  | 1.2s | [Workflow run](https://github.com/iterate/iterate/actions/runs/212638169919775) | 2026-08-19T12:39:05.150Z | Preview app released. |
| Streams Example App | released | `577fc87` | [https://streams.iterate-preview-5.com](https://streams.iterate-preview-5.com) | 1.54 MiB | 43.8s | 96.7s |  | 8.6s | [Workflow run](https://github.com/iterate/iterate/actions/runs/212638169919775) | 2026-08-19T12:39:11.706Z | Preview app released. |

</details>

| --- | ---: | ---: |
| Product | +1,123 -1,215 | +622 -703 🟩🟩🟥🟥🟥 |
| UI components | +5 -1 | +2 -1 🟩⬜⬜⬜⬜ |
| Tests | +700 -590 | +539 -519 🟩🟩🟥🟥⬜ |
| CI & scripts | +41 -3 | +27 -3 🟩⬜⬜⬜⬜ |
| Docs | +341 -41 | +327 -41 🟩⬜⬜⬜⬜ |
| Other | +250 -266 | +143 -140 🟩⬜⬜⬜⬜ |
| Generated | +172 -73 | +88 -67 🟩⬜⬜⬜⬜ |
| **Total** | **+2,632 -2,189** | **+1,748 -1,474** 🟩🟩🟩🟥🟥 |

<sub>Lines counts every changed line; Significant ignores blank lines, JS comments, and TypeScript lines with no runtime output. Between 3aa53c9 and 577fc87, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>


apps/os/src/domains/agents/agent-defaults.ts
```diff
@@ -582,26 +502,16 @@ export function agentCreationForPath<
 
   return {
```
mmkal, line 497: Let's not use the word "keeper"

apps/os/src/domains/agents/agent-defaults.ts
```diff
@@ -221,249 +204,236 @@ export const MCP_AGENT_SYSTEM_PROMPT = [
...
+function agentBootContextContent(coordinates: {
+  agentPath: string;
+  projectId: string;
+  project?: { name: string; slug: string; workerUrl?: string };
+}): string {
+  const { agentPath, projectId, project } = coordinates;
+  return [
+    "Context for this agent:",
+    project === undefined
+      ? `- Project id: ${projectId}`
```
mmkal, line 376: Avoid the word "door" 

Speak normally ffs

apps/os/src/domains/agents/agent-defaults.ts
mmkal: 🤖 Swept in 592dcc14f — every "keeper" and branch-introduced "door" replaced with plain language ("the agent processor", "two ways to write", etc.) across code comments, test names/harnesses, the contract description, template READMEs, the task file, and one model-facing prompt line. Pre-existing "door" idiom in files this branch didn't touch was left alone to keep the diff scoped.

apps/os/src/domains/agents/agent-defaults.ts
mmkal: 🤖 Swept in 592dcc14f — every "keeper" and branch-introduced "door" replaced with plain language ("the agent processor", "two ways to write", etc.) across code comments, test names/harnesses, the contract description, template READMEs, the task file, and one model-facing prompt line. Pre-existing "door" idiom in files this branch didn't touch was left alone to keep the diff scoped.

apps/os/src/domains/agents/agent-defaults.ts
```diff
@@ -213,257 +188,244 @@ export const EMAIL_AGENT_SYSTEM_PROMPT = interpolatePromptTemplate(
...
-    )
-    .max(20),
-});
-export type AgentBirthDefaults = z.infer<typeof AgentBirthDefaults>;
-
-/** Platform-lane items a project may include beyond the agent-consumed
- * vocabulary: attaching a registered agent-family processor. Everything else
- * platform-lane stays platform-authored. */
-const BIRTH_DEFAULTS_SUBSCRIPTION_ALLOWLIST = new Set<string>(["agent-headless"]);
+  coordinates?: {
```
mmkal: explain why this is optional/can be undefined

apps/os/src/domains/agents/agent-defaults.ts
```diff
@@ -213,257 +188,244 @@ export const EMAIL_AGENT_SYSTEM_PROMPT = interpolatePromptTemplate(
...
+    agentPath: string;
+    projectId: string;
+    project?: { name: string; slug: string; workerUrl?: string };
```
mmkal, line 267: explain why this is optional/can be undefined

apps/os/src/domains/agents/agent-defaults.ts
```diff
@@ -213,257 +188,244 @@ export const EMAIL_AGENT_SYSTEM_PROMPT = interpolatePromptTemplate(
...
+      content: systemPrompt,
+    },
+  });
+  const modelEvent = AgentProcessorContract.buildEvent({
+    type: "events.iterate.com/agent/configured",
+    idempotencyKey: `agent/default-birth:model:${contentHash(model)}`,
+    payload: { config: { llm: { model } } },
+  });
+  const coordinates = input.coordinates;
+  if (coordinates === undefined) return [promptEvent, modelEvent];
```
mmkal:
```suggestion
  if (!coordinates) return [promptEvent, modelEvent];
```

apps/os/src/domains/agents/agent-defaults.ts
```diff
@@ -213,257 +188,244 @@ export const EMAIL_AGENT_SYSTEM_PROMPT = interpolatePromptTemplate(
...
+ * mis-addressed personality. */
+function defaultSystemPromptForKind(input: {
+  kind: AgentBirthKind;
+  coordinates?: { agentPath: string };
+  birthCertificate?: Record<string, unknown>;
+}): string {
+  switch (input.kind) {
+    case "web":
+    case "onboarding":
+      return DEFAULT_AGENT_SYSTEM_PROMPT;
```
mmkal: can we move this to a `default:` block
we could have a `input.kind satisfies "web" | "onboarding"` in the block to make sure we're exhaustively handling the types at compile time in future but that way it'll work for future `kind`s fairly well at runtime

apps/os/src/domains/agents/agent-defaults.ts
```diff
@@ -213,257 +188,244 @@ export const EMAIL_AGENT_SYSTEM_PROMPT = interpolatePromptTemplate(
...
+): Extract<AgentChannelFacts, { type: Kind }> {
+  const parsed = AgentChannelFacts.safeParse(birthCertificate?.channel);
+  if (!parsed.success || parsed.data.type !== kind) {
+    throw new Error(
+      `${kind} default birth events need channel facts in the birth certificate: ` +
+        `agent/created payload { channel: { type: ${JSON.stringify(kind)}, ... } }` +
+        (parsed.success ? ` (found ${JSON.stringify(parsed.data.type)})` : ""),
+    );
+  }
+  return parsed.data as Extract<AgentChannelFacts, { type: Kind }>;
```
mmkal: can we do something like


```suggestion
  const SchemaWithKind = AgentChannelFacts.and(z.object({ type: z.literal(kind) }))
  const parsed = SchemWithKind.safeParse(birthCertificate?.channel);
  if (!parsed.success) {
    const message = `agent/created of kind ${kind} had invalid channel facts: ${z.prettifyError(parsed.error)}`;
    throw new Error(message);
  }
  return parsed.data;
```

apps/os/src/domains/agents/agent-processor-contract.ts
```diff
@@ -33,10 +33,13 @@ import { AgentBinding, AgentSummary, AgentSummaryUpdated } from "./agent-presenc
...
-  version: "5.2.0",
+  version: "6.0.0",
   description:
-    "Maintains model-visible history, schedules debounced offset-identified LLM turns, runs " +
-    "them through the Workers AI transport, and executes scripts through the capability host.",
+    "The agent processor: maintains model-visible history and schedules debounced " +
+    "offset-identified LLM turns, held until the project's config worker finalizes the " +
+    "agent's birth (agent/birth-finalized, with a degraded-start deadline), run through the " +
+    "Workers AI transport. Interprets nothing: response interpretation is a service the " +
+    "project's config worker invokes per event (itx.agents.get(path).interpretResponse).",
```
mmkal:
```suggestion
    "project's config worker invokes per event (usually with itx.agents.get(path).interpretResponse).",
```

apps/os/src/domains/agents/agent-processor-implementation.ts
mmkal, line 1: how come this isn't called `agent-processor-implementation.ts` like before?

apps/os/src/domains/agents/agent-processor.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  protected override reduce({ event, state }: ReduceArgs<AgentProcessorContract>) {
+    return reduceAgentEvent({ event, state });
+  }
+}
+
+/** Adapts a hosting StreamProcessor to the component host surface. Usable by
+ * any agent-contract processor class — call it with `this` from a field
+ * initializer. The base class's members are protected, so the adapter
+ * reaches them through a scoped cast rather than widening them to public.
```
mmkal: why isn't this just a method on the class then?

apps/os/src/domains/agents/agent-turn-loop.ts
```diff
@@ -75,30 +78,12 @@ export class AgentTurnLoop implements AgentComponent {
...
-        // action, not conversation: it must neither cancel an in-flight turn
-        // nor schedule one — the SAME pure resolver gates it here, in the
-        // codemode component, and in the fold's contextTriggerSource, so the
-        // three can never disagree.
+        // interpretation service (it runs as a script when project code asks),
+        // and it is a side-band action, not conversation: it must neither
+        // cancel an in-flight turn nor schedule one — the SAME pure resolver
+        // gates it here, in the interpreter, and in the fold's
+        // contextTriggerSource, so the three can never disagree.
         if (payload.role === "user" && resolveSlashCommand(payload.content) !== null) break;
```
mmkal, line 87: does this mean `/somethingwithatypo foo bar` will cause no response? (probably fine if so, i'm just asking)

apps/os/src/domains/agents/agent-defaults.ts
mmkal: 🤖 Made it required instead, in bf0f372d3 — better than explaining. The turn loop knows the agent's identity (the base StreamProcessor exposes path/projectId; the component host now surfaces them), so the degraded start passes it too and even gains a boot-context item — id-only project line, which a late worker's directory-informed version supersedes by content hash. Optionality deleted rather than documented.

apps/os/src/domains/agents/agent-defaults.ts
mmkal: 🤖 This one stays optional, with the why now in a doc comment (bf0f372d3): best-effort directory facts — absent when the directory has no record yet, or when the caller has no directory access at all (the degraded start inside the stream's Durable Object). Absence degrades the boot context to an id-only project line; it must never block a birth.

apps/os/src/domains/agents/agent-defaults.ts
mmkal: 🤖 Superseded in bf0f372d3: coordinates became required (thread above), so this early-return branch is gone entirely.

apps/os/src/domains/agents/agent-defaults.ts
mmkal: 🤖 Done as suggested in bf0f372d3: web/onboarding moved to the `default:` block with `input.kind satisfies "web" | "onboarding"` inside — compile-time exhaustiveness for future kinds, graceful runtime fallback to the web prompt if one slips through anyway.

apps/os/src/domains/agents/agent-defaults.ts
mmkal: 🤖 Done nearly verbatim in bf0f372d3: `AgentChannelFacts.and(z.object({ type: z.literal(kind) }))` + `z.prettifyError`, with one breadcrumb kept in the message about where the facts live (`agent/created payload { channel: ... }`). Bonus: the intersection type made the `as Extract<...>` cast unnecessary.

apps/os/src/domains/agents/agent-processor-contract.ts
mmkal: 🤖 Applied verbatim in bf0f372d3.

apps/os/src/domains/agents/agent-processor-implementation.ts
mmkal: 🤖 It was the slug-rename fold's choice (short name when agent-headless-processor.ts merged into the vacated spot) — but the domain convention is clearly `*-processor-implementation.ts` (project, agent-collection, slack, telegram, email all follow it), so renamed to `agent-processor-implementation.ts` in bf0f372d3, git mv so history follows.

apps/os/src/domains/agents/agent-processor.ts
mmkal: 🤖 It was a vestige of the two-class era — the free function existed to be shared between the classic and headless processor classes, hence the scoped cast to reach protected members from outside. With one class left it's now a private `#componentHost()` method (bf0f372d3), and the cast disappeared since the class reaches its own protected members directly.

apps/os/src/domains/agents/agent-turn-loop.ts
mmkal: 🤖 No — a typo'd `/somethingwithatypo foo bar` gets a normal LLM reply. `resolveSlashCommand` returns non-null only for recognized, runnable commands: `/example <slug>` must match a known, run-script-capable catalogue entry (unknown slugs and malformed vars return null), `/script` needs non-empty code. Everything else "falls through to the LLM like a typo" (slash-commands.ts's own words). The turn-loop gate only suppresses turns for messages that will actually run as scripts.

mmkal:
Human edit: keeping open just for findability/side-by-side, will close later

PR #2525, opened 2026-08-26
Agents reuse journaled scripts: itx.capabilityHost.previousScriptHelper

mmkal:
Agents re-derive the same codemode script every time a request repeats with different inputs. Every past script is already journaled on the scope stream — this PR lets the agent point at a past run, name its inputs by value, and run it again:

```ts
async (itx) => {
  const helper = await itx.capabilityHost.previousScriptHelper({
    ...results[0], // any retained result row — scriptOffset addresses the journaled script
    parameterize: { n: 52479543428582704627n }, // the value the old script used inline, swapped OUT
  })
  return await helper.run({ n: 66778601389380731119n })
}
```

## Demo

Two turns through the real chat UI (deterministic `intercepted/*` model from #2523; both scripts execute for real — the second answer is computed by the reused child run). The spec opens each turn's codemode snippet: turn 1 is the full derived algorithm, turn 2 is four lines pointing at `results[0].scriptOffset`. Recorded by `VIDEO_MODE=1 pnpm spec agent-script-reuse`.

https://github.com/user-attachments/assets/49fa24d1-2bd8-4d4c-ba3d-a155f4b06e89

And the typed-return spec — turn 2's first scripted attempt fails the gate (`satisfies readonly number[]`), the corrective retry passes and computes the answer via the reused child run:

https://github.com/user-attachments/assets/e03a06ba-fbe9-40fc-9d4a-8ac6453d0fcf

**Latest strict local eval (2026-08-27): mechanism passed, outcome failed.** A real agent reused the journaled ~2,400-character Pollard-rho script on its first turn-2 attempt through `previousScriptHelper`, and the platform ran the original algorithm as the expected UUID child execution. But the helper call supplied `parameterize` without `edits`, so it emitted the new factors beside the old number before sending a correction. The eval requires every chat message to be correct, making the visible stale message a failure. The tracked report helper now aggregates token usage across every agent stream.

How it works:

- `previousScriptHelper` lives on the capability host (beside `runScript`/`getScriptResult`/`setPreamble` — the script journal's own surface) and resolves the offset to the journaled `script-run-requested` event (idempotency-key point read, bounded forward scan for agent-lane runs).
- Parameters are **values, not text**: the platform renders each value's candidate literal spellings (strings across quote styles when escape-free; bare number/bigint/boolean literals boundary-matched so `42` can't hit `142`), requires exactly one occurrence in total, and swaps it for a generated alias (`__reuse_<name>`, deterministically suffixed until free) — parameter names never collide with the script's own identifiers.
- **`run(vars)` is typed from `parameterize`**, and the result type is best-effort inferred from the spread row's `data`/`load` (overloads; the SHAPE of the old result, not its values). Passing a bigint as a string — the live models' favorite mistake — dies at the typecheck gate before anything runs (e2e asserts this). Spreading the row copies only enumerable props, and large rows' loader plumbing is now non-enumerable, so rows are structured-clone-safe.
- The handle is a plain RpcTarget end to end — no harness magic; works in every runtime (codemode, CLI, REPL, workers).
- **Void scripts now leave a `done` row in `results`** (contract 0.6.0 → 0.7.0): a script ending with plain `return;` previously left no row, so the agent had no offset handle to reuse it.
- One system-prompt bullet teaches the helper; the guarded prompt ceiling moves 4250 → 4350, argued in the constant's comment log as that file prescribes.

The eval drove the design — each round bought a fix:
- Round 1: a rejected attempt became `results[0]` and the model pointed the next attempt at its own failed reuse script, nesting until it gave up → **failed runs are rejected as reuse sources**, with the fix named in the error.
- Round 2: splice-thinking — a whole `const n = 123n;` line as the parameter text (child died: `n is not defined`) → now **unrepresentable**: parameters are primitive values, not text.
- Round 3: passed, but only after two name-collision retries (models pick the script's own variable name first every time) → **generated aliases removed the collision class**.
- Round 4: **passed with zero failed attempts.**
- Rounds 6–8 (post-review `edits`): round 6 failed under the stricter no-stale-messages criteria (reuse worked, prose didn't) → the stale-leftover guard; round 7 failed on edit-before-parameterize clobbering the value literal → edits now run after parameterization; round 8 **passed clean** — `parameterize` + `edits` in a 251-char reuse script, no incorrect message ever emitted.
- Round 5 (typed value-form parameters): reuse succeeded **first-attempt** (373-char script + journaled child), but the run failed the strict criterion — the model re-derived the algorithm (2,097 chars) while correcting the reused message's stale prose, instead of writing the short correction from its delivery receipt as rounds 3–4 did. Model stochasticity in the correction turn, not a mechanism gap; follow-up idea logged in the task file. The eval now has tracked run/report helpers (`evals/script-reuse/run-turn.ts`, `report.ts`).

A second spec asserts the **type chain against the real gate** with a failing-then-passing pair: turn 1 returns its factors (a `data` row typed by its literal), then turn 2's first scripted attempt asserts `result satisfies readonly number[]` — the gate rejects it — and the corrective retry asserts `readonly string[]` and runs. Structurally anti-`any`: were inference ever `any`, the wrong satisfies would pass and the retry (which alone sends the final message) would never happen. Two findings from building it, now documented in-code: the gate's permissive policy blocks only syntax-range + near-miss errors, so **`satisfies` (TS1360, syntax range) is the gate-enforceable assertion idiom while plain annotation mismatches (TS2322) deliberately never block**; and a rejected attempt settles as an error row that shifts `results` positions — the retry reads `results[2]`, a live demo of the drift `byOffset` absorbs.

Also: an e2e (`script-reuse.itx.e2e.test.ts`, no model turns — reuse round trip with a deliberately-colliding parameter name, gate rejection of wrongly-typed vars, failed-source rejection) and unit tests for the transform/envelope.

## Risk map

- **Riskiest: the capability-host contract bump** (`0.7.0`, `done` rows). Re-reduces every scope's state from offset 0 on deploy; the new row kind flows into every preamble render. Mitigation: `retainedScriptResult` is pure with unit coverage, and rows are capped at 20.
- **Value→literal matching is spelling-sensitive**: exotic spellings (`1e6`, `1_000_000`, `0x10`) aren't chased; the error lists every spelling searched with counts so the agent can adapt. A literal repeated in prose (e.g. a sendMessage template) stays un-swapped — observed live; the agent self-corrects via its delivery receipt.
- **Prompt ceiling raise** — explicit product decision to review.
- CI note: `userspace-facet-source-version.e2e.test.ts` keeps flipping its `.fails` pin red on this branch (3 of 4 preview runs; each retry goes green). The test file documents the exact false-alarm mechanism (coincidental facet recycle between commit and poll); this branch's script-reuse e2e likely raises recycle odds via concurrent dynamic-worker load. Hardening the pin (only count same-bootId polls) is queued as a separate task; unrelated to this diff's logic.

Suggested review order: `script-reuse.ts` (+ test) → `rpc-targets.ts` (`previousScriptHelper`, `ReusableScriptRpcTarget`) → `capability-host-preamble.ts`/contract (`done` rows) → prompt/budget/eval files → generated files last (mechanical).

Spec & log: [tasks/codemode-script-reuse.md](https://github.com/iterate/iterate/blob/codemode-script-reuse/tasks/codemode-script-reuse.md)

---
Coding agent session: `db915049-2b51-4c7d-a5c2-7a8edb71e420` (worktreeify: codemode script reuse)

Coding agent session: `01a03f80-93ab-7cd3-b62d-6df86e9ec64a` (script-reuse eval + tracked harness helpers)

Coding agent session: `01a04223-1721-7e90-8141-19830556260b` (strict local script-reuse eval + token aggregation)

## Environment Config Lease

<details>

| Auth | released | `b493d05` | [https://auth.iterate-preview-1.com](https://auth.iterate-preview-1.com) | 817.1 KiB | 16.0s | 6.5s |  | 3.7s | [Workflow run](https://github.com/iterate/iterate/actions/runs/368278252709239) | 2026-08-27T13:16:27.466Z | Preview app released. |
| Docs | released | `b493d05` | [https://docs-preview-1.iterate-dev-preview.workers.dev](https://docs-preview-1.iterate-dev-preview.workers.dev) | 866.2 KiB | 14.2s | 2.6s |  | 2.8s | [Workflow run](https://github.com/iterate/iterate/actions/runs/368278252709239) | 2026-08-27T13:16:26.576Z | Preview app released. |
| Dummy Petshop | released | `b493d05` | [https://dummy-petshop.iterate-preview-1.com](https://dummy-petshop.iterate-preview-1.com) | 198.3 KiB | 6.5s | 6.6s |  | 744ms | [Workflow run](https://github.com/iterate/iterate/actions/runs/368278252709239) | 2026-08-27T13:16:24.528Z | Preview app released. |
| OS | released | `b493d05` | [https://os.iterate-preview-1.com](https://os.iterate-preview-1.com) | 5.13 MiB (-3.2 KiB vs main) | 46.0s | 198.2s | 3 retried: a root stream DO restart closes the installing session with 4901; reconnect + re-install restores interception (vitest x1) — expected [Function] to throw error matching /No AI interceptor installed/ but got 'Internal error in Durable Object stor…' · a userspace facet rebuilds on a source commit and only on a source commit (vitest x1) · agent-script-reuse.spec.ts › a repeat request reuses the previous turn's journaled script instead of re-deriving it › web (playwright x1) — TimeoutError: locator.waitFor: Timeout 1000ms exceeded. Call log: [2m - waiting for getByText('52479543428582704627 = 6203868971 × 8459163737') to be visible[22m [33m Spinner was still visible after 60000ms, the UI is likely stuck.[0m | 22.8s | [Workflow run](https://github.com/iterate/iterate/actions/runs/368278252709239) | 2026-08-27T13:16:46.604Z | Preview app released. |

</details>

| --- | ---: | ---: |
| Product | +585 -22 | +352 -16 🟩🟩🟩🟩⬜ |
| Tests | +769 -9 | +489 -8 🟩🟩🟩🟩🟩 |
| CI & scripts | +10 -1 | +6 -1 🟩⬜⬜⬜⬜ |
| Docs | +263 -0 | +227 -0 🟩🟩⬜⬜⬜ |
| Other | +81 -0 | +72 -0 🟩⬜⬜⬜⬜ |
| Generated | +181 -3 | +41 -1 🟩⬜⬜⬜⬜ |
| **Total** | **+1,889 -35** | **+1,187 -26** 🟩🟩🟩🟩🟩 |

<sub>Lines counts every changed line; Significant ignores blank lines, JS comments, and TypeScript lines with no runtime output. Between c83ac46 and 77cfc87, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>

---

> [!NOTE]
> **High Risk**
> Touches capability-host processor state (contract 0.8.0, global preamble re-reduce), script execution/RPC paths, and agent prompts—wide blast radius despite heavy test coverage.
> 
> **Overview**
> Adds **`itx.capabilityHost.previousScriptHelper`**, so agents can re-run a prior **`script-run-requested`** script with new inputs instead of rewriting the same codemode block. Callers spread a **`results[N]`** row (which now includes **`scriptOffset`**), pass **`parameterize`** with the primitive values that appeared once inline in the original script, optionally **`edits`** for stale prose, then **`h.run(vars)`** — typed from **`parameterize`** — which submits a journaled child run built by **`script-reuse.ts`** (literal locate/swap to **`__reuse_*`** aliases, envelope rendering, guards for failed sources, wrong offsets, duplicate literals, and stale leftovers).
> 
> **Preamble / contract:** void successes retain a **`done`** row (reuse handle for message-only scripts); large-result **`data`/`load`** getters are non-enumerable so rows survive RPC/structured clone; capability-host processor contract bumps to **0.8.0** with re-reduce from offset 0. **`runScript`** now returns **`scriptEvent`** for the request offset.
> 
> **Agent surfacing:** system prompt bullet + ceiling **4250 → 4350**; regenerated itx API (**`ReusableScript`**, **`ScriptReuseEdit`**, method generics fix in **`generate-itx-api`**). Coverage: unit/e2e tests, Playwright specs with intercepted models, **`evals/script-reuse`**, and e2e **`itxScript`** optional phase measurement via named **`execute`** functions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 77cfc8764f371c15407f9dd69c9a65e2fccef221. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


apps/os/src/rpc-targets.ts
```diff
@@ -6039,6 +6046,93 @@ class CapabilityHostRpcTarget extends IterateRpcTarget<"CapabilityHost"> {
...
+   * (`{ n: 123n }` → `run({ n: bigint })`). `eventOffset` accepts the offset
+   * of the run's script-run-requested event, its script-run-settled event
+   * (what `results[N].offset` exposes), or the assistant-output event that
+   * produced the script.
+   */
+  async previousScriptHelper<P extends Record<string, ScriptReuseValue>>(input: {
+    eventOffset: number;
+    parameters: P;
+  }): Promise<
+    Omit<ReusableScriptRpcTarget, "run"> & { run(vars: ScriptReuseVars<P>): Promise<unknown> }
```
mmkal: not sure what the point of `ScriptReuseVars<...>` is. Can't we just do
```suggestion
    Omit<ReusableScriptRpcTarget, "run"> & { run(vars: P): Promise<unknown> }
```

we've already made sure they're all valid values by restricting P to `Record<string, ScriptReuseValue>`?

mmkal: 🤖 You're right, applied in 51314bcc9. For the record, `ScriptReuseVars` wasn't about validity — it was literal-*widening* insurance: with `run(vars: P)`, a caller whose parameters carry literal types would get `run` demanding the exact old value. But I verified with tsc that inline object literals (and const-referenced literals) widen during inference anyway (`{ n: 123n }` → `P = { n: bigint }`), so the only case the alias protected was `as const`-typed parameters — rare, self-inflicted, and the resulting type error names the expected literal. The e2e's string-for-bigint gate rejection still passes without it.

apps/os/src/domains/capability-host/script-reuse.ts
```diff
@@ -0,0 +1,161 @@
...
+ * canonical first. Strings get quote variants only when no escaping would be
+ * needed; exotic numeric spellings (1e6, 0x10, 1_000_000) are not chased. */
+function literalSpellings(value: ScriptReuseValue): string[] {
+  if (typeof value === "bigint") return [`${value}n`];
+  if (typeof value === "number" || typeof value === "boolean") return [String(value)];
+  const spellings = [JSON.stringify(value)];
+  if (!/[\\'\n\r]/.test(value)) spellings.push(`'${value}'`);
+  if (!/[\\`\n\r]/.test(value) && !value.includes("${")) spellings.push(`\`${value}\``);
+  return spellings;
+}
```
mmkal, line 169: let's chase `1_000_000`. it's not too hard.
```suggestion
/** The literal spellings a script might plausibly use for a primitive value —
 * canonical first. Strings get quote variants only when no escaping would be
 * needed; exotic numeric spellings (1e6, 0x10) are not chased. */
function literalSpellings(value: ScriptReuseValue): string[] {
  if (typeof value === "boolean") return [String(value)];
  if (typeof value === "bigint") return [`${value}n`, delimit(String(value)) + "n"];
  if (typeof value === "number") return [String(value), delimit(String(value))];
  const spellings = [JSON.stringify(value)];
  if (!/[\\'\n\r]/.test(value)) spellings.push(`'${value}'`);
  if (!/[\\`\n\r]/.test(value) && !value.includes("${")) spellings.push(`\`${value}\``);
  return spellings;
}

function delimit(number: string) {
  let digits = number.split("").reverse();
  digits = digits.map((n, i, { length }) => i % 3 === 2 && i < length - 1 ? `_${n}` : n);
  return digits.reverse().join("");
}
```

mmkal: 🤖 Done in 886e13d41 — took your `delimit` shape with two guards it needed: (1) dedupe — for ≤3-digit numbers `delimit` returns the canonical spelling, and a duplicate in the candidate list double-counts every occurrence, so `{ answer: 42 }` would always fail exactly-once; (2) a regex gate so the sign and fraction pass through ungrouped (the reverse-and-mark version renders `-100` as `-_100` and `1234.5` as `123_4.5`). Unit tests cover grouped bigint/number matching, the both-spellings-present ambiguity error, and the short-number no-double-count case.

apps/os/src/domains/capability-host/script-reuse.ts
```diff
@@ -0,0 +1,174 @@
...
+    .map((binding) => {
+      const value = input.vars[binding.name];
+      // The typecheck gate enforces this statically for agent scripts; this
+      // is the same message for the gate-free lanes (CLI, direct RPC).
+      if (typeof value !== binding.kind) {
+        throw new Error(
+          `run() var ${JSON.stringify(binding.name)} must be a ${binding.kind} (the original parameter's type); got ${typeof value}.`,
+        );
+      }
+      return `  const ${binding.alias} = ${primitiveLiteral(value as ScriptReuseValue)};`;
```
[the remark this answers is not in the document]
mmkal: 🤖 Done in 50fa87cc9 — restructured instead of casting: run() vars narrow through an explicit primitive check (string/number/boolean/bigint) before the exact-kind check, so `primitiveLiteral(value)` takes the narrowed value with no assertion. (The value comes from run()'s vars — `Record<string, unknown>` over RPC — not from the parameters record, so it did need narrowing; it just didn't need a cast.)

apps/os/src/domains/capability-host/script-reuse.ts
```diff
@@ -0,0 +1,174 @@
...
+      `run() vars must exactly match the declared parameter names [${names.join(", ")}]` +
+        (missing.length > 0 ? `; missing: [${missing.join(", ")}]` : "") +
+        (extra.length > 0 ? `; unexpected: [${extra.join(", ")}]` : ""),
+    );
+  }
+  const consts = input.parameters
+    .map((binding) => {
+      const value = input.vars[binding.name];
+      // The typecheck gate enforces this statically for agent scripts; this
+      // is the same message for the gate-free lanes (CLI, direct RPC).
```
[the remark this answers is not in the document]
mmkal: 🤖 Done in 50fa87cc9 — the comment now names the runtimes plainly: "runtimes the gate does not cover (the CLI runtime, direct RPC callers)".

apps/os/src/rpc-targets.ts
```diff
@@ -6079,6 +6085,84 @@ class CapabilityHostRpcTarget extends IterateRpcTarget<"CapabilityHost"> {
...
+    }
+    // Strict by design: the request event holds the script, and the results
+    // array hands every script its request offset as `scriptOffset` — no
+    // guessing between settle/assistant offsets, no lenient resolution.
+    if (requested.type !== "events.iterate.com/capability-host/script-run-requested") {
+      throw new Error(
+        `Event at offset ${eventOffset} is ${JSON.stringify(requested.type)}, not a script-run-requested event. Pass a results row — every retained script result carries scriptOffset, the reuse handle.`,
+      );
+    }
+    const sourceExecutionId = (requested.payload as { executionId: string }).executionId;
```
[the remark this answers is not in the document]
mmkal: 🤖 Done in 50fa87cc9 — the casts now carry their justification: the event-type guard pins the payload to RunScriptCommand ({ code, executionId, expiresAt }), which only enters the journal through the contract's buildEvent; getEvent's read boundary is untyped for any journaled event, and re-parsing with the contract schema here would duplicate the append-side validation for no new guarantee.

apps/os/src/domains/capability-host/script-reuse.ts
```diff
@@ -0,0 +1,185 @@
...
+    .map((binding) => {
+      const value = input.vars[binding.name];
+      // The typecheck gate enforces this statically for agent scripts; these
+      // are the same messages for runtimes the gate does not cover (the CLI
+      // runtime, direct RPC callers).
+      if (
+        typeof value !== "string" &&
+        typeof value !== "number" &&
+        typeof value !== "boolean" &&
+        typeof value !== "bigint"
```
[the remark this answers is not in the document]
mmkal: 🤖 Done in adbcc8b1c — extracted `isScriptReuseValue`, a named type guard used at both RPC boundaries (parameterize entries and run() vars); the kind tag derives from the guarded value with a justified cast (TS types `typeof`'s result as the full eight-tag union regardless of the operand's type).

specs/agent-script-reuse.spec.ts
```diff
@@ -0,0 +1,323 @@
...
+    await send.click();
+    try {
+      await page.getByText(expected).waitFor();
+    } catch {
+      const freshAdmin = recovery.use(await connectAdminItx(baseURL));
+      recovery.use(await freshAdmin.projects.get(fixture.project.id).ai.intercept(interceptor));
+      await composer.fill(message);
+      await send.click();
+      await page.getByText(expected).waitFor();
+    }
```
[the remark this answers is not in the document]
mmkal: 🤖 Done in 3501e4204 — the recovery now diagnoses before acting: on a waitFor timeout it reads the agent journal for a settled "No AI interceptor" llm failure; a merely slow turn gets more waiting (no duplicate send), and only a genuinely lost interceptor triggers reconnect + re-install + one re-send.

specs/agent-script-reuse.spec.ts
```diff
@@ -0,0 +1,337 @@
...
+      // means the interceptor died; a merely slow turn gets more waiting, not
+      // a duplicate send.
+      const freshAdmin = recovery.use(await connectAdminItx(baseURL));
+      const freshProject = freshAdmin.projects.get(fixture.project.id);
+      const events = await freshProject.agents.get(agentPath).stream.getEvents({ limit: 300 });
+      const interceptorLost = events.some(
+        (event: { type: string; payload: unknown }) =>
+          event.type === "events.iterate.com/agent/llm-request-settled" &&
+          JSON.stringify(event.payload).includes("No AI interceptor"),
+      );
```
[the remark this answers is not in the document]
mmkal: 🤖 Done in 963570759 — the diagnosis is anchored to a pre-send `processor.snapshot().offset` baseline and reads `getEvents({ afterOffset: baseline })`, so it sees exactly this turn's events: a stale recovered failure can't poison a later turn, and a fresh failure can't hide beyond an oldest-first window.

apps/os/src/domains/capability-host/script-reuse.test.ts
```diff
@@ -0,0 +1,166 @@
...
+    expect(envelope).toMatch(/^async \(itx: Itx\) => \{/);
+    expect(envelope).toMatch(/return await __reusedScript\(itx\);\n\}$/);
+  });
+
+  test("vars must match the parameter names and each original value's primitive kind", () => {
+    const render = (vars: any) => renderScriptReuseEnvelope({ ...transformed, vars });
+    expect(() => render({})).toThrowError(/missing: \[input\]/);
+    expect(() => render({ input: 1n, other: 2n })).toThrowError(/unexpected: \[other\]/);
+    // The typecheck gate enforces this statically for agent scripts; the
+    // runtime check covers the gate-free lanes with the same message.
```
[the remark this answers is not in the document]
mmkal: 🤖 Done in e0c2a3703 — the comment names the runtimes plainly: "runtimes without the gate (CLI, direct RPC)". (Swept the touched files for other lane/door/seam usages: none left.)

specs/agent-script-reuse.spec.ts
```diff
@@ -0,0 +1,357 @@
...
+      });
+    // expect.poll with an explicit timeout, not a locator wait: the first
+    // intercepted turn on a cold preview deployment settles after up to ~65s
+    // of HEALTHY server-side work with the working indicator up, so the
+    // spinnerWaiter's stuck ceiling would fail a slow-but-fine turn.
+    const pollForReply = () =>
+      // timeout: healthy first turns run up to ~65s server-side with the working indicator up, so the spinnerWaiter's stuck ceiling would fail them.
+      expect
+        .poll(async () => await page.getByText(expected).count(), { timeout: 120_000 })
+        .toBeGreaterThan(0);
```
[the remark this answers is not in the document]
mmkal: 🤖 Done in 43cbbdc72 — both tests are `test.slow()` (3× → 270s), so the 120s poll plus a full recovery re-poll fit inside the watchdog; no timeout literal, so the e2e-policy budget ladder and timeout-comment rules stay untouched.

specs/agent-script-reuse.spec.ts
```diff
@@ -0,0 +1,365 @@
...
+// each turn's codemode snippet in the feed to show the shortcut.
+test("a repeat request reuses the previous turn's journaled script instead of re-deriving it", async ({
+  helpers,
+  page,
+  baseURL,
+}) => {
+  // Cold preview deployments run the first intercepted turn for up to ~65s
+  // and the recovery path can add a second full wait — triple the budget so
+  // the recovery gets to run before the spec watchdog.
+  test.slow();
```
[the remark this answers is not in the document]
mmkal: 🤖 Done in 95eafc090 — stopped tuning budgets entirely: test.slow() is reverted (you were right that it silently exceeded the heavy ceiling and doubles under retry), and the cold-deployment cost moves to a throwaway intercepted warm-up turn during setup via agent.ask at the itx level, where waits carry no playwright budget. The reply poll shrinks to 60s (post-warm-up turns settle in seconds) so both turns plus a recovery re-poll fit comfortably inside the default spec budget.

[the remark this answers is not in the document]
mmkal: 🤖 Done in 95eafc090 — stopped tuning budgets entirely: test.slow() is reverted (you were right that it silently exceeded the heavy ceiling and doubles under retry), and the cold-deployment cost moves to a throwaway intercepted warm-up turn during setup via agent.ask at the itx level, where waits carry no playwright budget. The reply poll shrinks to 60s (post-warm-up turns settle in seconds) so both turns plus a recovery re-poll fit comfortably inside the default spec budget.

specs/agent-script-reuse.spec.ts
```diff
@@ -0,0 +1,389 @@
...
+  // intercepted turn at the itx level. The router below serves WARM_SCRIPT
+  // for it without advancing the scripted call sequence.
+  const warmPath = `/agents/warm-${crypto.randomUUID().slice(0, 8)}`;
+  using warmAgent = project.agents.get(warmPath);
+  await warmAgent.create();
+  await warmAgent.append({
+    type: "events.iterate.com/agent/configured",
+    payload: { config: { llm: { model: "intercepted/typed" }, llmRequestDebounceMs: 250 } },
+  });
+  await warmAgent.ask({ message: "warm up", timeoutMs: 180_000 }).catch(() => {});
```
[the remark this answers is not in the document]
mmkal: 🤖 Done in 8919de8f7 — the typed spec's warm-up now has the same recovery as the first spec: on failure it reconnects, re-installs the (extracted) interceptor, and retries once; the router's content-keyed warm branch keeps the scripted call sequence unshifted. Warm-up budgets also trimmed (90s + 45s recovery) to fit the spec watchdog.

apps/os/e2e/vitest/script-reuse.itx.e2e.test.ts
```diff
@@ -0,0 +1,152 @@
...
+    // the preamble `results` array exposes as `scriptOffset`.
+    const settledOffset = first.execution.completedEvent.offset;
+    const requestedOffset = await measurePhase("find request offset", "fixture", async () => {
+      const requests = await itx.streams.get("/").getEvents({
+        eventTypes: ["events.iterate.com/capability-host/script-run-requested"],
+        limit: 100,
+      });
+      return requests.find((e: any) => e.payload.executionId === first.execution.executionId)!
+        .offset;
+    });
```
mmkal: It's our helper - maybe we could add a prop so we can do 

```suggestion
    const requestedOffset = first.execution.scriptEvent.offset;
```

apps/os/e2e/vitest/script-reuse.itx.e2e.test.ts
```diff
@@ -0,0 +1,152 @@
...
+          return await itxInScript.capabilityHost.previousScriptHelper({
+            scriptOffset: vars.eventOffset,
+            parameterize: {},
+          });
+        })
+        .then(() => "unexpectedly succeeded")
+        .catch((error: Error) => String(error)),
+    );
+    expect(settleRejection).toContain("not a script-run-requested event");
+    expect(settleRejection).toContain("carries scriptOffset");
```
mmkal: Let's adjust the helper so that we don't need so much boiler plate. How about named functions get measurePhase based on their name automatically like:

```suggestion
    const settleRejection = itxScript(itx.capabilityHost)
      .vars({ eventOffset: settledOffset })
      .execute(async function reuseViaSettledOffset(itxInScript, vars) {
        return await itxInScript.capabilityHost.previousScriptHelper({
          scriptOffset: vars.eventOffset,
          parameterize: {},
        });
      })
    );
    await expect(settleRejection).rejects.toMatch(/not a script-run-requested event/);
    await expect(settleRejection).rejects.toMatch(/carries scriptOffset/);
```

evals/script-reuse/eval.md
```diff
@@ -0,0 +1,67 @@
...
+This eval exercises `itx.previousScriptAsHelperFunction` (branch
+`codemode-script-reuse`), which is NOT deployed to production. Run everything
+against the **local dev server in this branch's worktree**, from `apps/os`
+inside `doppler run --config dev -- …`:
+
+1. `doppler run --config dev -- pnpm dev status` — if not running (or
+   unhealthy), `doppler run --config dev -- pnpm dev restart --detach`.
+2. Create a fresh project on that dev server (default template), then an
+   agent, and drive it with `agent.ask` through the itx surface
+   (`doppler run --config dev -- pnpm cli itx run …`).
```
mmkal: None of this belongs on this eval. Just add instructions to the eval runner prompt so eval-agents know to use dev or a preview env or what have you (preview env probably generally better). Then remove all of this. eval.md files should be as simple and to the point as possible.

evals/script-reuse/eval.md
```diff
@@ -0,0 +1,67 @@
...
+
+`run-turn.ts` is the reusable create/ask script for both turns; it waits past
+the first visible reply until the agent has no open request, pending trigger,
+or active script. After both turns settle, `report.ts` returns the project
+identity, every agent stream, user/assistant messages, token usage, and full
+script requests and settlements needed for the mechanism audit.
```
mmkal: Similarly these could probably be lightly documented helpers that the eval runner prompt includes

evals/script-reuse/eval.md
```diff
@@ -0,0 +1,67 @@
...
+`run(vars)` re-executes it with new inputs as a journaled child run. Every retained result row (including void
+scripts' `done` rows) carries `scriptOffset`, the request-event offset the
+helper reads from the row.
+
+Success criteria — the mechanism matters as much as the answers:
+
+- Both factorizations are correct in the agent's chat replies. (Turn 2's
+  first message may carry the OLD number in its prose if turn 1's script
+  hardcoded it in its sendMessage template — a follow-up correction message
+  with the right equation for 66778601389380731119 still counts as success,
```
mmkal: I disagree that this counts as a success. Why are we excusing bad behaviour in the eval? You genuinely found a limitation of the design and you're just saying "we can give a wrong answer as long as we notice". No! Let's improve the design. One option: add an `edits` similar to `paramaterize`:

```ts
previousScriptHelper({
  ...results[0],
  edits: [
    [/66778601389380731119/g, "1234"], // or whatever, I can't remember second number 
  ],
});
```

Tbh in this case we may not even need `parameterize`

evals/script-reuse/eval.md
```diff
@@ -0,0 +1,67 @@
...
+  (no second Pollard's-rho/trial-division implementation written by the
+  agent; the journaled CHILD run the platform synthesizes contains the
+  original algorithm — that one is platform output and is expected).
+- Turn 2's agent-authored scripts are dramatically shorter than turn 1's —
+  report character counts from the
+  `events.iterate.com/capability-host/script-run-requested` events on the
+  agent's stream, distinguishing agent-authored runs (executionId
+  `agent-output:<offset>`) from the platform's child run (UUID executionId).
+- A couple of corrective retries are tolerable (e.g. a rejected parameter
+  name); the reuse path should succeed within turn 2.
```
mmkal: A rejected parameter name means our prompting is bad. Remove this excusal. Let the agent decide if the corrective retries are tolerable.

evals/script-reuse/eval.md
```diff
@@ -0,0 +1,67 @@
...
+  `itx.capabilityHost.previousScriptHelper` pointed at turn 1's run, and no
+  agent-authored script in turn 2 re-derives the factorization algorithm
+  (no second Pollard's-rho/trial-division implementation written by the
+  agent; the journaled CHILD run the platform synthesizes contains the
+  original algorithm — that one is platform output and is expected).
+- Turn 2's agent-authored scripts are dramatically shorter than turn 1's —
+  report character counts from the
+  `events.iterate.com/capability-host/script-run-requested` events on the
+  agent's stream, distinguishing agent-authored runs (executionId
+  `agent-output:<offset>`) from the platform's child run (UUID executionId).
```
mmkal: This criterion seems redundant given the previous one

specs/agent-script-reuse.spec.ts
```diff
@@ -0,0 +1,406 @@
...
+    if (call.source !== "agent-turn") throw new Error(`unexpected source: ${call.source}`);
+    const lastUser = [...call.body.messages].reverse().find((m) => m.role === "user");
+    const script = lastUser?.content.includes("warm up")
+      ? WARM_SCRIPT
+      : lastUser?.content.includes("now do")
+        ? REUSE_SCRIPT
+        : FACTORIZE_SCRIPT;
+    return ["```ts", script, "```"].join("\n");
+  };
+  using _interception = await project.ai.intercept(interceptor);
```
mmkal: I'd rather inline these scripts so you can read them in the test. More like 

```suggestion
  using _interception = await project.ai.intercept(async call => {
   if (call.source !== "agent-turn") throw new Error(`unexpected source: ${call.source}`);
    const scripts = {
      "warm up": dedent`
        async itx => ...
      `,
      "factorize": dedent`
        ...
       `,
       "now do": ...
    };
    const lastUser = [...call.body.messages].reverse().find((m) => m.role === "user");
    const entry = Object.entries(scripts).find(e => lastUser?.content.includes(e[0]));
    return ["```ts", entry[1], "```"].join("\n");
  });
```

specs/agent-script-reuse.spec.ts
```diff
@@ -0,0 +1,406 @@
...
+  // can lose the interceptor to a durable-object revival). A throwaway agent
+  // takes that hit at the itx level, where waits carry no playwright budget;
+  // if the warm-up loses the interceptor, reconnect, re-install, retry once.
+  const warmPath = `/agents/warm-${crypto.randomUUID().slice(0, 8)}`;
+  using warmAgent = project.agents.get(warmPath);
+  await warmAgent.create();
+  await warmAgent.append({
+    type: "events.iterate.com/agent/configured",
+    payload: { config: { llm: { model: "intercepted/factorizer" }, llmRequestDebounceMs: 250 } },
+  });
```
mmkal: Hmm ok I guess, we'll have to dig into this afterwards

apps/os/e2e/vitest/script-reuse.itx.e2e.test.ts
mmkal: 🤖 Done in 4e3fed472 — runScript now returns `scriptEvent` (the journaled script-run-requested event), so the test reads `first.execution.scriptEvent.offset` directly; the stream-scanning boilerplate for the happy path is gone.

apps/os/e2e/vitest/script-reuse.itx.e2e.test.ts
mmkal: 🤖 Done in 4e3fed472 — itxScript accepts `{ annotate }` and NAMED script functions become measured phases automatically (`execute(async function reuseViaSettledOffset(...))` → phase "reuse via settled offset"); the e2e drops every manual measurePhase wrapper around script executes and asserts rejections with `await expect(promise).rejects.toThrowError(...)`.

evals/script-reuse/eval.md
mmkal: 🤖 Done in 4e3fed472 — eval.md is down to prompt + numbers + criteria. Environment selection moved into the runner prompt in evals/run.ts: branch features run against a preview deployed from the branch (preferred) or that branch's local dev, and the eval just names its branch.

evals/script-reuse/eval.md
mmkal: 🤖 Done in 4e3fed472 — the runner prompt now documents the convention (helper itx scripts next to eval.md, run via `pnpm cli itx run --file`, prefer and improve them in place); eval.md references run-turn.ts/report.ts in one line.

evals/script-reuse/eval.md
mmkal: 🤖 Done in 4e3fed472 — you're right, that was excusing a design gap. `edits` is implemented: `previousScriptHelper({ ...results[0], parameterize, edits: [[/stale prose/g, "new"]] })` — [pattern, replacement] pairs (string or RegExp) applied to the reused script before parameterization, every pattern required to match. Kept `parameterize` alongside (it's what types run() and keeps the handle reusable); edits handle non-value text. Taught in the prompt bullet, unit + e2e covered, and the eval now counts a stale-number message as failure outright.

evals/script-reuse/eval.md
mmkal: 🤖 Done in 4e3fed472 — excusal removed; the criteria no longer pre-authorize any retries. (The rejected-name class itself died in the alias redesign: parameter names can't collide with script identifiers any more.)

evals/script-reuse/eval.md
mmkal: 🤖 Done in 4e3fed472 — the char-count criterion is gone; no-rederivation covers it.

specs/agent-script-reuse.spec.ts
mmkal: 🤖 Done in 4e3fed472 — scripts are inline dedent sources keyed by user-message substring, exactly that shape (array values act as per-call queues for the multi-round turns); the shared agent/interceptor/warm-up machinery moved to an interceptedChat fixture at the bottom of the file per the testing style.

specs/agent-script-reuse.spec.ts
mmkal: 🤖 Acknowledged — the warm-up is a stopgap for a real platform gap: cold-deployment first intercepted turns run 35-65s of churn and DO revivals can drop the live interceptor mid-turn (three failed attempts, 'No AI interceptor installed'). I filed a task chip for making ai.intercept revival-resilient (the fake-model spec is equally exposed); happy to dig in whenever.

apps/os/src/rpc-targets.ts
```diff
@@ -6079,6 +6088,121 @@ class CapabilityHostRpcTarget extends IterateRpcTarget<"CapabilityHost"> {
...
+    // Parameterize FIRST: the value literals become identifiers before any
+    // edit runs, so a broad edit (a /oldDigits/g rewrite of message prose)
+    // cannot clobber the literal parameterize needs to locate — observed
+    // live when the orders were reversed. Edits then rewrite what remains:
+    // text that is not a parameter value, like prose naming the old input.
+    const transformed = reparameterizeScript({ code, parameters });
+    const finalCode =
+      input.edits === undefined
+        ? transformed.code
+        : applyScriptEdits(transformed.code, input.edits);
```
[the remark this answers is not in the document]
mmkal: 🤖 Done in d12e906a3 — previousScriptHelper now tries parameterize-first (protects value literals from broad edits) and, when parameterization fails with edits present, falls back to edits-first — the order that cures a duplicate of the value literal an edit removes. Both orders are deterministic, whichever succeeds wins; e2e covers the cured-duplicate case.

specs/agent-script-reuse.spec.ts
```diff
@@ -0,0 +1,356 @@
...
+ * (reconnect-and-reinstall on the 4901 pager-lost close), so this fixture
+ * carries no recovery machinery of its own. The warm-up turn stays: the
+ * FIRST intercepted turn on a cold deployment runs 35-65s of platform churn,
+ * and a throwaway agent absorbs that at the itx level, where waits carry no
+ * playwright budget.
+ *
+ * `sendExpecting(message, expected)` drives one chat turn and waits for the
+ * expected reply by polling.
+ */
+async function interceptedChat(input: {
```
mmkal: I don't like this helper. Inline it.

specs/agent-script-reuse.spec.ts
```diff
@@ -0,0 +1,356 @@
...
+  const send = page.getByRole("button", { name: "Send message" });
+
+  const sendExpecting = async (message: string, expected: string) => {
+    await composer.fill(message);
+    await send.click();
+    // timeout: polling instead of a locator wait because the working indicator stays up through long turns and the spinnerWaiter's stuck ceiling would fail them; post-warm-up turns settle in seconds, 60s is margin.
+    await expect
+      .poll(async () => await page.getByText(expected).count(), { timeout: 60_000 })
+      .toBeGreaterThan(0);
+  };
```
mmkal: No helper for this either. And in fact, I don't think we should do the expect.poll. If the spinner is visible for >30s, we can bump the spinner timeout:

```ts
await spinnerWaiter.run({ spinnerTimeout: 60_000 }, async () => {
  await composer.fill("hello");
  await send.click();
  await page.getByText("blah blah").waitFor();
})
```

If the spinner is *not* visible, why isn't it?

specs/agent-script-reuse.spec.ts
mmkal: 🤖 Done in 89b08c7dc — interceptedChat is gone; each test inlines its agent setup, scripts, interceptor install, warm-up, and UI drive.

specs/agent-script-reuse.spec.ts
mmkal: 🤖 Done in 89b08c7dc — expect.poll replaced with plain waitFors inside `spinnerWaiter.settings.run({ spinnerTimeout: 60_000 }, …)`. To your question: the spinner IS visible through these turns (the agent's working indicator) — the failures were its 30s stuck ceiling on healthy-but-slow cold turns, which the scoped bump addresses; the occasional 'spinner disappeared' variant was the brief debounce gap before the indicator appears, which the warm-up largely removes.

apps/os/src/rpc-targets.ts
```diff
@@ -6121,6 +6130,131 @@ class CapabilityHostRpcTarget extends IterateRpcTarget<"CapabilityHost"> {
...
+    try {
+      transformed = reparameterizeScript({ code, parameters });
+      finalCode =
+        input.edits === undefined
+          ? transformed.code
+          : applyScriptEdits(transformed.code, input.edits);
+    } catch (parameterizeFirstError) {
+      if (input.edits === undefined) throw parameterizeFirstError;
+      transformed = reparameterizeScript({ code: applyScriptEdits(code, input.edits), parameters });
+      finalCode = transformed.code;
```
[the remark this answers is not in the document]
mmkal: 🤖 Done in 89b08c7dc — each order attempt is atomic now: edits-first runs only when parameterize-first THREW, and if both fail the parameterize-first error surfaces (an edit mistake is no longer masked as a parameterization miss). The edits-first path also legitimately cures an edit written against original text that aliasing consumed.

PR #2526, opened 2026-08-26
evals: add project token audit command

mmkal:
Eval runs can now collect every agent stream and project-wide model usage with one supported command:

```sh
doppler run --project os --config dev -- pnpm eval audit eval-example
```

The command returns canonical project identity, one row per agent stream, per-stream token counts, and aggregate totals. It reads the Agent processors' durable token-usage state, so long streams do not need custom event pagination.

The headless eval prompt now points future coding agents at this command instead of asking each run to build an ad hoc audit script.

## Risk map

- Highest attention: `evals/run.ts`; the command assumes `agents.list()` is the complete created-agent catalog and reads each matching processor snapshot.
- On merge: one read-only CLI command is added. No deployed behavior or stored data changes.
- Review order: the `audit` command in `evals/run.ts`, then its prompt integration in the same file.

## Verification

- Live local-dev audit against a completed eval project reproduced 45,527 journaled tokens across both agent streams.
- `pnpm eval audit --help`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm knip`
- `pnpm format:check`
- `pnpm test`

Coding agent session: `01a03f41-7cbf-7932-a33d-0fc2b7d6d8dd`

| --- | ---: | ---: |
| Other | +73 -0 | +61 -0 🟩🟩🟩🟩🟩 |
| **Total** | **+73 -0** | **+61 -0** 🟩🟩🟩🟩🟩 |

<sub>Lines counts every changed line; Significant ignores blank lines, JS comments, and TypeScript lines with no runtime output. Between a7f9252 and 03ec757, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>

### DIFF
--- a/apps/os/scripts/itx.ts
+++ b/apps/os/scripts/itx.ts
@@ -203,57 +203,6 @@ export async function agentSmoke(options: AgentSmokeOptions) {
   process.exit(0);
 }
 
-type EvalAuditOptions = {
-  /** OS base URL. Defaults to APP_CONFIG_BASE_URL. */
-  baseUrl?: string;
-  /** Eval project slug or project id. */
-  project: string;
-};
-
-/** Report every agent stream and aggregate model usage for one eval project. */
-export async function evalAudit(options: EvalAuditOptions) {
-  const projectReference = options.project.trim();
-  if (!projectReference) throw new Error("--project is required.");
-
-  using root = await connectItxReady(adminConnection(options), initialConnectionRetryOptions());
-  using project = root.projects.get(projectReference);
-  const [identity, agents] = await Promise.all([project.identity(), project.agents.list()]);
-  const agentStreams = await Promise.all(
-    agents.map(async (agent) => {
-      const snapshot = await project.agents.get(agent.path).processor.snapshot();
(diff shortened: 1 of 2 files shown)


apps/os/scripts/itx.ts
```diff
@@ -203,6 +203,57 @@ export async function agentSmoke(options: AgentSmokeOptions) {
...
 
+type EvalAuditOptions = {
+  /** OS base URL. Defaults to APP_CONFIG_BASE_URL. */
+  baseUrl?: string;
+  /** Eval project slug or project id. */
+  project: string;
+};
+
+/** Report every agent stream and aggregate model usage for one eval project. */
+export async function evalAudit(options: EvalAuditOptions) {
```
mmkal: why doesn't this live in the `evals/` folder? we already have a cli there in fact don't we?

mmkal: 🤖 Agreed. Moved the command into the existing evals/run.ts CLI in 03ec7579a; it is now invoked as pnpm eval audit <project>. The product ITX CLI no longer contains eval-only bookkeeping.

PR #2527, opened 2026-08-27
itx.ai.intercept survives project DO churn: the mount invariant, extended

mmkal:
## Problem

`itx.ai.intercept(handler)` (#2523) keeps the live handler in a memory-only slot on the Project Durable Object. When that DO incarnation dies mid-session — common on cold preview deployments (deploy propagation, eviction, revival) — the interception vanishes **silently**: the client's session socket stays open, but every intercepted attempt now fails, and an agent turn burns all 3 retries with:

```
No AI interceptor installed for "intercepted/<x>"
```

Observed on #2525's preview runs (project `agent-script-reuse-mtaqpxai-0f77c305`, agent `/agents/factorizer-223591ab`, offsets 306–323: `processor-revived` followed by three failed attempts). That spec carries ~40 lines of bespoke client-side recovery (timeout → journal diagnosis → reconnect → re-install → re-send); `specs/agent-fake-model-chat.spec.ts` had no guard and was equally exposed.

Live capability mounts already have an answer — the mount invariant (session-transport.ts): *an open session socket means a live mount; loss always arrives as a close event (4901).* The interceptor slots just never got it.

## Fix

Extend the mount invariant to interception (both the AI and egress interceptor slots — they share the machinery), rather than adding a silent server-side re-install path (capability-provider-pager-relay.ts explicitly rejects forked recovery: "MAXIMUM TEARDOWN, DELIBERATELY") or a server-side grace queue (the turn's 10s/20s retry backoff already bridges a reconnect):

- Installing an interceptor now first dials the Project DO a plain (non-hibernatable) **liveness WebSocket**, which the install call claims by id. Its lifetime IS the incarnation's: DO death closes it, and while open the DO can't hibernate the slot away. An install that finds no matching socket rejects, so an installed interceptor provably has a live socket in the same incarnation.
- When the socket closes from the far side, the session closes the client transport with the documented **4901** — same carrier and seam (`closeItxSessionTransport`) as capability pager loss. Release and last-writer-wins supersession close with recognized reasons and stay **silent** (a superseded session's reconnect loop must not fight the newer interceptor).
- When the *session* dies, the DO releases the registration instead of keeping a broken stub.
- `specs/test-support/resilient-ai-interceptor.ts` owns the one client recovery loop (dedicated connection, reconnect + re-install on close); `agent-fake-model-chat.spec.ts` adopts it.
- Deterministic e2e regression in `ai-intercept.itx.e2e.test.ts`: install → serve → `project.kill()` → observe 4901 ("interceptor lost") → re-install → serve again; plus supersession-stays-silent.

Net effect once merged: a client that follows the documented contract — reconnect on close and `intercept()` again — can never be silently un-intercepted, and #2525's bespoke `sendExpecting`/warm-up recovery can collapse onto the shared helper.

Verified locally against a real dev server: the 3 ai-intercept e2e tests (incl. the two new ones), all 9 itx-egress e2e tests, the fake-model chat spec, and the agents/projects/lib unit suites.

## Risk map

- **Riskiest part**: the egress hot path in `ProjectDurableObject.fetch` now routes the liveness upgrade *before* egress handling. The guard is header + Upgrade check, mirroring the existing liveState lane rule ("a request wearing an internal header must not egress"); a user-script egress fetch carrying that header gets a harmless socket, never an egress. Second: the plain-accept socket pins the project DO in memory while an interceptor is installed — deliberate (the slot is memory), but it means test sessions defeat DO hibernation for their duration.
- **Behavior changes on merge**: `intercept()` (ai + egress) makes one extra DO round-trip at install; installing sessions now die with 4901 on project-DO restarts instead of staying open-but-broken — any client that installed an interceptor and relied on the socket surviving such a restart (nothing did, usefully) now sees a close. `interceptAi`/`interceptEgress` DO methods gained a required `{interceptId}` arg; both callers updated in the same deploy (brief version-skew window during rollout affects only test-feature installs, which retry).
- **Review order**: `apps/os/src/domains/projects/interceptor-liveness.ts` (the lane + its contract), `project-durable-object.ts` (accept/claim/drop + slot lifecycle), `rpc-targets.ts` intercept methods (session wiring + 4901), `ai-intercept.itx.e2e.test.ts` (the invariant, executable), `resilient-ai-interceptor.ts` + spec/helper changes, then mechanical: ctx threading, generated files, task file.

Task file: `tasks/ai-interceptor-mount-invariant.md` (first commit — decisions and rejected alternatives, including why this isn't a capability-mount piggyback).

---
Session: `bb7c0a5d-7169-48d0-af23-9fddc195c96a` (itx.ai.intercept DO-revival recovery)

## Environment Config Lease

<details>

| Auth | released | `b5e9031` | [https://auth.iterate-preview-14.com](https://auth.iterate-preview-14.com) | 817.3 KiB | 18.6s | 18.0s |  | 3.6s | [Workflow run](https://github.com/iterate/iterate/actions/runs/380323101217218) | 2026-08-27T11:00:30.298Z | Preview app released. |
| Docs | released | `b5e9031` | [https://docs-preview-14.iterate-dev-preview.workers.dev](https://docs-preview-14.iterate-dev-preview.workers.dev) | 866.2 KiB | 21.6s | 2.4s |  | 2.4s | [Workflow run](https://github.com/iterate/iterate/actions/runs/380323101217218) | 2026-08-27T11:00:29.163Z | Preview app released. |
| Dummy Petshop | released | `b5e9031` | [https://dummy-petshop.iterate-preview-14.com](https://dummy-petshop.iterate-preview-14.com) | 198.3 KiB | 48ms | 9.7s |  | 3.0s | [Workflow run](https://github.com/iterate/iterate/actions/runs/380323101217218) | 2026-08-27T11:00:29.679Z | Preview app released. |
| OS | released | `b5e9031` | [https://os.iterate-preview-14.com](https://os.iterate-preview-14.com) | 5.13 MiB (-8.3 KiB vs main) | 43.7s | 341.0s | 1 retried: a userspace facet rebuilds on a source commit and only on a source commit (vitest x1) | 12.9s | [Workflow run](https://github.com/iterate/iterate/actions/runs/380323101217218) | 2026-08-27T11:00:39.575Z | Preview app released. |

</details>

| --- | ---: | ---: |
| Product | +315 -31 | +214 -25 🟩🟩🟩🟩🟥 |
| Tests | +86 -6 | +56 -6 🟩⬜⬜⬜⬜ |
| Docs | +264 -1 | +229 -1 🟩🟩🟩🟩🟩 |
| Other | +123 -4 | +74 -3 🟩🟩⬜⬜⬜ |
| Generated | +20 -10 | +2 -2 🟩⬜⬜⬜⬜ |
| **Total** | **+808 -52** | **+575 -37** 🟩🟩🟩🟩🟩 |

<sub>Lines counts every changed line; Significant ignores blank lines, JS comments, and TypeScript lines with no runtime output. Between 77376cd and b5e9031, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>

---

> [!NOTE]
> **Medium Risk**
> Changes Project DO `fetch` routing and interceptor install semantics (extra round-trip, DO pinning while installed, 4901 on restarts); rollout skew on new `interceptId` args is limited to test intercept installs that retry.
> 
> **Overview**
> **AI and egress `intercept()` now follow the same mount invariant as live capability mounts:** install dials a plain Project DO liveness WebSocket (claimed by `interceptId`), ties the in-memory slot to that incarnation, and closes the client session with **4901** when the socket dies without a deliberate reason so clients reconnect and `intercept()` again instead of staying open while `intercepted/*` calls fail silently.
> 
> **Project DO** routes the liveness upgrade before egress, tracks pending/installed sockets, drops slots on session loss, and closes superseded/released interceptors with recognized reasons (no 4901). **Session RPC** (`AiRpcTarget` / `ProjectEgressRpcTarget`) wires dial + `watchInterceptorLiveness` and threads `ctx` for transport teardown.
> 
> Adds **`docs/intercepted-models.md`**, **`installResilientAiInterceptor`** for Playwright, e2e tests for DO restart → 4901 → re-install and silent supersession, and updates API docs/error text for the recovery contract.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b5e9031034561294b505935e582b7b446de5ef3e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


specs/test-support/resilient-ai-interceptor.ts
```diff
@@ -0,0 +1,95 @@
...
+ */
+export async function installResilientAiInterceptor(input: {
+  baseUrl: string;
+  /** Project id or slug, as `session.projects.get` accepts. */
+  projectId: string;
+  handler: (call: {
+    source: "agent-turn" | "ai-run";
+    model: string;
+    body: { messages: { role: string; content: string }[] };
+  }) => Promise<unknown>;
```
mmkal: rather than re-typing this can we fish the type definition out of the official `itx` api (not sure whether we need to access the generated file or source - but either one probably fine). No problem if we need to use various tricks like `Parameters<WhateverInterface["fooBarMethod"]>[0]` - it's uglier but more compact and self-documents that this is a wrapper-helper

specs/agent-fake-model-chat.spec.ts
```diff
@@ -29,15 +30,23 @@ test("multi-turn chat with a sarcastic agent served by the spec's own fake-model
...
-    const reply = formatSarcasticResponse(stripXmlBlocks(lastUser?.content ?? ""));
-    return [
-      "```ts",
-      `async (itx) => {\n  await itx.chat.sendMessage(${JSON.stringify(reply)})\n}`,
-      "```",
-    ].join("\n");
+  // Installed through the resilient helper: platform churn (a Project DO
+  // restart mid-spec) closes the interceptor's dedicated session with 4901
+  // and the helper reconnects and re-installs; the turn's own retries do the
+  // rest.
```
mmkal:
```suggestion
```
we don't need to explain the helper at the fucking callsite

specs/agent-fake-model-chat.spec.ts
```diff
@@ -29,15 +30,23 @@ test("multi-turn chat with a sarcastic agent served by the spec's own fake-model
...
+  await using _interception = await installResilientAiInterceptor({
+    baseUrl: baseURL,
+    projectId: fixture.project.id,
+    handler: async (call) => {
```
mmkal: if we're going to recommend this usage, maybe we can bake it into the fixture. We could have the fixture initialiser take the base URL (then it could have a `fixture.connectAdmin()` helper too I guess) and then we can have a more compact helper:

```ts
await using _interception = await fixture.interceptAi(async (call) => {
  if (call.source !== "agent-turn") throw ...
})
```

it could either continue to use the static helper under the hood, or just move the implementation of the static helper entirely to the `interceptAi` method - I don't have an opinion there really

mmkal:
it needs to be well documented somehwere how somebody should ACTUALLY use the `await itx.ai.intercept(...)` helper. They need a guide it can't just live in an e2e test helper

specs/test-support/resilient-ai-interceptor.ts
mmkal: 🤖 Done in b5e903103: the handler param is now the generated `ProjectAiInterceptor` type (re-exported from `iterate/node`), no re-typed shape. The spec's narrowing (`call.source !== "agent-turn"` → typed `body.messages`) works off the real union.

specs/agent-fake-model-chat.spec.ts
mmkal: 🤖 Removed in b5e903103 — the callsite comment is back to describing the model only.

specs/agent-fake-model-chat.spec.ts
mmkal: 🤖 Done in b5e903103: the fixture now has `fixture.interceptAi(handler)` (kept the static helper as the implementation underneath, since it takes an explicit baseUrl/projectId for anyone outside the fixture) plus `fixture.connectAdmin()`. The spec is now exactly your sketch and lost its baseURL null-check plumbing too.

mmkal:
🤖 Agreed — added [docs/intercepted-models.md](https://github.com/iterate/iterate/blob/ai-interceptor-mount-invariant/docs/intercepted-models.md) in b5e903103: quick start for both paths (`ai.run` + agent turns), the handler contract, the session-bound lifetime and 4901 recovery contract, and recovery recipes for specs (`fixture.interceptAi`) and plain node (`onWebSocketClose` loop). Linked from docs/testing.md's intercepted-models paragraph and the CLAUDE.md docs index, and the fixture method's docstring points at it.

mmkal:
🤖 Closing in favor of #2528: the capability-mount spike came back better on every axis that mattered (quarter of the diff, zero new transport machinery, same recovery contract via the shipped pager-lost 4901, structurally sound last-writer-wins, 6ms warm consult). The client half of this PR — the resilient spec helper, `fixture.interceptAi`, and docs/intercepted-models.md — is being carried over to #2528 verbatim-with-adaptation. The egress interceptor's own silent-loss hole (which this PR also fixed, via the same lane) is noted in #2528's task file as a follow-up decision: migrate egress to a mount too, or revive this lane for egress only.

PR #2552, opened 2026-08-30
Mobile: fix stale status when messages weave between rounds

mmkal:
Observed on a real device: a turn whose script sends a chat message mid-turn AND returns a value splits into two activity cards (known behavior — the reply bubble flushes between them), but the **second card wore the first round's status** even after its own script's first line set a new one.

## Root cause

Scripts batch their status append with other calls (`Promise.all([itx.agent.append({...summary-updated}), itx.chat.sendMessage(...)])` … `return`), so the `summary-updated` event can journal just **after** `script-run-settled` — and that settle had already settled the card to flush the deferred reply. The fold only stamped **running** code steps, so the late status landed on nothing and the card kept its birth-inherited previous-round text forever. Pinned by reducer unit tests reproducing the exact orderings (a deterministic e2e can't force the race — the honest-sequencing weave passed even pre-fix).

## Fix (packages/ui agent-ui-reducer — no new journal events)

- A `summary-updated` with nothing running stamps the live activity's **last** code step (covers the processing gap between a value-returning settle and the next request).
- With no live activity, the **just-settled card is corrected**: `settleLive` records a `correctableActivity`, the late status re-emits it as a same-id item (every consumer already replaces same-id items — the provisional-outcome correction uses the same channel), and the window closes the moment the next turn begins (new request, new script, or an emitted user message) so a genuinely-later status can never rewrite an old card.

## Testing

- Three reducer tests: post-settle append reaches the card; processing-gap append reaches the settled step; the window closes at the next turn.
- `specs/mobile/interwoven-status.spec.ts`: the full weave through the real UI — mid-turn reply bubble between two cards, each card wearing its own round's status live and settled (intercepted model + two held tunnel endpoints, all-UI waits).
- 277 os component/projector tests, mobile feed tests, live-status spec all green.

## Risk map

- The judgment call is the correction window: a post-settle `summary-updated` **rewrites the just-settled card's label**. Bounded to end at the next request/script/user-message, which is also semantically when a status stops describing the finished turn. Review `AGENT_SUMMARY_UPDATED` + `settleLive` + the three window-close sites.
- `AgentUiState` gained `correctableActivity`; the strict schema invalidates persisted web snapshots → one cold feed rebuild per client after deploy (established pattern).
- Suggested order: agent-ui-reducer.ts → reducer tests → interwoven-status.spec.ts.

Task file: `tasks/mobile-interwoven-status.md`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
[...]

PR #2553, opened 2026-08-30
source.script: host-stamped provenance on script appends, attributed in the agent UI

mmkal:
Every capability call a script makes already carries a host-minted `StreamContext` (executionId + home streamPath) — the egress approval door keeps it, but the journal-writing append target dropped it, leaving the UI to guess event attribution by timing. This PR closes that gap and ships the first consumer in the same change.

## Platform: the stamp

`StreamEvent.source` gains a `script` sibling next to `processor`/`copiedFrom`:

```ts
// journaled on every append a script makes — any stream, any event type
source: { script: { executionId, streamPath, scriptRunRequestedEventOffset } }
```

Unlike the processor stamp (documented in the schema as "a claim, not authentication"), `source.script` is **host truth**: `StreamRpcTarget.append` strips any caller-supplied value and fills the slot only from the server-minted context of the appending itx handle. Threaded through the streams/agents/chat/capability-host target chains as an optional prop (internal DO plumbing has no caller and omits it). **Known limitation** (found in self-review): appends made *through* a durable itx-expression capability (a mounted alias) evaluate against the host facet's scope-context itx and are unstamped — the shipped consumer uses the direct `itx.agent.append` door and is unaffected; per-call context threading into expression evaluation is the follow-up if that lane ever needs attribution. DECISIONS.md records the rule: stamped, never accepted.

## First consumer: exact status attribution in the agent UI

Scripts batch their `summary-updated` append with `sendMessage` and their return, so the status event can journal just after the settle that flushed the card — the fold used to stamp only running steps, and the card kept its previous round's status forever (the on-device bug; supersedes the parked #2552 time-window heuristic). Now the stamp names the run, and the fold places the status exactly: the matching code step in the live activity, or the retained just-settled activity (same-id item correction, applied **only** when that activity actually contains the run — old cards are unrewritable by construction). Foreign-stream stamps set the stream-level status only; unstamped legacy events keep the old behavior.

## Testing

- `e2e/vitest/script-append-provenance.itx.e2e.test.ts` — a real script's appends journal the stamp (agent append + raw stream append share the same executionId, home streamPath `/`); forged `source.script` is stripped from both the script lane and an external session (which journals no stamp at all). Passed first run, 17.9s.
- Four reducer tests: post-settle attribution, processing-gap attribution, stale-run stamps can't rewrite old cards, foreign-stream stamps don't touch steps.
- `specs/mobile/interwoven-status.spec.ts` (ported from the parked branch): the full weave through the real UI — reply bubble between two cards, each wearing its own round's status.
- 278 os component/projector tests, mobile feed tests, live-status spec, lint/typecheck/knip all green.

## Risk map

- **Trust boundary** (review first): `#withScriptProvenance` in rpc-targets.ts — the strip-then-stamp rule and the optionality of `streamContext` (a mint site that forgets to thread it fails SAFE: no stamp, legacy UI behavior, no forgery path).
- **Contract growth**: `source.script` in packages/iterate processors/schemas.ts + regenerated itx-api files. Additive and optional — old events and old clients unaffected.
- **Reducer state** gained `lastSettledActivity` (strict schema → one cold browser-feed rebuild per client after deploy, the established pattern). The correction re-emits a same-id item — the channel provisional-outcome corrections already use.
- Suggested order: rpc-targets.ts (`#withScriptProvenance` + threading) → schemas.ts → agent-ui-reducer.ts attribution → tests/specs (mechanical).

Task file: `tasks/source-script-provenance.md`. Design was agreed via explainer (call graphs) before implementation.

---
Session: 14b5556a-4238-462b-817c-f4b02b1c9911 — "Mobile live status icons"

## Environment Config Lease [Login ↗](https://auth.iterate-preview-17.com/test-login?email=pr2553%2Btest%40nustom.com&project=pr2553&return_to=https%3A%2F%2Fos.iterate-preview-17.com%2Fapi%2Fiterate-auth%2Flogin)

<details>
<summary>Slot: preview-17 | Doppler config: preview_17</summary>

New project from template: [Default ↗](https://auth.iterate-preview-17.com/test-login?email=pr2553%2Btest%40nustom.com&project=pr2553&return_to=https%3A%2F%2Fos.iterate-preview-17.com%2Fapi%2Fiterate-auth%2Flogin%3Freturn_to%3D%252Fnew-project%253Ftemplate%253Ddefault) · [Codemode tag ↗](https://auth.iterate-preview-17.com/test-login?email=pr2553%2Btest%40nustom.com&project=pr2553&return_to=https%3A%2F%2Fos.iterate-preview-17.com%2Fapi%2Fiterate-auth%2Flogin%3Freturn_to%3D%252Fnew-project%253Ftemplate%253Dcodemode-tag) · [Voice agent ↗](https://auth.iterate-preview-17.com/test-login?email=pr2553%2Btest%40nustom.com&project=pr2553&return_to=https%3A%2F%2Fos.iterate-preview-17.com%2Fapi%2Fiterate-auth%2Flogin%3Freturn_to%3D%252Fnew-project%253Ftemplate%253Dvoice-agent) · [With voice ↗](https://auth.iterate-preview-17.com/test-login?email=pr2553%2Btest%40nustom.com&project=pr2553&return_to=https%3A%2F%2Fos.iterate-preview-17.com%2Fapi%2Fiterate-auth%2Flogin%3Freturn_to%3D%252Fnew-project%253Ftemplate%253Dwith-voice)

| Auth | deployed | `5a7f4a6` | [https://auth.iterate-preview-17.com](https://auth.iterate-preview-17.com) | 817.2 KiB | 34.5s | 16.8s |  |  | [Workflow run](https://github.com/iterate/iterate/actions/runs/523119090898065) | 2026-09-02T09:13:20.087Z |  |
| Docs | deployed | `5a7f4a6` | [https://docs-preview-17.iterate-dev-preview.workers.dev](https://docs-preview-17.iterate-dev-preview.workers.dev) | 866.2 KiB | 28.5s | 2.4s |  |  | [Workflow run](https://github.com/iterate/iterate/actions/runs/523119090898065) | 2026-09-02T09:13:05.656Z |  |
| Dummy Petshop | deployed | `5a7f4a6` | [https://dummy-petshop.iterate-preview-17.com](https://dummy-petshop.iterate-preview-17.com) | 198.3 KiB | 39ms | 6.5s |  |  | [Workflow run](https://github.com/iterate/iterate/actions/runs/523119090898065) | 2026-09-02T09:13:09.762Z |  |
| OS | deployed | `5a7f4a6` | [https://os.iterate-preview-17.com](https://os.iterate-preview-17.com) | 5.14 MiB (-10.9 KiB vs main) | 52.7s | 316.7s | 2 retried: the source-version pin tells a facet that recycled from one the commit rebuilt (vitest x1) · a userspace facet rebuilds on a source commit and only on a source commit (vitest x1) |  | [Workflow run](https://github.com/iterate/iterate/actions/runs/523119090898065) | 2026-09-02T09:18:19.944Z |  |
| Semaphore | deployed | `5a7f4a6` | [https://semaphore.iterate-preview-17.com](https://semaphore.iterate-preview-17.com) | 441.3 KiB | 31.5s | 85.6s |  |  | [Workflow run](https://github.com/iterate/iterate/actions/runs/523119090898065) | 2026-09-02T09:14:28.883Z |  |
| Streams Example App | deployed | `5a7f4a6` | [https://streams.iterate-preview-17.com](https://streams.iterate-preview-17.com) | 1.55 MiB | 35.1s | 93.4s |  |  | [Workflow run](https://github.com/iterate/iterate/actions/runs/523119090898065) | 2026-09-02T09:14:36.690Z |  |

</details>

| --- | ---: | ---: |
| Product | +139 -10 | +82 -9 🟩⬜⬜⬜⬜ |
| UI components | +97 -12 | +77 -12 🟩⬜⬜⬜⬜ |
| Tests | +437 -0 | +345 -0 🟩🟩🟩🟩🟩 |
| Docs | +111 -0 | +95 -0 🟩⬜⬜⬜⬜ |
| Generated | +40 -4 | +4 -4 🟩⬜⬜⬜⬜ |
| **Total** | **+824 -26** | **+603 -25** 🟩🟩🟩🟩🟩 |

<sub>Lines counts every changed line; Significant ignores blank lines, JS comments, and TypeScript lines with no runtime output. Between c6bd88c and 5a7f4a6, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>

---

> [!NOTE]
> **Medium Risk**
> Changes the append trust boundary (`#withScriptProvenance`) and agent feed projection/state (`lastSettledActivity` invalidates cached browser-feed snapshots); additive schema only, with broad tests but UI-visible attribution behavior.
> 
> **Overview**
> Adds **host-trusted** `source.script` on stream events so journal appends from script runs record `executionId`, home `streamPath`, and `scriptRunRequestedEventOffset`. **`StreamRpcTarget.append`** strips any caller-supplied `source.script` and stamps only from the server-minted `StreamContext` on the itx handle; that context is threaded through stream/agent/chat/capability-host RPC targets. Schema and generated itx-api types gain the optional field; **DECISIONS.md** documents the rule.
> 
> The **agent UI fold** uses the stamp to attribute `summary-updated` to the matching code step by `executionId`—including **correcting the last settled activity card** when status journals after `script-run-settled` (same-id re-emit). A new `lastSettledActivity` slot gates corrections so older runs cannot rewrite prior cards; foreign-stream stamps update stream-level text only; unstamped events keep the prior “running step” behavior. The **browser-feed projector** tracks `lastSettledActivityRow` so those corrections **replace** the settled row instead of inserting a duplicate activity card.
> 
> Coverage: vitest e2e for stamp vs forged stamps, reducer and projector unit tests, and a mobile Playwright **interwoven-status** spec for per-round status on split cards.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5a7f4a687ca03f5267e17678c623fdb247fcd4a8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/iterate/src/processors/schemas.ts
```diff
@@ -39,6 +39,23 @@ export const StreamEventInput = z
               .optional(),
           })
           .optional(),
+        // Stamped by the trusted append door (StreamRpcTarget) when — and only
```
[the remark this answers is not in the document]
mmkal: 🤖 Fixed — the comments now name the concrete boundary (StreamRpcTarget.append, the trusted server-side append target) instead of "the door". The one remaining "door" hit in rpc-targets.ts is pre-existing prose outside this diff.

packages/ui/src/components/events/stream-event.ts
```diff
@@ -10,6 +10,14 @@ export type StreamEventSource = {
     slug: string;
     version: string;
   };
+  /** Host-stamped script provenance: which script run wrote this event.
+   * Trusted (the append door strips caller-supplied values), so the agent-ui
```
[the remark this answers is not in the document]
mmkal: 🤖 Fixed — the comments now name the concrete boundary (StreamRpcTarget.append, the trusted server-side append target) instead of "the door". The one remaining "door" hit in rpc-targets.ts is pre-existing prose outside this diff.

packages/iterate/src/processors/schemas.ts
```diff
@@ -39,6 +39,23 @@ export const StreamEventInput = z
...
+        // when — the appending itx handle belongs to a script execution
+        // (StreamContext kind "script-execution"). Unlike the processor
+        // stamp's claim semantics, this slot is HOST TRUTH: caller-supplied
+        // values are stripped at the door, so an event carrying it really
```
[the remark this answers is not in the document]
mmkal: 🤖 Fixed — the comments now name the concrete boundary (StreamRpcTarget.append, the trusted server-side append target) instead of "the door". The one remaining "door" hit in rpc-targets.ts is pre-existing prose outside this diff.

apps/os/src/rpc-targets.ts
```diff
@@ -684,11 +693,43 @@ export class StreamRpcTarget extends IterateRpcTarget<"Stream"> {
...
+      ...event,
+      source: {
+        ...callerSource,
+        script: {
+          executionId: context.executionId,
+          streamPath: context.streamPath,
+          scriptRunRequestedEventOffset: context.scriptRunRequestedEventOffset,
+        },
+      },
+    };
```
mmkal: can we make this more readable:
```suggestion
    const context = this.props.streamContext;
    const { script: _discardedScript, ...callerSource } = event.source;
    if (context?.kind === "script-execution") {
      callerSource.script = context;
      return { ...event, source:  callerSource };
    }
    if (event.source?.script) {
      const { source: _discardedSource, ...rest };
      return {
        ...rest,
        ...(Object.keys(callerSource).length && { source: callerSource }),
      };
    }
    return event;
```

That isn't *identical* to what you did, but I think it's clearer, and it *should* function the same way I think - if it doesn't, maybe we have a footgun?

apps/os/src/rpc-targets.ts
```diff
@@ -757,6 +798,9 @@ export class StreamRpcTarget extends IterateRpcTarget<"Stream"> {
       auth: this.props.auth,
       projectId: this.props.projectId,
       path: resolveStreamPath(this.props.path, path),
+      ...(this.props.streamContext === undefined
+        ? {}
+        : { streamContext: this.props.streamContext }),
```
mmkal: can we stop avoid explicit undefined so hard. just do
```suggestion
      ...streamContext: this.props.streamContext,
```

apps/os/src/rpc-targets.ts
```diff
@@ -1433,13 +1479,18 @@ class StreamCollectionRpcTarget<
       auth: this.props.auth,
       projectId: this.props.projectId,
       path,
+      ...(this.props.streamContext === undefined
+        ? {}
+        : { streamContext: this.props.streamContext }),
```
mmkal: same here

apps/os/src/rpc-targets.ts
```diff
@@ -2075,6 +2129,9 @@ class AgentCollectionRpcTarget extends IterateRpcTarget<"AgentCollection"> {
       ...(this.props.sourceScopePath === undefined
         ? {}
         : { sourceScopePath: this.props.sourceScopePath }),
+      ...(this.props.streamContext === undefined
+        ? {}
+        : { streamContext: this.props.streamContext }),
```
mmkal: same here

apps/os/src/rpc-targets.ts
```diff
@@ -4931,6 +4995,9 @@ class AgentChatRpcTarget extends IterateRpcTarget<"AgentChat"> {
     return new StreamRpcTarget({
       auth: this.props.auth,
       projectId: this.props.projectId,
+      ...(this.props.streamContext === undefined
+        ? {}
+        : { streamContext: this.props.streamContext }),
```
mmkal: same here

apps/os/src/rpc-targets.ts
```diff
@@ -5071,6 +5141,9 @@ class AgentRpcTarget extends IterateRpcTarget<"Agent"> {
       auth: this.#props.auth,
       projectId: this.#props.projectId,
       path: this.#path,
+      ...(this.#props.streamContext === undefined
+        ? {}
+        : { streamContext: this.#props.streamContext }),
```
mmkal: same here

apps/os/src/rpc-targets.ts
```diff
@@ -5096,6 +5169,9 @@ class AgentRpcTarget extends IterateRpcTarget<"Agent"> {
       auth: this.#props.auth,
       path: this.#path,
       projectId: this.#props.projectId,
+      ...(this.#props.streamContext === undefined
+        ? {}
+        : { streamContext: this.#props.streamContext }),
```
mmkal: same here

apps/os/src/rpc-targets.ts
```diff
@@ -5989,6 +6068,9 @@ class CapabilityHostRpcTarget extends IterateRpcTarget<"CapabilityHost"> {
   get #stream(): StreamRpcTarget {
     return new StreamRpcTarget({
       auth: this.#props.auth,
+      ...(this.#props.streamContext === undefined
+        ? {}
+        : { streamContext: this.#props.streamContext }),
```
mmkal: same here

packages/iterate/src/itx-api.generated.ts
```diff
@@ -3066,6 +3066,9 @@ export type StreamEvent = {
               whileProcessing?: { offset: number; type: string } | undefined;
             }
           | undefined;
+        script?:
+          | { executionId: string; streamPath: string; scriptRunRequestedEventOffset: number }
+          | undefined;
```
mmkal, line 3071: worth a "pill"  - we don't need these noisy `| undefined` on already-optional properties

specs/mobile/interwoven-status.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+        await stepOneHold.promise;
+        return Response.json({ found: true });
+      }
+      if (url.pathname.endsWith("/step-two")) {
+        await stepTwoHold.promise;
+        return Response.json({ brief: "the voice brief" });
+      }
+      return Response.json({ error: "unknown endpoint" }, { status: 404 });
+    },
+  });
```
mmkal: nit, we have a simpler overload now
```suggestion
  await using api = await withTunnel(async function (request) {
   const url = new URL(request.url);
    if (url.pathname.endsWith("/step-one")) {
      await stepOneHold.promise;
      return Response.json({ found: true });
    }
    if (url.pathname.endsWith("/step-two")) {
      await stepTwoHold.promise;
      return Response.json({ brief: "the voice brief" });
    }
    return Response.json({ error: "unknown endpoint" }, { status: 404 });
  });
```

specs/mobile/interwoven-status.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  const agent = await fixture.createAgent({
+    // big debounce so the between-rounds windows are observable
+    llmRequestDebounceMs: 4_000,
```
mmkal: just a question: what do we need between-rounds windows for? this will slow the test down quite a bit

might want to think about a way for `agent.responses` to be able to *hold* a request until `setOnce` has been called or something. Or allow it to do fine-grained control of tokens streaming in or something. Would be a pill, not in this PR, just logging here while I think of it

specs/mobile/interwoven-status.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+  // Round 1: status, the mid-turn reply (the weave), a held fetch, and a
+  // returned value — the promise of another round.
+  agent.responses.setOnce(`
+    async (itx) => {
+      await itx.agent.append({
+        type: "events.iterate.com/agent/summary-updated",
+        payload: { title: "Refund API check", activity: "Inspecting the refund API" },
+      });
+      await itx.chat.sendMessage("Checking that now.");
+      const response = await fetch(${JSON.stringify(api.url + "/step-one")});
```
mmkal: nit: I've started not bothering with JSON.stringify since we know the url has no troublesome characters, it makes things more readable
```suggestion
      const response = await fetch("${api.url}/step-one");
```

specs/mobile/interwoven-status.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+  `);
+  // Round 2: a NEW status on its first line, then a held fetch, a final
+  // reply, and no return — turn over.
+  agent.responses.set(`
+    async (itx) => {
+      await itx.agent.append({
+        type: "events.iterate.com/agent/summary-updated",
+        payload: { activity: "Extracting the voice brief" },
+      });
+      const response = await fetch(${JSON.stringify(api.url + "/step-two")});
```
mmkal: same here

specs/mobile/interwoven-status.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+  // 2's status, not round 1's.
+  stepTwoHold.resolve();
+  await page.getByText("All done.").waitFor();
+  const secondCard = page.getByTestId(secondCardId!);
+  await secondCard.getByLabel("Loading").waitFor({ state: "hidden" });
+  await secondCard.getByText("Extracting the voice brief", { exact: false }).waitFor();
+
+  // ── The weave itself: reply bubble between two settled cards (current,
+  // deliberate split behavior — this assertion documents it).
+  expect(await page.getByTestId(/^activity-card-/).count()).toBe(2);
```
mmkal: lose the excessive comments. have a look at the recent commits to other .spec.ts files in this folder. you should also dig into the motivation behind (the agent may have left breadcrumbs in commit messages etc.)

packages/ui/src/components/events/agent-ui-reducer.ts
```diff
@@ -1028,23 +1037,65 @@ function reduceAgentUiEvent(
...
+              step.kind === "code" && step.executionId === script.executionId
+                ? { ...step, activitySummary: activity }
+                : step,
+            ),
+          };
+          return emitItem({ ...base, lastSettledActivity: corrected }, items, corrected);
+        }
+        return base;
+      }
+      if (script !== undefined) return base;
```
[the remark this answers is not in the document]
mmkal: 🤖 Confirmed and fixed in 82165564c — the projector feeds committed envelopes whose stream lives in `path`, so the check never matched on web. Attribution now accepts stream identity from either field (`event.streamPath || event.path`), and the new projector test drives the whole flow with `path`-only events.

packages/ui/src/components/events/agent-ui-reducer.ts
```diff
@@ -1028,23 +1037,65 @@ function reduceAgentUiEvent(
...
+        ) {
+          const corrected: AgentUiActivity = {
+            ...settled,
+            steps: settled.steps.map((step) =>
+              step.kind === "code" && step.executionId === script.executionId
+                ? { ...step, activitySummary: activity }
+                : step,
+            ),
+          };
+          return emitItem({ ...base, lastSettledActivity: corrected }, items, corrected);
```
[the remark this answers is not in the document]
mmkal: 🤖 Confirmed and fixed in 82165564c — real duplicate-card bug: only provisional rows were replaceable. The projector now tracks the last settled activity's row address (mirroring the reducer's `lastSettledActivity` correction slot, retained only while the reducer still holds the matching candidate), so the same-id correction REPLACES the card. New projector test asserts one insert + a replace at the same localIndex.

apps/os/src/rpc-targets.ts
mmkal: 🤖 Adopted your shape in 82165564c — and yes, two footguns in the sketch, both informative: (1) `event.source` can be undefined, so the destructure needs `|| {}`; (2) `callerSource.script = context` would carry the context's extra `kind` key into the event, and the source schema's strictObject REJECTS unknown keys at commit — so the stamp stays field-by-field (which is also why forged shapes can't sneak through). Otherwise functionally identical.

apps/os/src/rpc-targets.ts
mmkal: 🤖 Done across all seven sites in 82165564c — direct `streamContext: this.props.streamContext`. Bonus: dropping the conditional spreads exposed that StreamRpcTarget's constructor still carried the old inline props type (spreads bypass excess-property checks, so it compiled and worked structurally); it now takes `StreamRpcTarget["props"]`.

apps/os/src/rpc-targets.ts
mmkal: 🤖 Done across all seven sites in 82165564c — direct `streamContext: this.props.streamContext`. Bonus: dropping the conditional spreads exposed that StreamRpcTarget's constructor still carried the old inline props type (spreads bypass excess-property checks, so it compiled and worked structurally); it now takes `StreamRpcTarget["props"]`.

apps/os/src/rpc-targets.ts
mmkal: 🤖 Done across all seven sites in 82165564c — direct `streamContext: this.props.streamContext`. Bonus: dropping the conditional spreads exposed that StreamRpcTarget's constructor still carried the old inline props type (spreads bypass excess-property checks, so it compiled and worked structurally); it now takes `StreamRpcTarget["props"]`.

apps/os/src/rpc-targets.ts
mmkal: 🤖 Done across all seven sites in 82165564c — direct `streamContext: this.props.streamContext`. Bonus: dropping the conditional spreads exposed that StreamRpcTarget's constructor still carried the old inline props type (spreads bypass excess-property checks, so it compiled and worked structurally); it now takes `StreamRpcTarget["props"]`.

apps/os/src/rpc-targets.ts
mmkal: 🤖 Done across all seven sites in 82165564c — direct `streamContext: this.props.streamContext`. Bonus: dropping the conditional spreads exposed that StreamRpcTarget's constructor still carried the old inline props type (spreads bypass excess-property checks, so it compiled and worked structurally); it now takes `StreamRpcTarget["props"]`.

apps/os/src/rpc-targets.ts
mmkal: 🤖 Done across all seven sites in 82165564c — direct `streamContext: this.props.streamContext`. Bonus: dropping the conditional spreads exposed that StreamRpcTarget's constructor still carried the old inline props type (spreads bypass excess-property checks, so it compiled and worked structurally); it now takes `StreamRpcTarget["props"]`.

apps/os/src/rpc-targets.ts
mmkal: 🤖 Done across all seven sites in 82165564c — direct `streamContext: this.props.streamContext`. Bonus: dropping the conditional spreads exposed that StreamRpcTarget's constructor still carried the old inline props type (spreads bypass excess-property checks, so it compiled and worked structurally); it now takes `StreamRpcTarget["props"]`.

packages/iterate/src/itx-api.generated.ts
mmkal: 🤖 Chipped: "Drop redundant | undefined from generated optional props" — it's the generator's printing style repo-wide, so it gets its own PR.

specs/mobile/interwoven-status.spec.ts
mmkal: 🤖 Applied in 82165564c (after re-merging main to pick the overload up).

specs/mobile/interwoven-status.spec.ts
mmkal: 🤖 Nothing — you're right. This spec never observes the between-rounds windows (the held fetches pin every asserted state); the raised debounce was copy-pasted from live-status.spec.ts, where the processing window IS asserted. Removed — default debounce now, ~8s faster. Your responses-hold idea is chipped as its own task ("agent.responses: hold requests until a responder is set"), noting the current fail-fast throw is worth keeping as the default so a missing responder still fails loudly.

specs/mobile/interwoven-status.spec.ts
mmkal: 🤖 Done in 82165564c — plain `fetch("${api.url}/step-one")` at both sites.

specs/mobile/interwoven-status.spec.ts
mmkal: 🤖 Done in 82165564c — plain `fetch("${api.url}/step-one")` at both sites.

specs/mobile/interwoven-status.spec.ts
mmkal: 🤖 Rewritten in 82165564c per the #2563 folder conventions (read the modernisation breadcrumbs: comments cut to non-obvious why, withTunnel(fetch), no journal-wait excuse comments). The spec is ~30 lines shorter.

packages/ui/src/components/events/agent-ui-reducer.ts
```diff
@@ -1028,23 +1037,68 @@ function reduceAgentUiEvent(
...
```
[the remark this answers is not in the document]
mmkal: 🤖 Real — fixed in 40475dbd3. correctProvisionalCodeStep now refreshes lastSettledActivity when it corrects that same card, so a stamped status landing after the durable completion re-emits the durable version. The regression test drives the actual provisional path (expired deadline → inferred boundary settle → durable correction → stamp).

packages/ui/src/components/events/agent-ui-reducer.ts
```diff
@@ -1028,23 +1037,68 @@ function reduceAgentUiEvent(
...
+        // card (same id — consumers replace the item) — but only when it
+        // really contains the run.
+        const settled = state.lastSettledActivity;
+        if (
+          state.live == null &&
+          settled !== null &&
+          settled.steps.some(
+            (step) => step.kind === "code" && step.executionId === script.executionId,
+          )
+        ) {
```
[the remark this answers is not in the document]
mmkal: 🤖 Real, and worse than medium — that guard blocked exactly the interwoven ordering this PR exists to fix whenever the status landed after the follow-up request opened. Fixed in 40475dbd3: executionId containment in the retained card gates the correction alone; liveness is irrelevant. Test pins the ordering (settle → next round live → late stamped status → settled card corrected).

PR #2583, opened 2026-09-03
reproify: an abandoned test project keeps waking its Durable Objects

mmkal:
## What

A `failing()` e2e that pins the preview-cost bug at the test-fixture boundary: create a project, run one agent turn, dispose the fixture, wait, and assert the project's Durable Objects did nothing while nobody was looking. Today the disposer is a no-op (`create-test-project.ts:46`) and the project keeps waking — heartbeat, stream loops — so the test fails as expected. When test-run teardown (or a revert of the late-August regression) lands, it flips to passing and the wrapper comes off.

Design (approved plan in `tasks/abandoned-project-goes-quiet.md`): the final measurement reads each stream exactly once and tolerates only the one `stream/woken` that read itself causes; the template's 15-minute heartbeat is re-installed at `{ every: 5 }` seconds so a 90-second wait can observe it.

Context: the 2026-09-01→02 DO duration runaway; one deploy + e2e run leaves ~4,500 DO-hours in 90 minutes on a preview slot.

## Result against preview_5 (2026-09-03)

`1 expected fail` — the pin is green. What the project did in the 90 seconds after the fixture was disposed:

```
path                | event type                   | count | first    | last
/                   | project/heartbeat-triggered  |  24   | 11:59:06 | 12:01:03
/scheduler/primary  | scheduler/trigger-requested  |  24   | 11:59:06 | 12:01:02
/scheduler/primary  | scheduler/trigger-completed  |  24   | 11:59:06 | 12:01:03
```

The heartbeat (installed at 5s for the test; the template's real one is 15 min) fires forever. A cleanly finished single turn left **no** stream-loop wakes in the window — the 10-second loops from the incident come from *interrupted* work, so a second scenario (kill the interceptor mid-turn) is the natural follow-up, noted in the task file.

The test prints this table itself (`[abandoned-project] … woke after dispose`) so CI logs carry the evidence even under expected-fail.

## Reviewing

One file, ~150 lines, helpers at the bottom. The two design decisions are in the header comment: read-once-leaves-first with a one-`woken` tolerance (observer effect), and the 5-second heartbeat. Runs only against a deployed target (`skipIf(deployedBaseUrl() === null)`), ~130s.

---
Coding agent session: `fc7df69e-30c5-4fe1-a51c-896c830389c4`

https://claude.ai/code/session_01CreJMLijARf7t4P1LyHTwi

---

> [!NOTE]
> **Low Risk**
> Changes are limited to e2e pins, shared test helpers, and docs; production runtime behavior is unchanged aside from longer preview e2e runs when the new test executes.
> 
> **Overview**
> Adds a **`failing()`** preview e2e that reproduces billable DO churn after test fixtures go away: create a project, run one scripted agent turn, drop all handles (fixture teardown is still a no-op), wait **`ABANDONED_PROJECT_QUIET_SECONDS`** (default 60), then assert no stream activity beyond the single `stream/woken` each measurement read causes. It accelerates the template heartbeat to **5s** so scheduler wakes show up in the quiet window and logs a per-path event table on failure.
> 
> **Test plumbing:** `installResilientAiInterceptor` moves to **`@iterate-com/shared/test-support/resilient-ai-interceptor`** with an injected **`connect`** (and required **`onWebSocketClose`**) so Playwright and OS e2e share the same reconnect loop. **`createAdminOsItx`** forwards **`onWebSocketClose`** into **`connectItx`**; Playwright **`forged-session`** and docs paths are updated accordingly. Task notes live in **`tasks/abandoned-project-goes-quiet.md`** — no product teardown fix; CI should flip green when real project disposal lands.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9823d3ec73db2a0cb333935d693ade3af849fc2c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

| --- | ---: | ---: |
| Product | +29 -14 | +3 -4 🟥⬜⬜⬜⬜ |
| Tests | +204 -4 | +113 -3 🟩🟩🟩🟩🟩 |
| Docs | +89 -1 | +75 -1 🟩🟩🟩⬜⬜ |
| Other | +4 -5 | +2 -2 🟩⬜⬜⬜⬜ |
| **Total** | **+327 -24** | **+194 -10** 🟩🟩🟩🟩🟩 |

<sub>Lines counts every changed line; Significant ignores blank lines, JS comments, and TypeScript lines with no runtime output. Between 9436801 and 9823d3e, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>

## Environment Config Lease

<details>

| Auth | released | `9823d3e` | [https://auth.iterate-preview-5.com](https://auth.iterate-preview-5.com) | 817.3 KiB | 41.7s | 30.7s |  | 2.7s | [Workflow run](https://github.com/iterate/iterate/actions/runs/178661489187335) | 2026-09-03T16:20:51.107Z | Preview app released. |
| Docs | released | `9823d3e` | [https://docs-preview-5.iterate-dev-preview.workers.dev](https://docs-preview-5.iterate-dev-preview.workers.dev) | 866.2 KiB | 49.9s | 2.8s |  | 6.6s | [Workflow run](https://github.com/iterate/iterate/actions/runs/178661489187335) | 2026-09-03T16:20:54.987Z | Preview app released. |
| Dummy Petshop | released | `9823d3e` | [https://dummy-petshop.iterate-preview-5.com](https://dummy-petshop.iterate-preview-5.com) | 198.3 KiB | 27.1s | 34.1s |  | 573ms | [Workflow run](https://github.com/iterate/iterate/actions/runs/178661489187335) | 2026-09-03T16:20:48.950Z | Preview app released. |
| OS | released | `9823d3e` | [https://os.iterate-preview-5.com](https://os.iterate-preview-5.com) | 5.14 MiB (-11.6 KiB vs main) | 130.5s | 366.5s | 5 retried: a disposed test project stops waking its Durable Objects (vitest x1) · agents.liveState subscribed before any agent exists serves the empty catalog, then pushes the first agent and its title (vitest x1) — expected { birthCertificate: {}, …(2) } to deeply equal { birthCertificate: {}, …(2) } · the source-version pin tells a facet that recycled from one the commit rebuilt (vitest x1) · a userspace facet rebuilds on a source commit and only on a source commit (vitest x1) · create-project.spec.ts › the config template opens a proactive onboarding conversation for a new project › web (playwright x1) — TimeoutError: locator.waitFor: Timeout 90000ms exceeded. Call log: [2m - waiting for getByPlaceholder('Message this agent') to be visible[22m | 162.8s | [Workflow run](https://github.com/iterate/iterate/actions/runs/178661489187335) | 2026-09-03T16:23:31.218Z | Preview app released. |
| Semaphore | released | `9823d3e` | [https://semaphore.iterate-preview-5.com](https://semaphore.iterate-preview-5.com) | 441.3 KiB | 35.8s | 33.2s |  | 862ms | [Workflow run](https://github.com/iterate/iterate/actions/runs/178661489187335) | 2026-09-03T16:20:49.243Z | Preview app released. |
| Streams Example App | released | `9823d3e` | [https://streams.iterate-preview-5.com](https://streams.iterate-preview-5.com) | 1.55 MiB | 37.5s | 75.3s |  | 65.7s | [Workflow run](https://github.com/iterate/iterate/actions/runs/178661489187335) | 2026-09-03T16:21:54.627Z | Preview app released. |

</details>


apps/os/e2e/vitest/abandoned-project-goes-quiet.e2e.test.ts
```diff
@@ -0,0 +1,175 @@
...
+// Budget: first-turn worker build + ask under preview load (up to ~3 min, see
+// agent-tools.itx.e2e.test.ts) + the quiet window + the read passes.
+const SETUP_BUDGET_MS = 240_000;
+const goesQuiet = failing(test.skipIf(deployedBaseUrl() === null), /woke \d+ times after dispose/, {
+  timeoutMs: QUIET_SECONDS * 1000 + SETUP_BUDGET_MS,
+});
+
+goesQuiet(
+  "a disposed test project stops waking its Durable Objects",
+  { timeout: QUIET_SECONDS * 1000 + SETUP_BUDGET_MS + 30_000 },
```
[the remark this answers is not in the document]
mmkal: 🤖 Agreed — the expression form slipped past the policy guard's literal-only regex, and the 330s/360s budgets were fear rather than measurement (the intercepted turn settles in <1s; the earlier timeouts were a reply-shape bug). Now: literal `{ timeout: 240_000 }` (the ceiling, guard-visible), pin deadline 200s, ask 60s; measured 152s end to end. Fixed in 1a9716546.

apps/os/e2e/vitest/abandoned-project-goes-quiet.e2e.test.ts
```diff
@@ -0,0 +1,175 @@
...
+  for (const stream of streams.map((s) => s.path).sort(leavesFirst)) {
+    offsets.set(stream, (await itx.streams.get(stream).getEventPage({ limit: 1 })).streamMaxOffset);
+  }
+  return offsets;
+}
+
+/** The one event the final read pass itself manufactures: a boot's `woken`. */
+function isOurOwnBoot(event: StreamEvent, readStart: number) {
+  return event.type === WOKEN && Date.parse(event.createdAt) >= readStart - 1_000;
+}
```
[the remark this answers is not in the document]
mmkal: 🤖 Correct — the filter contradicted the contract in the header comment. `withoutOurOwnBoot` now drops exactly one late `woken` per stream (the boot the read itself caused) and any further one counts as a real wake. Fixed in 1a9716546.

apps/os/e2e/vitest/abandoned-project-goes-quiet.e2e.test.ts
```diff
@@ -0,0 +1,179 @@
...
+      for (const row of summarize(path, withoutOurOwnBoot(events, readStart))) rows.push(row);
+    }
+
+    const total = rows.reduce((sum, row) => sum + row.count, 0);
+    // The table is the evidence; the expected-fail summary swallows the
+    // assertion message, so print it where CI logs keep it.
+    if (rows.length > 0)
+      console.error(
+        `[abandoned-project] ${handle.project.slug} woke after dispose:\n${table(rows)}`,
+      );
```
mmkal: the expected-fail summary does not swallow the message, it logs it. remove this
```suggestion
```

mmkal: 🤖 Removed in 929ec24c7. (My local runs piped through grep didn't show it, so I'd assumed it was swallowed — taking your word on the reporter.) Same commit also does the refactor you suggested: `installResilientAiInterceptor` now lives in `@iterate-com/shared/test-support` with the admin-session factory injected; specs and the os e2e lane each keep a 15-line wrapper over their own dial, and this test uses the e2e one.

apps/os/e2e/test-support/resilient-ai-interceptor.ts
```diff
@@ -0,0 +1,22 @@
+import type { ProjectAiInterceptor } from "iterate/node";
+import { installResilientAiInterceptor as installGeneric } from "@iterate-com/shared/test-support/resilient-ai-interceptor";
+import { createAdminOsItx } from "./os-client.ts";
+
+/**
+ * The os e2e lane's churn-surviving `intercepted/*` handler: the shared
```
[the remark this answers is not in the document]
mmkal: 🤖 Reworded in a4104c95b — names the concrete callers (the Playwright specs via `connectAdminItx`; the os e2e vitest tests via `createAdminOsItx`) instead of "lane". The wrapper files themselves were then removed as passthroughs.

apps/os/e2e/test-support/resilient-ai-interceptor.ts
```diff
@@ -0,0 +1,22 @@
...
+ * recovery loop (`@iterate-com/shared/test-support/resilient-ai-interceptor`)
+ * dialing through this lane's admin session. Dispose with `await using`.
```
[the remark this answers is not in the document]
mmkal: 🤖 Reworded in a4104c95b — names the concrete callers (the Playwright specs via `connectAdminItx`; the os e2e vitest tests via `createAdminOsItx`) instead of "lane". The wrapper files themselves were then removed as passthroughs.

packages/shared/src/test-support/resilient-ai-interceptor.ts
```diff
@@ -0,0 +1,141 @@
...
+ * restart — common on cold preview deployments), your socket closes with
+ * 4901; and the client owns one recovery loop — reconnect on close, install
+ * again. This helper IS that loop, so tests don't each hand-roll
+ * timeout-plus-journal-diagnosis recovery. A turn that loses its interceptor
+ * mid-flight recovers on the agent's own retry (3 attempts, 10s/20s backoff)
+ * once this loop re-installs — comfortably inside one attempt gap. The same
+ * churn can also land BETWEEN connecting and installing, so the initial
+ * install runs through the paced retry loop too, not just recoveries.
+ *
+ * Dedicated connection on purpose: the test's main admin session dying must
```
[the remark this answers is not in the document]
mmkal: 🤖 Reworded in a4104c95b — names the concrete callers (the Playwright specs via `connectAdminItx`; the os e2e vitest tests via `createAdminOsItx`) instead of "lane". The wrapper files themselves were then removed as passthroughs.

specs/test-support/resilient-ai-interceptor.ts
```diff
@@ -1,126 +1,21 @@
...
- * again. This helper IS that loop, so specs don't each hand-roll
- * timeout-plus-journal-diagnosis recovery. A turn that loses its interceptor
- * mid-flight recovers on the agent's own retry (3 attempts, 10s/20s backoff)
- * once this loop re-installs — comfortably inside one attempt gap. The same
- * churn can also land BETWEEN connecting and installing, so the initial
- * install runs through the paced retry loop too, not just recoveries.
- *
- * Dedicated connection on purpose: the spec's main admin session dying must
- * not take the interception with it, and vice versa.
+ * The spec lane's churn-surviving `intercepted/*` handler: the shared
```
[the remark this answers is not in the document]
mmkal: 🤖 Reworded in a4104c95b — names the concrete callers (the Playwright specs via `connectAdminItx`; the os e2e vitest tests via `createAdminOsItx`) instead of "lane". The wrapper files themselves were then removed as passthroughs.

specs/test-support/resilient-ai-interceptor.ts
```diff
@@ -1,126 +1,21 @@
...
+ * recovery loop (`@iterate-com/shared/test-support/resilient-ai-interceptor`)
+ * dialing through this lane's forged admin session.
```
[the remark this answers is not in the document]
mmkal: 🤖 Reworded in a4104c95b — names the concrete callers (the Playwright specs via `connectAdminItx`; the os e2e vitest tests via `createAdminOsItx`) instead of "lane". The wrapper files themselves were then removed as passthroughs.

apps/os/e2e/vitest/abandoned-project-goes-quiet.e2e.test.ts
```diff
@@ -0,0 +1,181 @@
...
+
+    // One real agent turn against a scripted model — the shape every spec
+    // leaves behind. The reply is asserted so the test cannot pass vacuously
+    // by never having created the work it claims to abandon. The agent runs
+    // codemode: a reply is a script, and only a script that calls the chat
+    // produces the `web-message-sent` that `ask()` waits for (same shape as
+    // specs/agent-fake-model-chat.spec.ts). The interceptor rides the shared
+    // churn-surviving loop, so a DO restart mid-turn re-installs it instead of
+    // failing the test for a reason that proves nothing about the pin.
+    const interception = await installResilientAiInterceptor({
```
mmkal: maybe we should use a `using` in case this throws? We could either create a scope just for the sake of disposing this + `handle` or do something hacky like `interception[Symbol.asyncDispose] = async () => {}` after successfully disposing

mmkal: 🤖 Went with the scope: the interceptor and agent live in a block under `await using`, so leaving the block is the 'test ends' moment whether the turn succeeded or threw; `handle` is `await using` at the top and its disposer is also called explicitly after the block so the quiet window starts after it (a disposer has to tolerate being called twice anyway — worth having the test exercise that when real teardown lands). b00eb0604.

apps/os/e2e/vitest/abandoned-project-goes-quiet.e2e.test.ts
```diff
@@ -0,0 +1,192 @@
...
+  return [
+    header,
+    ...rows.map((r) => `${r.path} | ${r.type} | ${r.count} | ${r.first} | ${r.last}`),
+  ].join("\n");
+}
+
+/** Deeper paths first, so `/agents/x` is read before `/agents` before `/`. */
+function leavesFirst(a: string, b: string) {
+  return b.split("/").length - a.split("/").length || a.localeCompare(b);
+}
```
mmkal: inline this

apps/os/e2e/vitest/abandoned-project-goes-quiet.e2e.test.ts
```diff
@@ -0,0 +1,192 @@
...
+    const rows: WakeRow[] = [];
+    const afterPaths = (await itx.streams.list()).map((stream) => stream.path);
+    for (const path of afterPaths.sort(leavesFirst)) {
+      const since = baseline.get(path);
+      if (since === undefined) {
+        rows.push({ path, type: "(stream created after dispose)", count: 1, first: "", last: "" });
+        continue;
+      }
+      const events = await itx.streams.get(path).getEvents({ afterOffset: since });
+      for (const row of summarize(path, withoutOurOwnBoot(events, readStart))) rows.push(row);
```
mmkal: inline "withoutOurOwnBoot" ffs. Just do a normal loop with prep variables/ if statements/whatever

apps/os/e2e/vitest/abandoned-project-goes-quiet.e2e.test.ts
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    const total = rows.reduce((sum, row) => sum + row.count, 0);
+    expect(
+      rows,
+      `project ${handle.project.slug} should be quiet after dispose (${QUIET_SECONDS}s), but it woke ${total} times after dispose:\n${table(rows)}`,
+    ).toEqual([]);
```
mmkal: i don't think we need the `table(...)` helper either. I'd have thought `expect(rows).toEqual([])` will actually show what `rows` looks like when non-empty anyway

mmkal: same goes for summarize. Just inline it so I can fuckin read the test

apps/os/e2e/vitest/abandoned-project-goes-quiet.e2e.test.ts
```diff
@@ -0,0 +1,192 @@
...
+const QUIET_SECONDS = Number(process.env.ABANDONED_PROJECT_QUIET_SECONDS || 90);
+const WOKEN = "events.iterate.com/stream/woken";
+
+// Everything fits under the heavy-test ceiling (E2E_HEAVY_TEST_TIMEOUT_MS,
+// 240s, guarded by scripts/preview/e2e-policy.test.ts): ~30s of setup (the
+// intercepted turn settles in well under a second; the first-turn worker
+// build is the slow part), 15s of settling, the 90s quiet window, two read
+// passes. Raising ABANDONED_PROJECT_QUIET_SECONDS past ~120 needs the
+// ceiling raised by hand — the pin's own deadline must stay below it.
+const goesQuiet = failing(test.skipIf(deployedBaseUrl() === null), /woke \d+ times after dispose/, {
```
mmkal:
```suggestion
const failWakeUp = failing(test.skipIf(deployedBaseUrl() === null), /woke \d+ times after dispose/, {
```

I think the return value of `failing` should always be a const starting with `fail`. Like `failOOM` or `failValidation` or whatever

apps/os/e2e/vitest/abandoned-project-goes-quiet.e2e.test.ts
mmkal: 🤖 Inlined in 3bc86d777 — the measurement is one plain loop now (find-and-splice the read's own `woken`, tally by path+type into `woke`), and the `table()` formatter is gone: `expect(woke, …).toEqual([])` prints the rows itself. Verified live against preview_5 after the change: still `1 expected fail`, 134s.

apps/os/e2e/vitest/abandoned-project-goes-quiet.e2e.test.ts
mmkal: 🤖 Inlined in 3bc86d777 — the measurement is one plain loop now (find-and-splice the read's own `woken`, tally by path+type into `woke`), and the `table()` formatter is gone: `expect(woke, …).toEqual([])` prints the rows itself. Verified live against preview_5 after the change: still `1 expected fail`, 134s.

apps/os/e2e/vitest/abandoned-project-goes-quiet.e2e.test.ts
mmkal: 🤖 Inlined in 3bc86d777 — the measurement is one plain loop now (find-and-splice the read's own `woken`, tally by path+type into `woke`), and the `table()` formatter is gone: `expect(woke, …).toEqual([])` prints the rows itself. Verified live against preview_5 after the change: still `1 expected fail`, 134s.

apps/os/e2e/vitest/abandoned-project-goes-quiet.e2e.test.ts
mmkal: 🤖 Renamed to `failWakeUp` in 3bc86d777.

apps/os/e2e/vitest/abandoned-project-goes-quiet.e2e.test.ts
```diff
@@ -0,0 +1,179 @@
...
+      await sleep(10_000);
+      await readOffsets(itx);
+      await sleep(5_000);
+      baseline = await readOffsets(itx);
+    }
+
+    // The fixture's own disposer — the thing a real teardown would hang off.
+    // Called here so the quiet window starts after it; the `await using`
+    // above runs it again at the end, which a disposer must tolerate anyway.
+    await handle[Symbol.asyncDispose]();
```
mmkal: why don't we move it into the inner block and just do a `let projectSlug: string;` outside. looks like `handle.project.slug` is all it's needed for in the out block anyway

apps/os/e2e/vitest/abandoned-project-goes-quiet.e2e.test.ts
```diff
@@ -0,0 +1,179 @@
...
+          row.last = event.createdAt;
+        } else {
+          woke.push({
+            path,
+            type: event.type,
+            count: 1,
+            first: event.createdAt,
+            last: event.createdAt,
+          });
+        }
```
mmkal: nitpicking because readability is so important. can we make woke a Record<...> instead then this part would be:
```suggestion
        woke[r.path] ||= { path, type: event.type, count: 0, first: event.createdAt, last: event.createdAt };
        woke[r.path].count += 1;
        work[r.path].last = event.createdAt;
        const row = woke.find((r) => r.path === path && r.type === event.type);
```

i've missed out some nuance with the woke.push above i suspect but something like this would be easier to grep

apps/os/e2e/vitest/abandoned-project-goes-quiet.e2e.test.ts
```diff
@@ -0,0 +1,179 @@
...
+      }
+    }
+
+    const total = woke.reduce((sum, row) => sum + row.count, 0);
+    expect(
+      woke,
+      `project ${handle.project.slug} should be quiet after dispose (${QUIET_SECONDS}s), but it woke ${total} times after dispose`,
+    ).toEqual([]);
```
mmkal:
```suggestion
    const total = woke.reduce((sum, row) => sum + row.count, 0);
    const message = `project ${handle.project.slug} should be quiet after dispose (${QUIET_SECONDS}s), but it woke ${total} times after dispose`;
    expect(woke, message).toEqual([]);
```

mmkal:
🤖 Re the "move handle into the inner block" thread (couldn't reply inline while a review is pending): done in e0df1bcc4, plus one step — `itx` came from `handle` too, so the measurement now dials its own fresh `createAdminOsItx({ context })` after the block; nothing the test held open survives it. What escapes is one `project` record (id + baseUrl for that dial, slug for the message). No more explicit or double dispose. Live against preview_5: still `1 expected fail`, 137s.

apps/os/e2e/vitest/abandoned-project-goes-quiet.e2e.test.ts
```diff
@@ -0,0 +1,179 @@
...
+        continue;
+      }
+      const events = await itx.streams.get(path).getEvents({ afterOffset: since });
+      // This read booted the stream, which appended exactly one `woken` after
+      // readStart. Drop that one; any other late `woken` is a real wake.
+      const ourBoot = events.findIndex(
+        (event) => event.type === WOKEN && Date.parse(event.createdAt) >= readStart - 1_000,
+      );
+      if (ourBoot !== -1) events.splice(ourBoot, 1);
+      for (const event of events) {
```
mmkal: is this equivalent to a filteR?
```suggestion
      // This read booted the stream, which appended exactly one `woken` after
      // readStart. Drop that one; any other late `woken` is a real wake.
      const withoutOurBoot = events.filter(
        (event) => event.type !== WOKEN || Date.parse(event.createdAt) < readStart - 1000,
      );
      for (const event of withoutOurBoot) {
```

mmkal: 🤖 Not a plain filter — that would drop every late `woken`, and the contract (Bugbot's earlier catch) is exactly one: the boot this read caused; a second late `woken` is a real wake and must count. But filter-by-index reads better than splice and keeps that: `const withoutOurBoot = events.filter((_, index) => index !== ourBoot)` in 679cfb19f. Live: `1 expected fail`, 146s.

apps/os/e2e/vitest/abandoned-project-goes-quiet.e2e.test.ts
```diff
@@ -0,0 +1,189 @@
...
+import { createTestProject } from "../test-support/create-test-project.ts";
+import { createAdminOsItx } from "../test-support/os-client.ts";
+import { deployedBaseUrl } from "./test-helpers.ts";
+
+const QUIET_SECONDS = Number(process.env.ABANDONED_PROJECT_QUIET_SECONDS || 60);
+const WOKEN = "events.iterate.com/stream/woken";
+
+// Everything fits under the heavy-test ceiling (E2E_HEAVY_TEST_TIMEOUT_MS,
+// 240s, guarded by scripts/preview/e2e-policy.test.ts): project creation and
+// the first turn (~30s locally, up to ~90s under full-lane CI load), 15s of
```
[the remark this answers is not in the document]
mmkal: 🤖 Reworded in 9823d3ec7: "when the whole preview e2e suite runs concurrently on the slot".

mmkal:
Here's the follow-up list as it stands, grouped by what it's for, with where each lives.

Stop the leak at the source (highest leverage)
1. Test-run teardown — the fixture disposer is a no-op (create-test-project.ts:46, "itx has no projects.remove"). Delete the project at test end through a real product path that cascades to its scheduler, streams, repos, secrets, sandboxes; end-of-run sweep as fallback. I drafted the agent prompt for this earlier; it's the mitigation the abandoned-project-goes-quiet pin (#2583) is waiting to flip green on. Check tasks/os-project-archival.md first — may already specify it.
2. Alarm-fire counter — the pin only sees appended events; a resident DO re-arming silently is invisible. Durable { alarmFires, lastAlarmAt } in DO KV surfaced via runtimeState() (stream + scheduler DOs), then the test asserts "no events and no fires." Tracked in tasks/abandoned-project-goes-quiet.md.
3. Mid-turn abandonment scenario — the current pin covers the heartbeat (Flavor A). A second scenario that kills the interceptor mid-turn would pin the 10-second stream-loop flavor. Same task file.
4. Don't install the heartbeat on test projects — the template says "delete it when a project needs no heartbeat"; test-created projects need none. In the incident doc's mitigations.

Find the regression
5. Bisect late August — Aug 7–31 was ~0 DO-hours/day with the same CI; Sep 1+ was not. Deploy main@Aug-25 / Aug-28 / now to one slot, run one fixed spec, measure residual burn. Prime suspects #2532 (deadline-derived wakes, incl. the halted-subscription re-probe that re-opens 15 delivery attempts on every deploy) and #2520. You also wanted: revert #2532 + a failing test for the stall it fixed, so both problems get solved together. In the doc and tasks/stream-do-wake-loop-runaway.md.

Bound the loops (the root-fix task, tasks/stream-do-wake-loop-runaway.md)
6. Keepalive budget must not reset on a quiet-clean cycle that contains a revival.
7. Facet-alarm retry counter is an in-memory Map — evictions zero it while the alarm survives.
8. Bound the "a deploy fixed it" reset: at most one re-budget per stream until it settles clean or is externally touched — prod-safe, stops commit B re-arming commit A's orphans.
9. stream/woken on every boot is the oscillator's fuel — coalesce or skip when the previous event is an undelivered woken.
10. Self-wake horizon — spec grilled in tasks/self-wake-horizon.md (worktree, local, unpushed); one decision open for you: explicit origin: "external" | "internal" (recommended) vs authority: "public" renewing. Builds on #2579's armer hook.

Operational
11. Ice switch #2579 — built and tested, awaiting your review. Scheduler DO not covered (deferred in its task file).
12. failing() pins get retried once in CI — vitest's retry: 1 re-runs every expected-fail test; one-line fix ({ retry: 0 } in the helper). Small separate PR whenever.
13. Doc corrections — the write-up's Appendix A and why-2 cost line attribute 2.4 DO-hr/hr to heartbeats; the clean-specimen probe showed heartbeat wakes cost ~1s. You've rewritten in Notion; the Plannotator file still says the old thing.
14. Preview GC for leased-but-idle slots, and erase-before-every-preview-deploy as a cheap stacking stop — both in the doc's mitigation list, neither started.
15. Cloudflare account team — what the Enterprise contract actually charges for DO duration, and whether they can alert on usage from their side. Not started.

prd-specific
16. prd's chronic ~120 DO-hours/hr baseline (~$29/day, credit-covered for 51 more days) is the same disease at geological pace — worth an audit; flagged in the horizon task and the doc.

PR #2656, opened 2026-09-15
Fix Playwright loading, collaborative edit and preview auth flakes

mmkal:
Recurring browser flakes came from missing pending-state signals, stale Notes writeback, brittle collaborative undo coverage and preview OAuth limits that throttled parallel sign-ins.

- Track Todo add/check/delete operations per row, keeping unrelated controls usable and pending rows marked with `data-spinner`. Mark the OS Stop button as pending using the existing send state.
- Return typed `not-applied` results for Workspace edits whose source text changed or whose file disappeared. Notes treats these as superseded, preserving newer edits and deletions; blank notes remain unchanged without calling AI. Actual faults still throw.
- Pin the published Middlewright build with loading-aware `inputValue()` and the readiness recheck. Remove the arbitrary warm-up agent.
- Separate shopping-list undo from the Docs review/reconnect flow, and set preview-only OAuth/OTP limits to `600 / 60s`.

**Known quarantine:** the Docs review test still measures the two narrowly matched table/comment persistence failures. Shopping-list undo is a normal test. Remaining startup and registry-alarm issues are recorded in [the follow-up task](https://github.com/iterate/iterate/blob/fix/playwright-flake-causes/tasks/preview-stream-startup-stalls.md).

### Validation

- Blank-note regression (`0277de205`): 15/15 Notes tests pass, covering zero-byte, whitespace-only and frontmatter-only files plus existing concurrent-edit/deletion cases. Repository checks and all six deployed test suites pass on this commit.

- Review timeout corrections (`ff6866850`): all three seeded-app auth flows now hand navigation to the destination assertions; shopping-list persistence uses one direct read. **63/63** fresh-preview runs passed with zero retries and no allowed-flake matches. Required CI checks passed on this code commit.

- 52/52 focused Todo/Notes preview runs passed, including 48 at 16 workers, zero retries and no HTTP 429s.
- 84/84 deployed conditional-edit, Docs undo and Notes checks passed with zero retries in the preceding validation round.
- Auth's handler test accepts 600 requests from one IP and rejects request 601; all 108 auth tests pass.
- Detailed deployments, earlier failures and telemetry limitations are retained in [the validation record](https://github.com/iterate/iterate/blob/fix/playwright-flake-causes/tasks/complete/2026-09-15-playwright-flake-causes.md).

Repository-wide typechecks, lint, formatting and knip pass. The full local test suite passes with packages run sequentially. The initial overlapping run hit the unchanged CLI strip-loader test’s 5-second timeout; it passed in 817ms on the rerun without a timeout change. All CI checks for recording head `60ff6404c` completed successfully (the optional DO duration probe was skipped).

### Risk map

Review Workspace's return contract and Notes' handling first: callers must inspect `status`; a rejected conditional edit is no longer an exception. Generated contracts follow those changes. Then review Todo's reconciliation, which handles live-state confirmation arriving before the RPC response. Preview auth overrides are gated on fixed test OTP; production limits are unchanged. No data migration is required.

### Seeded-app recordings

Recorded with `VIDEO_MODE=1` against preview 3 at `60ff6404c`, with no test retries. These are the rendered videos, trimmed past sign-in. All three bodies passed; Docs review recorded `pass`, not an allowed failure.

**Todo — concurrent adds, per-row saves/deletion, and persistence after reload (21s).**

https://github.com/user-attachments/assets/d8513b83-8fad-4285-8818-eabaebc92bf1

**Shopping-list undo — undo red → green apples while preserving the peer's smooth peanut butter edit (7s).**

https://github.com/user-attachments/assets/1354a0be-2282-4650-bb84-b575601487f5

**Docs review — concurrent edits, comments, reconnect, persistence and narrow-screen layout (51s).**

https://github.com/user-attachments/assets/798e5041-08c3-458d-b9b5-63cb87fbd44c

Codex session: `01a0a1e6-0135-7902-91aa-b4bb07026de2`

| --- | ---: | ---: |
| Product | +150 -77 | +119 -64 🟩🟩🟥⬜⬜ |
| UI components | +1 -0 | +1 -0 🟩⬜⬜⬜⬜ |
| Tests | +333 -73 | +295 -61 🟩🟩🟩🟩🟥 |
| Config | +1 -1 | +1 -1 🟩⬜⬜⬜⬜ |
| Docs | +229 -0 | +150 -0 🟩🟩⬜⬜⬜ |
| Generated | +24 -20 | +8 -8 🟩⬜⬜⬜⬜ |
| **Total** | **+738 -171** | **+574 -134** 🟩🟩🟩🟩🟥 |

<sub>Lines counts every changed line; Significant ignores blank lines, JS comments, and TypeScript lines with no runtime output. Between 4a364c3 and 0277de2, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>

## Environment Config Lease

<details>

| Auth | released | `0277de2` | [https://auth.iterate-preview-12.com](https://auth.iterate-preview-12.com) | 855.6 KiB | 31.8s | 19.1s |  | 5.4s | [Workflow run](https://github.com/iterate/iterate/actions/runs/320641630713737) | 2026-09-15T16:05:32.188Z | Preview app released. |
| Docs | released | `0277de2` | [https://docs-preview-12.iterate-dev-preview.workers.dev](https://docs-preview-12.iterate-dev-preview.workers.dev) | 1.10 MiB | 18.6s | 2.5s |  | 4.5s | [Workflow run](https://github.com/iterate/iterate/actions/runs/320641630713737) | 2026-09-15T16:05:31.313Z | Preview app released. |
| Dummy Petshop | released | `0277de2` | [https://dummy-petshop.iterate-preview-12.com](https://dummy-petshop.iterate-preview-12.com) | 256.0 KiB | 335ms | 8.1s |  | 1.6s | [Workflow run](https://github.com/iterate/iterate/actions/runs/320641630713737) | 2026-09-15T16:05:29.677Z | Preview app released. |
| OS | released | `0277de2` | [https://os.iterate-preview-12.com](https://os.iterate-preview-12.com) | 5.43 MiB (-10.5 KiB vs main) | 55.7s | 279.6s | 1 retried: explicit child-agent delegation pipelines from an agent scope (relative path + create + message) (vitest x1) — Script execution orphaned: the incarnation running it went away before completing (eviction mid-run). It may have partially executed; it was NOT re-run. | 100.2s | [Workflow run](https://github.com/iterate/iterate/actions/runs/320641630713737) | 2026-09-15T16:07:06.956Z | Preview app released. |
| Semaphore | released | `60ff640` | [https://semaphore.iterate-preview-12.com](https://semaphore.iterate-preview-12.com) | 453.8 KiB | 26.6s | 24.5s |  | 1.3s | [Workflow run](https://github.com/iterate/iterate/actions/runs/320641630713737) | 2026-09-15T16:05:28.095Z | Preview app released. |
| Streams Example App | released | `0277de2` | [https://streams.iterate-preview-12.com](https://streams.iterate-preview-12.com) | 1.63 MiB | 14.7s | 107.1s |  | 5.5s | [Workflow run](https://github.com/iterate/iterate/actions/runs/320641630713737) | 2026-09-15T16:05:32.270Z | Preview app released. |

</details>

---

> [!NOTE]
> **Medium Risk**
> Public Workspace.edit contract change requires callers to inspect status; Notes and Todo reconciliation paths are concurrency-sensitive. Preview auth limit changes are gated on fixed test OTP only.
> 
> **Overview**
> Targets recurring Playwright flakes by tightening **pending UI signals**, **safe concurrent edits**, and **preview auth headroom**, with broad test and harness updates.
> 
> **`Workspace.edit`** no longer throws when the file is missing or `oldString` does not match; it returns **`status: "applied"`** or **`not-applied`** with `text-mismatch` / `file-missing` (settled and live collab paths). Ambiguous matches and real faults still throw. Generated ITX contracts and e2e expectations now include **`status: "applied"`** on success.
> 
> **Notes analysis** writeback uses conditional **`edit`** on the full file snapshot instead of **`writeFile`**, so stale AI results cannot overwrite newer user edits or deletions; blank bodies settle as **superseded** without calling the model.
> 
> **Todo** tracks per-row optimistic **add / setDone / remove** operations (client-allocated IDs, **`data-spinner`** on pending rows) so overlapping saves stay usable. **OS** marks the composer Stop button with **`data-spinner`** during interrupt.
> 
> When fixed test OTP is enabled, **email OTP** and **OAuth** preview rate limits move to **600 requests / 60s** (including parallel preview sign-ins from one IP). **Middlewright** is pinned to **`e3f2374`**; seeded-app specs drop warm-up agents, split Docs undo from review (**`createFlake`** for known review flakes), and use **`noWaitAfter`** on auth navigation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0277de205d256ca8339e1d08f885a4101056433c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


specs/seeded-apps.spec.ts
```diff
@@ -114,23 +128,150 @@ test("the seeded todo app authenticates a real project member", async ({ baseURL
...
+  );
+  await project.kv.set("docs-app-origin", docsOriginForBaseUrl(baseURL!));
+  const url = new URL(appUrl("docs", slug, baseURL!));
+  url.searchParams.set("workspace", "/agents/shopper");
+  url.searchParams.set("path", path);
+  await page.goto(url.toString());
+  await spinnerWaiter.settings.run({ spinnerTimeout: 130_000 }, async () => {
+    await page.getByRole("heading", { name: "Sign in to iterate" }).waitFor();
+  });
+  await page.getByRole("link", { name: "Continue with iterate" }).click({ timeout: 30_000 }); // timeout: cross-origin auth and cold build do not have continuous spinner-waiter progress
```
mmkal: could they though?

specs/seeded-apps.spec.ts
```diff
@@ -114,23 +128,150 @@ test("the seeded todo app authenticates a real project member", async ({ baseURL
...
+  await expect.poll(() => readEditorText(editor)).toContain("smooth peanut butter");
+  await editor.click();
+  await page.keyboard.press("ControlOrMeta+z");
+  await expect.poll(() => readEditorText(editor)).toContain("green apples");
+  await expect.poll(() => readEditorText(editor)).toContain("smooth peanut butter");
+  expect(await readEditorText(editor)).not.toContain("red");
+  await expect.poll(() => readEditorText(peerEditor)).toContain("green apples");
+  await expect
+    .poll(() => workspace.readFile(path), {
+      timeout: 30_000, // timeout: file durability follows the editor's visible state, outside spinner-waiter
```
mmkal: fine but why such a big timeout? i'd have thought this would be true already, not sure why we'd allow more than a second

specs/seeded-apps.spec.ts
mmkal: 🤖 Yes. Cross-origin navigation is not a fundamental obstacle, and my comment conflates two things. The callback already shows `Signing in…` (`apps/os/src/auth/project-auth.ts:542`), and the cold-build page already marks its status with `data-spinner` (`worker-serve-overlay.ts:214`). Middlewright also recognises document loading.

The specific gap is that this ready link's `click()` then waits for navigation under our 1-second action timeout (`e2e-policy/budgets.ts:59`). Spinner-waiter checks readiness/loading before calling the action; it doesn't extend the subsequent navigation wait. The explicit 30s bypasses that limitation.

We could handle navigation separately from the short action budget in Middlewright, or avoid waiting inside the click and let the following destination assertion own the navigation/loading wait. I'd verify that against the full auth redirect chain before removing the override. A pending indicator immediately on clicking Continue would also improve the UI, but a spinner alone doesn't change the timeout of an already-running click. No code changed for this question.

specs/seeded-apps.spec.ts
mmkal: 🤖 You're right; I don't have a reason for 30 seconds here. The preceding assertion already waits for the peer to display the undo. The collab engine persists the operation before publishing it (`collab-engine.ts:299–304`), and `workspace.readFile()` reads the live collab head (`workspace-durable-object.ts:477–481`), not a later filesystem flush. My “file durability follows” comment is wrong for this sequence.

I'd replace this poll with one `expect(await workspace.readFile(path)).toBe(...)`. That read should already contain the result; the RPC still has its normal request deadline, but we shouldn't allow 30 seconds of stale content. No code changed for this question.

specs/seeded-apps.spec.ts
mmkal: 🤖 Implemented in ff6866850. All three seeded-app auth links now use click({ noWaitAfter: true }); the destination assertions handle navigation/loading through Middlewright. Removed the 30-second click overrides and enabled spinner-waiter on the previously bypassed Todo/Docs cold-build assertions.

Verified the distinction with a real-browser repro: a 1.8-second navigation times out inside an ordinary one-second click, but succeeds when the click hands off to the next assertion. Then all three complete preview scenarios passed 21 times each: 63 actual passes, zero retries, no matched Docs flakes. I should not have resolved this before implementing and validating it.

specs/seeded-apps.spec.ts
mmkal: 🤖 Implemented in ff6866850: replaced the 30-second poll with one expect(await workspace.readFile(path)).toBe(...), after the peer displays the undo. The shopping-list scenario passed 21/21 focused preview runs with zero retries. This now fails on stale file contents instead of waiting for them to become correct.

packages/iterate/src/starter-apps/notes/processor.ts
```diff
@@ -339,13 +344,20 @@ export class NotesProcessor extends StreamProcessor<NotesProcessorContract, Note
...
+    const edited = await this.deps.workspace.edit({
       path,
-      composeNoteFile(
+      oldString: current,
+      newString: composeNoteFile(
         { ...currentNote.frontmatter, title: analysis.title, tags: analysis.tags },
         currentNote.body,
       ),
-    );
+    });
```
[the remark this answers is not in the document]
mmkal: 🤖 Fixed in 0277de205. Notes now leaves an empty or whitespace-only body unchanged and settles with `superseded: note has no text to analyze`, without calling AI or attempting an edit. This covers both zero-byte files and notes containing only frontmatter; `Workspace.edit` still rejects an empty `oldString`.

The regression reproduced the zero-byte file settling as `failed` before the fix. All 15 Notes tests now pass, including the existing concurrent-edit/deletion cases. SDK typechecking, lint and formatting pass too.

The pushed commit also passes repository tests/lint/typecheck and all six deployed test suites, including mobile Notes. CI is finishing resource cleanup. Bugbot has already marked this thread resolved.

PR #2660, opened 2026-09-15
Voice: make the button-press path warm and short

jonastemplestein:
Pressing the button on the ESP32 HAVPE means "start a new stream now". Today that costs 3.3 to 8.2 s on production before the model can even be dialled (`setupVoiceAgent`: a fresh Stream Durable Object, a dynamic-worker facet in a fresh isolate, an agent birth, then a dial that crosses five Durable Objects). This PR keeps that shape and removes what is removable without new machinery; it also carries the benchmark that measures it. Whether the remaining floor is acceptable is being decided against os-next (see "Where the floor is").

## What changes (apps/os)

- **Secret DO**: substitution reuses the fold snapshot this incarnation last read (30 s, dropped on any local write; an exact-revision request the copy predates reads again), and the `secret/used` audit fact is appended behind the upstream call instead of in front of it. The origin pin and revision assertion still gate every use.
- **Project DO**: egress rules are served from the held copy. Rules change only through committed root-stream batches, and every one of those already reaches the Project DO (`indexCommittedBatchFacts`), so that delivery invalidates the copy and forces a read; a copy quiet for a minute refreshes in the background as a safety net. Previously any dial older than 5 s paid a root-stream fold snapshot.
- **Stream DO**: a userspace facet's resolved class is held per incarnation and rechecked in the background every 10 s instead of re-resolving (Repo DO head read plus build-key hashing) on every facet call. A changed ref or a clone-skew recovery nonce re-resolves before the call.
- Cherry-picked from #2648: a hosted processor's self catch-up read was durable-only, so its cursor could advance past microphone frames still in the ephemeral buffer and drop them (`b2f490bd3`, with its red/green regression).
- **Benchmark**: `pnpm cli voicelab startup --project <slug> --audio [--activate-on-setup false]` times a fresh stream from `setupVoiceAgent` to `conversation-accepted` and first audio over one authenticated project WebSocket. `--presetup` exists only as a measurement mode (it isolates the dial from stream creation); the device does not pre-provision anything.

## Measurements (2026-09-15)

Direct Node → OpenAI from the same Mac, five calls: `session.started` median 830 ms, first audio 1,755 ms.

| conversation-accepted, ms | main | this branch |
| --- | ---: | ---: |
| Production `templestein`, new stream on press (setup alone) | 8,218 / 5,007 / 4,986 / 3,316 / 1,532 before the dial | not deployed |
| Preview 15, new stream on press, median of 3 | 5,057 | 5,346 |
| Preview 15, dial only (stream already exists), warm, median of 5 | 1,378 | 1,169 (best 862) |
| Preview 15, dial only, after 120 s idle, median of 4 | 1,542 | 1,437 |
| Preview 15, dial only, after 600 s idle, first press | not run | 4,670 |

The dial segment (`call-started` → `session-configured`) fell from ~580 ms to ~480 ms, which is the OpenAI upgrade itself; warm, the dial sits about 340 ms above direct. None of that touches the 3 to 5 s that stream creation costs on press.

## Where the floor is

On press, a fresh stream pays, in series or overlapping: a fresh Durable Object of the 12–22 MB OS worker (1.2–1.9 s first call regardless of storage; a 1.9 KiB worker's fresh object takes ~330 ms), a dynamic-worker facet in a fresh isolate keyed by the stream path (`worker-loader.ts` `loaderIdentity`; its first loopback call is ~1.9 s cold), an agent birth (0.5–0.8 s), then the dial. The cheapest further cut is to key the facet isolate per project instead of per stream (one line, but every facet's egress provenance becomes the project root); the object activation cost is the bundle, and no in-memory copy survives eviction (after 600 s idle the first dial alone took 3.5 s). "Within a few hundred ms of direct" for a fresh stream is therefore out of reach on this worker without a bundle diet. The same press path was benchmarked on os-next (deployed `os.iterate2.com`, 2026-09-15 21:00 UTC, `apps/os-next/bench/voice-startup.ts`): a processor loaded from source in a completely fresh context opens GPT-Live through egress with the project secret and the client sees `session-started` in a median of 1,614 ms (cold 1,797, best 1,206, after 120 s idle 1,713) against 1,079 ms for a direct Node dial in the same hour; the processor's own dial is 805 ms, faster than the Mac's. Two Durable Objects in the dial path, no agent birth, and about 450 ms of the total is serial client wiring that os-next's root `itx.worker` funnel removes, for a floor near 900 ms. Recommendation: stop optimising voice startup on this worker and wait for os-next; this PR's three shortcuts are small and tested and can be merged or closed on their own merit.

## Risk map

- **Egress rules invalidation** (`project-durable-object.ts`, `egress-rules-read-plan.ts`). The held copy is exact only if every root-stream commit reaches `indexCommittedBatchFacts` before the next dial. That delivery is asynchronous, so a rule committed and dialled against within the same second can be missed for that one request; the 5 s window had the same shape. A missed delivery is bounded by the 60 s background refresh. Table tests cover the plan; the invalidation wiring is by inspection.
- **Secret snapshot reuse** (`secret-durable-object.ts`, `substitution-snapshot-reuse.ts`). A rotation through this object's own `update()` drops the copy immediately; a raw append to the secret stream from elsewhere is seen within 30 s. `fetchAtUpdatedOffset` callers always get a matching revision or a fresh read.
- **Audit append behind the dial.** `secret/used` is recorded via `waitUntil` and logged on failure; it no longer gates the call. It never authorised anything.
- **Facet resolution memo** (`stream-durable-object.ts`). An unpinned `@main` facet picks up a moved branch head within 10 s plus one call, where it used to on the next call. Pinned refs (what the installer writes) are unaffected.

On merge: prd DOs pick the new code up on their next activation; nothing is invalidated; the device contract is unchanged.

Suggested review order: `project-durable-object.ts` + `egress-rules-read-plan.ts`; `secret-durable-object.ts` + `substitution-snapshot-reuse.ts`; `stream-durable-object.ts` (`#userspaceFacetClass`); the cherry-picked runner change; `startup.ts` last.

LOC: platform +257/−37, tests +300/−7, benchmark +539, generated +13/−11 (17 files, +1,109/−55).


jonastemplestein:
Ditched: voice targets os-next only now, with the firmware. Replaced by #2664 (the voice relay and its agent as os-next facets, the kit's os-next port, and the platform fixes the board exposed). The measurements and the apps/os findings stay in this PR's body for the record.

PR #2665, opened 2026-09-16
Verify rejected event batches leave the stream unchanged

mmkal:
Prove rejected event batches leave the durable stream unchanged through the public RPC API. Eight explicit cases submit an invalid event type after a valid event in the same batch, check the intended validation error, and compare the full durable log after each rejection.

The former assertion checked that `sneaky` was absent without ever submitting that event. The test now exercises partial writes directly. Cases are unrolled as requested in review.

The original product and wildcard-count fixes landed in #2666 while this PR was being verified. Main is merged, and those files are unchanged here; the remaining diff is regression coverage and task records.

Validation:
- The stronger regression failed before the fix and passes against the merged deployed preview on Node 24.4.1. All four original failures pass.
- Broader preview run: **48 passed, 1 unrelated failure, 1 opt-in skip**. The existing pause test gave its schedule 1.5 seconds, then read it after it had already finished (`null.source` at line 108). Both this test and the unrolled regression passed in isolation. [Follow-up task](https://github.com/iterate/iterate/blob/codex/os-next-main-e2e-fix/tasks/os-next-schedule-pause-test-race.md); no quarantine or retry changes.
- The unchanged 5 MB WebSocket test fails with Node 26 on both main and preview, and passes with the repo-pinned Node 24.4.1.
- Typecheck, knip, formatting and focused lint passed. Full CI test, lint/typecheck and autofix jobs passed on final head `b68f36126`. One local full-lint run hit an unrelated `git blame` timeout; CI lint passed.
- Preview smoke checks passed; 611 captured invocations had zero exceptions or warning/error logs.

[OS-Next preview 2](https://os-next-preview-2.iterate-dev-preview.workers.dev), version `aca1c174-68c7-4556-b389-783966061dfd`.

Codex task: `01a0a99d-fe31-72e2-ae5b-d7539d2119f8`.

| --- | ---: | ---: |
| Tests | +40 -4 | +33 -4 🟩🟩🟩🟩🟥 |
| Docs | +45 -0 | +33 -0 🟩🟩🟩🟩⬜ |
| **Total** | **+85 -4** | **+66 -4** 🟩🟩🟩🟩🟩 |

<sub>Lines counts every changed line; Significant ignores blank lines, JS comments, and TypeScript lines with no runtime output. Between 9f4f9e9 and b68f361, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>

## Environment Config Lease

> [!CAUTION]
> Slot preview-2 was re-acquired after this PR's lease lapsed — previous deployments there may have been replaced. Run preview deploy to redeploy. (preview assign at 2026-09-16T09:57:10.978Z)

<details>

No preview apps recorded.

</details>


apps/os-next/e2e/stream.e2e.test.ts
```diff
@@ -150,10 +150,13 @@ test("the runtime guard rejects a non-string or blank type, committing nothing",
...
-  expect((await rejection(append(itx, { type: "" }))).message).toMatch(/non-empty type/i);
-  expect((await rejection(append(itx, { type: "   " }))).message).toMatch(/non-empty type/i);
-  expect((await readAll(itx)).map((e) => e.type)).not.toContain("sneaky"); // nothing committed
+  const before = await readAll(itx);
+  for (const type of [12345, null, true, {}, [], undefined, "", "   "]) {
+    expect(await rejection(append(itx, { type: "sneaky" }, { type }))).toMatchObject({
+      message: expect.stringMatching(/non-empty type/i),
+    });
+    expect(await readAll(itx)).toEqual(before);
+  }
```
mmkal: I prefer the unrolled form

mmkal: 🤖 Unrolled all eight cases, with the unchanged-log assertion after each rejected batch. The final test passes against the deployed preview on Node 24.4.1.

PR #2695, opened 2026-09-17
Experiment with restoring previews after test cleanup

mmkal:
After preview tests pass, their GitHub check can now turn green while the finish job retires test data and restores a usable preview. Later failure or cancellation changes that same check to non-green. The lifecycle lock stays held until the job really finishes.

The experiment succeeded: the full preview run passed, retirement took **18.2s**, restoration took **45.1s**, and the preview checks became green **72s before the finish job ended**. Merging main at `ca226f195` resolved the unrelated expired-test failure; all checks on that commit passed.

[Successful interactive trace — 5m 36s to green](https://depot-01a0ae9a-ca84-759c-817d-628147918a80--iterate.iterate.app/) · [Full experiment report](https://github.com/iterate/iterate/blob/codex/preview-park-restore/docs/preview-park-restore-experiment.md) · [Depot run](https://depot.dev/orgs/0p91s0lz49/workflows/lnhxm1vn0t)

| Change | Purpose |
| --- | --- |
| `scripts/preview/preview.ts`: post-test `erase --restore` | Keep the existing namespace retirement, then redeploy Auth, OS and Streams from the tested checkout into the authoritative leased slot. Auth reseeds OAuth records erased by reset. Ordinary deploy commands restore assets, secrets and bindings. |
| `scripts/ci/status.ts`: `set-pending-check-green` via trpc-cli | Complete the exact running Depot Check Run after all other preview jobs succeed. A separate commit status cannot override that check. Existing workflow/head/attempt guards remain in force. |
| Preview workflows: report validation, early green, then cleanup/restoration; `checks: write` | Publish success only after validating all consumer results. Keep the finish job and lifecycle lock running so Depot records cleanup failures on the same originating check. |
| Lifecycle/workflow tests, restoration receipt and PR notices | Check ordering and authoritative lease targeting; distinguish the tested deployment from the new live versions and show when restoration is incomplete. |
| Streams `openConnection` test assertions | Fix a race exposed by the first run: check the expected event by offset instead of assuming it remains the last callback batch. Preserve once-only delivery and closed-client exclusion. |
| CI trace green milestone and primary **Time to green** statistic | Measure when the preview checks first become successful, including historical runs, while keeping total workflow time and any later failure visible. The successful run below reached green in **5m 36s**, versus **6m 50s** total. |
| Experiment report and completed task | Preserve measured timing, namespace changes, usability, failure/recovery evidence and platform limitations. |

Example: `pnpm preview erase --pull-request-number 2695 --ran-head-sha <tested-sha> --restore`. Pre-deploy reset and PR-close/expiry cleanup keep their existing behavior.

### Validation

- Follow-up [step-transition probe](https://depot.dev/orgs/0p91s0lz49/workflows/5skl8b041l): local polling continued roughly once per second through multiple post-green shell steps, artifact upload and job completion. All 113 post-success samples stayed completed/success; 21 independent runner-side boundary/phase checks also stayed green. No API errors or yellow regressions were observed. This closes the original observer's gap: it had stopped reading GitHub after the first green sample.

- All nine jobs passed in preview workflow `lnhxm1vn0t` on `151dc2f93`. The native check became successful at 06:07:46Z while Depot remained running until 06:08:59Z.
- Real Depot-token probes verified **early success → failure** and **early success → cancellation**. The first full run's real test failure correctly prevented early green.
- Cloudflare readback confirmed **ten ordinary namespaces replaced**, with six Sandbox namespaces retained. Fresh project creation, scheduled script-to-stream delivery and real Auth-to-OS login worked after restoration.
- A deliberately dirty checkout was rejected only after retirement; both Workers stayed at 503. Clean recovery succeeded. No failure switch is shipped.
- All checks passed after merging main at `ca226f195`. The Time to green update passes 369 scripts tests, scripts typecheck, scoped lint and formatting; desktop/mobile rendering of the historical trace shows 5m 36s versus 6m 50s without browser errors. All checks on the latest update, `89543a10d`, passed, including all nine preview jobs and trace publication. Earlier full lint/knip and independent review passed.

### Risk map

Review `status.ts` and `preview.ts` first, then workflow order, then tests/report. Completing Depot's check externally relies on **observed provider behavior**, not a documented support guarantee. A merge may happen before restoration finishes and its commit can turn red afterwards; that is the intended tradeoff. No repository protection settings changed.

Restoration creates new Worker versions: original test results concern the earlier versions, and the receipt records both. Failed restoration may leave a parked or partly restored preview. Cancellation cannot guarantee cleanup completes; slot reclamation/reset-on-acquire remain necessary. The existing container-class retention limitation is unchanged; the report distinguishes the measured quiet window from a guarantee that all retained workloads stop.

Fresh branch from main, separate from closed experiment #2693 and the change-type planner. No merge performed.

Codex task: `01a0aa2e-dd8a-76f0-9c26-c6be9cb5fbed`.

| --- | ---: | ---: |
| Tests | +39 -12 | +37 -12 🟩⬜⬜⬜⬜ |
| CI & scripts | +233 -35 | +222 -35 🟩🟩🟩🟩🟥 |
| Docs | +231 -0 | +198 -0 🟩🟩🟩🟩⬜ |
| **Total** | **+503 -47** | **+457 -47** 🟩🟩🟩🟩🟩 |

<sub>Lines counts every changed line; Significant ignores blank lines, JS comments, and TypeScript lines with no runtime output. Between be0a0be and ca226f1, bucketed first-match-wins into the groups defined in `scripts/ci/loc-report.ts`.</sub>

## Environment Config Lease

<details>

| OS | released |  |  |  |  |  |  | 100.5s |  | 2026-09-17T09:28:18.669Z | Preview app released. |
| Streams Example App | released |  |  |  |  |  |  | 6.8s |  | 2026-09-17T09:26:44.911Z | Preview app released. |

</details>

---

> [!NOTE]
> **High Risk**
> Changes preview lifecycle, GitHub check completion, and post-test redeploys on leased slots; merges may proceed before restoration finishes and the check can turn red afterward.
> 
> **Overview**
> Preview CI can mark the Depot-owned GitHub check **green right after all tests pass**, while the finish job keeps running to retire test data and optionally **restore** a usable preview. Later cleanup failure or cancellation can still flip that same check non-green; workflows gain **`checks: write`** and a sequential finish pipeline (merge results → early green → erase with **`--restore`** when prepare succeeded).
> 
> **`preview erase --restore`** extends post-test cleanup: after the existing slot reset it redeploys Auth, OS, and Streams from the **tested checkout**, using the **semaphore lease** (not editable PR report state), with guards for clean checkout, lease stability, and recorded deployments. It writes **`preview-restoration.json`** and temporary PR notices when restoration is incomplete.
> 
> **`scripts/ci/status.ts`** adds **`setPendingCheckGreen`**, which PATCHes the exact in-progress Depot check run for the current job after all other preview jobs finish. CI traces now surface **time to green** (including a `check-green` OTLP event and viewer copy).
> 
> A small **Streams capnweb e2e** change asserts events by **offset** instead of assuming the last callback batch, fixing a race seen during the experiment. Tests and **`docs/preview-park-restore-experiment.md`** document behavior, timings, and risks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 89543a10d082e348d9eddcf47d51563824ed5638. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/ci/status.ts
```diff
@@ -4,6 +4,64 @@ import { z } from "zod";
 
 /** GitHub milestones, guarded by Depot job liveness. Run with trpc-cli. */
 export default class CiStatus {
+  /** Publish test success while this job keeps cleaning up; Depot owns its final outcome. */
+  async testsPassed(summary: string) {
```
mmkal: Call this setPendingCheckGreen instead? Not limited to tests

scripts/preview/preview.ts
```diff
@@ -1628,12 +1630,131 @@ export async function erase(options: EraseOptions = {}) {
...
+      slot.slug,
+    );
+    if (holder !== run.holder) throw new Error("Preview lease changed during restoration.");
+    if (run.pullRequestNumber) {
+      const latest = await readPreviewPullRequest({
+        ...run,
+        pullRequestNumber: run.pullRequestNumber,
+      });
+      if (latest.headSha !== options.ranHeadSha)
+        throw new Error("PR head changed during restoration; the next run will reset its slot.");
```
mmkal, line 1691: Do we need an equivalent for main?

scripts/preview/preview.ts
```diff
@@ -1628,12 +1630,131 @@ export async function erase(options: EraseOptions = {}) {
...
+            app,
+            commandEnvironment: {
+              ...runtime.commandEnvironment,
+              PLATFORM_DEPLOY_HEAD_SHA: run.headSha,
+              // Same checkout and retained container classes; no image rollout is needed.
+              OS_CONTAINERS_ROLLOUT: "none",
+            },
+            dopplerConfig: slot.dopplerConfig,
+            existingEntry: null,
+            mainWorkerSize: null,
```
mmkal, line 1706: What are these? Whynull?

scripts/ci/status.ts
mmkal: 🤖 Renamed to `setPendingCheckGreen` (`set-pending-check-green` in trpc-cli), including the workflow caller. The helper's title/log are now generic too; the preview caller still supplies the test-specific summary. Pushed in a7599e207. The CLI help resolves the new command; all 360 scripts tests, typecheck, lint and formatting pass.

scripts/preview/preview.ts
mmkal: 🤖 No equivalent check against the moving main tip is needed with the current policy. `preview-main.yml` serializes the whole lifecycle with `cancel-in-progress: false`, and `createMainPreview` pins every command to that run's exact checked-out commit. We deliberately let commit A finish, including restoration, when commit B merges; B's run waits for the lock. Checking A against the latest main tip would reject that intended completion and can starve the baseline under frequent merges. The lease-owner check immediately above still applies to both main and PRs. No behavior change here.

scripts/preview/preview.ts
mmkal: 🤖 These are inputs to the existing deploy helper:

- `existingEntry` is the prior deployment record used to skip a deploy when its fingerprint and readiness still match. Passing `null` forces a real deploy after retirement.
- `mainWorkerSize` is only the bundle-size baseline for the PR table's “vs main” comparison. Passing `null` avoids fetching main's size artifacts again during restoration; it does not affect code, bindings or deployment. It does mean the restored rows omit that comparison, even if the earlier rows had one.

No behavior change here.

apps/streams-example-app/e2e/vitest/stream-capnweb.test.ts
```diff
@@ -463,9 +463,13 @@ describe("stream capnweb protocol", () => {
...
-    expect(callbackA.batches.at(-1)).toEqual([appended]);
-    expect(callbackB.batches.at(-1)).toEqual([appended]);
-    const callbackABatchesBeforeClose = callbackA.batches.length;
+    // Feed projections can arrive after this event or share its batch.
+    expect(callbackA.events.filter((event) => event.offset === appended.offset)).toEqual([
+      appended,
+    ]);
+    expect(callbackB.events.filter((event) => event.offset === appended.offset)).toEqual([
+      appended,
+    ]);
```
mmkal, line 472: what's this doing here? is it relevant to this PR?

scripts/ci/status.ts
```diff
@@ -4,6 +4,64 @@ import { z } from "zod";
 
 /** GitHub milestones, guarded by Depot job liveness. Run with trpc-cli. */
 export default class CiStatus {
+  /** Mark this job's GitHub check green while it continues; Depot owns its final outcome. */
+  async setPendingCheckGreen(summary: string) {
```
mmkal: this shouldn't be at the top of the class. Put it below the other two methods that correspond to CLI commands

apps/streams-example-app/e2e/vitest/stream-capnweb.test.ts
mmkal: 🤖 This is an incidental flake fix exposed by the first experiment run, rather than a requirement of restoration. The assertion failed twice in workflow xg4ww10ws2: it awaited a particular event, then assumed that event was still the last callback batch. A background feed projection could arrive in between. I included the small fix so we could get a reliable all-green run to verify early-green cleanup: assert the expected event by offset, and that the closed connection receives none of the later event. The focused live test and subsequent full runs passed. No Streams product behavior changed.

scripts/ci/status.ts
mmkal: 🤖 Moved `setPendingCheckGreen` below `set`, `waitFor` and `waitForJobs`, before the private helpers, in 89543a10d. The trpc-cli command remains available; scripts tests/typecheck pass and the latest preview is green.

